PR #99, opened 2026-07-24
fix: AppArmor-active gating + Arch profile, recovery stress tests, daemon panic isolation, lock-screen self-heal

archledger:
Bundled fixes + hardening from a deep stress-test/audit pass (validated live on archhost — Arch, AppArmor active, IR camera).

## Fixes
- **AppArmor**: gate the Repair + Login-wiring advisories on AppArmor being *enabled* (was distro-gated / read the CLI's own label); ship `usr.bin.irlumed` on Arch, load it when `apparmor_parser` exists.
- **irlume group**: create it (sysusers) so the socket is 0660 root:irlume, not 0666.
- **Lock-screen self-heal**: reconcile + Repair now detect a stripped `/etc/pam.d/kde` (was login-greeter-only).
- **Daemon panic isolation**: wrap each connection in `catch_unwind` so a bad-frame panic denies one request (→ password) instead of crashing the root daemon; plus source guards on `RgbView::pixel` (0-dim) and `eye_glint`/`eye_glint_contrast` (short IR frame).
- **Daemon hygiene**: zeroize the `/etc/shadow` read + crypt input; hard-fail (drop the connection) if the socket read/write deadline can't be set.

## Tests
- Recovery stress tests: wrong passphrase, missing/corrupt envelope, tampered TPM seal (swtpm subset in ci/preflight/hardware-suite).
- Panic-guard regressions for the degenerate-frame routes.

Security audits found **no** authz bypass, fail-open, PAM lock-out, or liveness bypass.


archledger:
## Validation ledger

What in this branch has been exercised on real hardware, and what has not. Written so the merge decision does not rest on memory.

### Validated on real hardware

| change | how, and on what |
|---|---|
| Face `sudo` on the Debian include layout (`8642b13`) | thinkpad, Ubuntu 26.04. A/B against 0.6.1 with both runs as root: the old build put the stanza at EOF and the module was never consulted (0 daemon lines); the new build put it above `@include common-auth` and the daemon was consulted, then fell through to the password. Correct password authenticates, wrong password fails, an already-misplaced stanza moves on re-run, second run is a no-op, reconcile is stable, polkit unchanged. Box restored to a byte-identical `/etc/pam.d`. |
| Renamed centre/edge cue + per-user floor (`146a7f3`) | Zenbook, real `plasmalogin` and `kde` PAM stacks. Face login and lock unlock both succeed; log shows `ir_center_edge_ratio=1.39` and `gate(per-user IR center/edge floor): live 1.39 vs floor 1.00`. |
| Capture-window / skew instrumentation (`763dae6`) | Zenbook, same run: `rgb/ir capture skew 190ms`, `524ms`, and `879ms` when the RGB self-heal fired. |
| `camera-tune` (`763dae6`) | Zenbook, real run through the daemon: measured 100% RGB and 83% IR retention, wrote `capture_mode.3277:0059:200901010001=concurrent`, daemon healthy afterwards. |
| Emitter re-fire requirement (`822bdf5`) | Two modules, two 30s runs each: ASUS built-in flat at 144, NexiGo N930W flat at ~37, neither self-clears. Control survives stream close and process exit. |
| Session-held streams (`055dbfe`) | Zenbook: 3 scans enrolled end to end through the held-session path against a scratch state dir, no fallback, faces found in both spectra. `session_bench` 8/8 pairs, 1912ms to 797ms. |
| `lit_mean` for the IR arm (`055dbfe`) | Zenbook: retention across runs tightened from 94/137/83% to 92/87/93%. |
| Everything from the earlier audit rounds (`079714d` through `968dae6`) | archhost and thinkpad, per the earlier validation rounds: AppArmor gating, recovery stress, panic isolation, lock-screen self-heal, remote-desktop deny, atomic seal writes. |

### NOT validated on hardware

| change | why not, and what it would take |
|---|---|
| Credential-release gesture gate (`387a563`, `c14babf`) | **Characterised and found wanting.** 17 attempts: holding still released 1 in 10, a single nod released 0 in 3, continuous nodding released 4 in 4. See #101. The detector accepts incidental movement, which is what an attacker holding a print produces. Needs the hand-held-print test before the gate can be credited with stopping that attack. It is default ON. |
| Fedora substack `facefirst` fix (`8642b13`) | The changed branch only runs on GDM below the GNOME on-demand gate or an unrecognised display manager. This box is plasmalogin, which is on-demand, so the path never executes here. Unit-tested only; would need an old-GNOME Fedora box or a deliberately unrecognised DM. |
| Sequential capture actually being used at auth (`763dae6`) | The ASUS measured healthy and stored `concurrent`, so the engine's sequential branch has never run on hardware. Would need a camera that measures as degraded, i.e. the NexiGo set as the active pair. |
| 3s cross-spectrum skew ceiling firing (`763dae6`) | Never exceeded in practice: the worst observed was 879ms. The deny-and-retry path is unit-tested only. |
| `SafeStream` teardown containment (`763dae6`) | Defensive. The STREAMOFF failure it contains has not been provoked, and provoking it needs a camera that errors on stop or a fault-injection harness. |
| Session-stable Y16 decode (`763dae6`) | Both attached IR nodes are GREY8, so the 16-bit path cannot be exercised here at all. Needs a Y16/Y10/Y12 IR camera. |
| Short IR-sequence reporting (`763dae6`) | Requires a camera that delivers frozen frames until the attempt budget runs out. Not reproducible on demand here. |

### The one that blocks a release

The gesture gate ships default ON and, on the only hardware it has been measured on, refuses users who follow its own instruction and releases 1 in 10 when consent is withheld. The wording fix (local, `de2fce4`) addresses the first half. The second half needs the print test.

Everything else in the not-validated column is either unreachable on the hardware we have, or a defensive path whose failure mode is a log line rather than a wrong answer.

Issue #108, opened 2026-07-26
Add a versioned machine contract for desktop integrations


loofiboss-bit:
## Context

[`plasma-irlume`](https://github.com/loofiboss-bit/plasma-irlume) is a KDE System Settings integration for irlume. Its read-only diagnostics support the documented 0.6.x CLI, but safe enrollment, preview, profile mutation, and transactional login activation require a published machine-facing contract.

The complete consumer requirements and proposed command/result shapes are documented here:

https://github.com/loofiboss-bit/plasma-irlume/blob/main/docs/UPSTREAM-API-REQUEST.md

## Requested surface

- versioned JSON for version, status, doctor, profiles, and login status;
- bounded NDJSON enrollment and claimed-user recognition-test events;
- explicit preview-only image bounds, typed positioning, cancellation, and cleanup semantics;
- fixed profile operations with opaque IDs and before/after identity;
- plan/apply/verify/rollback login transactions with password-fallback evidence;
- published schemas, compatibility policy, stable error codes, and release-derived tests.

## Security boundary

irlume remains the sole owner of camera access, inference, templates, TPM state, PAM planning/writes, verification, and rollback. The desktop integration owns presentation, cancellation, validation, and version gating. It will not parse mutation prose, access the private daemon protocol, or open the camera in parallel.

Synthetic consumer fixtures already exist in `plasma-irlume`, but they are intentionally not treated as proof of upstream support. A released and documented upstream contract is the remaining gate for live V2 workflows.

loofiboss-bit:
I opened #109 as a reviewed first slice: a versioned machine envelope, capability negotiation, read-only profile JSON, stable error codes, compatibility/privacy documentation, and process-level tests. It deliberately does not advertise the remaining enrollment/preview/mutation/login-transaction capabilities; the PR explains the engine/store and daemon-lifecycle work still needed.

loofiboss-bit:
The complete contract is now implemented as a reviewable stack:

- #109: machine envelope, capability negotiation, typed errors, read-only profiles
- #110: stable persisted profile/scan IDs with migrations
- #111: ID-based mutation primitives and exact rollback
- #112: public profile mutation commands
- #113: bounded JSONL enrollment/auth/add-scan streams, explicit bounded preview, cancellation, and disconnect rollback
- #114: fixed-scope plan/apply/verify/rollback login transactions for Plasma

[...]

PR #109, opened 2026-07-26
feat: add machine API foundation

loofiboss-bit:
## What & why

Add the first public, versioned machine-facing CLI surface for desktop
integrations, as a deliberately read-only foundation for #108.

- `irlume version --json` returns the contract and engine versions, public
  limits, and only capabilities that are implemented now.
- `irlume profiles list --json [--user USER]` wraps the existing typed daemon
  response without exposing daemon prose or the private socket protocol.
- Machine stdout is exactly one JSON document; usage, daemon availability,
  protocol, and operation failures use stable codes.
- Unknown, missing, and duplicate machine flags fail closed.
- `docs/MACHINE-API.md` freezes the compatibility and privacy rules.

This intentionally does **not** advertise enrollment streams, previews,
mutations, or login transactions. The current enrollment store identifies
profiles and scans by mutable names; inventing opaque IDs in the CLI would make
rename semantics unstable. Those remaining parts of #108 need engine/store and
daemon lifecycle work first.

## Testing

Fedora 44 container:

- `cargo fmt --all -- --check`
- `cargo test -p irlume-cli` — 279 passed, 12 hardware/model tests ignored
- `cargo clippy --workspace --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps`

The unfiltered workspace test command was also attempted locally. Its
`irlume-auth` engine tests require the repository's separately downloaded ONNX
weights and runtime; 15 tests stopped at model load in this model-free
container. The repository CI fetches and checksum-verifies those assets before
running the same workspace suite.

- [ ] Hardware-validated (not applicable: no PAM, daemon, or capture change)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,13 @@ All notable changes to irlume are documented here. This project adheres to
 
 ### Added
 
+- **Desktop integrations now have a small, versioned JSON foundation.**
+  `irlume version --json` advertises only implemented public capabilities, and
+  `irlume profiles list --json` exposes the existing read-only enrollment
+  summary without daemon prose or private socket access. The documented
+  contract keeps stdout machine-only, uses stable error codes, and deliberately
+  withholds mutation capabilities until the enrollment store owns opaque IDs.
+
 - **`sudo irlume camera-tune` measures whether your camera can read both sensors
   at once.** Some Hello modules stop exposing their colour stream properly while
   their infrared sibling is streaming. On a NexiGo HelloCam N930W the colour
--- a/README.md
+++ b/README.md
@@ -432,6 +432,7 @@ default IR-structure gate already rejects photos, screens, and video replays.
 |---|---|
 | **Install and set up**, guided or by hand | [`docs/SETUP.md`](docs/SETUP.md) |
 | Look up **every command and flag** | [`docs/COMMANDS.md`](docs/COMMANDS.md) |
+| Integrate through the versioned **machine API** | [`docs/MACHINE-API.md`](docs/MACHINE-API.md) |
 | Understand the **architecture** | [`docs/ARCHITECTURE.md`](docs/ARCHITECTURE.md) |
 | Read the **threat model** and standards mapping | [`docs/THREAT_MODEL.md`](docs/THREAT_MODEL.md) · [`docs/STANDARDS.md`](docs/STANDARDS.md) |
 | **Verify the claims** on my own machine | [`docs/VERIFY.md`](docs/VERIFY.md) |
--- a/crates/irlume-cli/src/machine.rs
+++ b/crates/irlume-cli/src/machine.rs
@@ -82,7 +82,11 @@ pub fn version(args: &[String]) -> ExitCode {
             json!({
                 "capabilities": CAPABILITIES,
                 "limits": {
-                    "max_profiles": 3
+                    // Read the engine's own constant rather than repeating the
+                    // number. A consumer displays this as the enrollment limit,
+                    // so a literal here would silently start lying the day the
+                    // store's limit changes.
+                    "max_profiles": irlume_core::storage::MAX_PROFILES
                 }
             }),
         ),
--- a/crates/irlume-cli/src/machine.rs
+++ b/crates/irlume-cli/src/machine.rs
@@ -1,0 +1,267 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright the irlume contributors.
+
+//! Public, versioned machine output for desktop integrations.
+//!
+//! Keep this module deliberately narrower than the daemon's private wire
+//! protocol. A capability is advertised only after its public command, output
+//! shape, and compatibility rules are covered here and in `docs/MACHINE-API.md`.
+
+use irlume_common::{ProfileSummary, Request, Response};
+use serde::Serialize;
+use serde_json::{json, Value};
+use std::process::ExitCode;
+
+pub const CONTRACT_VERSION: u32 = 1;
+
+const CAPABILITIES: &[&str] = &["version-json", "profiles-list-json"];
+
+#[derive(Serialize)]
+struct Document {
+    contract_version: u32,
+    engine_version: &'static str,
+    command: &'static str,
+    ok: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    data: Option<Value>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    error: Option<MachineError>,
+}
+
+#[derive(Serialize)]
+struct MachineError {
+    code: &'static str,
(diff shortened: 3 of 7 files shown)


archledger:
Keeping this one open. The shape is right: small, versioned, read-only, stdout machine-only, stable error codes, capability advertisement, and it deliberately does not invent opaque IDs before the store owns them. That last call is the one I would have argued for.

Detail is on #108, including the lock-screen regression your report led me to find and fix in #116.

Two things before this lands:

1. `main.rs` peeks for `--json` ahead of the subcommand match:

   ```rust
   (Some("profiles"), Some("list")) if args.iter().any(|arg| arg == "--json") => ...
   ```

   That works today but gets fragile as commands grow, and it means a `--json` appearing anywhere in the argument list selects the machine path. Route it through the normal command dispatch instead.

2. `limits.max_profiles` is hardcoded to `3` in `machine.rs`. The engine already knows the real limit (`irlume_core::storage`), so read it rather than freeze a literal into the published contract.

Also worth a look: your examples say `"engine_version": "0.6.1"`. Contract v1 will land against 0.7.0, which is where the socket fix and the reconcile timer are.

One process note: CI has not run on this yet because fork PRs from a first-time contributor need approval. I have checked that approving is safe here (the self-hosted runners deliberately trigger on `push` and schedule, never `pull_request`, so fork code cannot reach them, and the hosted workflow references no secrets), so I will approve the run.

archledger:
I reviewed this by building it on current `main` and running it, not just reading it, and then pushed two maintainer edits to your branch (`fd785e3`). Both are small; nothing changes behaviour except where `max_profiles` is read from. Details below so you can see exactly what I changed and why.

## What I verified working

| Check | Result |
|---|---|
| stdout purity | stderr **0 bytes**, exactly **1 line**, valid JSON, correct field types |
| exit codes | 0 success, 2 usage error, 1 failure, consistently |
| human paths | `profiles list`, `version`, `--version` all unaffected |
| merges on current main | clean, builds, tests pass |

Two calls in here are better than they might look from outside, and I want to name them:

**Declining to invent opaque IDs** before the store owns them is exactly right, and it is what I would have argued for.

**Not advertising `login-transactions`** means merging this switches on your KCM's capability negotiation while every privileged action stays fail-closed. I confirmed that by running your `validEnvelope` logic against this PR's real output: all five conditions pass, and `login-transactions` is absent. So irlume keeps the switch, which is the property both sides want.

## What I changed

**`limits.max_profiles` now reads `irlume_core::storage::MAX_PROFILES`.** The literal `3` was correct, but `plasma-irlume` renders this as the enrollment limit, so a duplicated literal would start misinforming users the day the store's limit changes. Added an integration test that compares the published value against the constant, so reverting to a literal fails CI.

**`docs/MACHINE-API.md` now states four things it did not.** I measured the gaps: authorization semantics, deprecation policy and timeout expectations had zero mentions.

- **A capability is not a compatibility promise.** Gate on `contract_version`; treat a capability as an availability check on top. This matters because of something your helper made me realise: a published capability name is effectively permanent public API. Your gate enables privileged behaviour on capability presence with no engine-version ceiling (`plasma-irlume.spec` requires `irlume >= 0.6.0`, no upper bound), so the day irlume advertises a name, every installed copy of the KCM acts on it, including old ones. That is a reasonable design given no alternative existed, but it means irlume must never reuse a name or publish one early. Now written down.
- **Authorization**: the daemon decides from the connecting process's credentials, not from what the caller claims. Root, or the owner of the records. `--user` selects an account, it does not grant anything.
- **Timeouts and cancellation**: reads return promptly, camera commands take as long as a capture, the consumer sets its own timeout, and terminating the process is the cancellation mechanism in contract v1.
- **Deprecation**: nothing documented is removed or redefined within a contract version.

## One gap this PR leaves, documented rather than hidden

Running the binary: `--user root`, `--user alice`, and `--user nonexistentuser999` all return byte-identical `operation-failed` with exit 1. That is because the daemon reports those outcomes to the CLI as prose rather than typed values, so `machine.rs` cannot distinguish them without string-matching, which would be worse.

Consequence for you: the KCM cannot tell "you are not permitted" from "the engine broke", so it should show a neutral failure rather than guess. The one thing the identical treatment does buy is that this command cannot be used to discover which accounts exist or which are enrolled.

A distinct authorization code needs typed daemon outcomes. That is upstream work on my side, it will be additive, and I am designing it now.

## Two notes, neither blocking

The `--json` peek before the subcommand match scans the whole argument vector. I tried to break it and could not: `enroll --name --json` still reaches the human enroll path, `profiles rename` is unaffected, and `profiles list --user --json` is correctly rejected as a usage error. So it is fragile rather than exploitable, and worth restructuring eventually.

`version_freezes_the_public_envelope_and_capabilities` constructs the document instead of calling `version()`, so it would pass if the real function changed. Your integration test covers the real binary, so this is duplication rather than a hole.

## Separately, about plasma-irlume

I read the repo and the project plan properly. The KAuth helper is careful work: absolute binary path, `setProgram`/`setArguments` so no shell is involved, `scope` allow-listed to three exact values, IDs validated by an anchored regex that cannot start with `-`, forced C locale, bounded output, timeouts, JSON rather than prose, and `auth_admin` on every mutating action. The plan's ownership split is also right, including assigning PAM wiring plans and rollback to the engine rather than the desktop.

Three things I would raise there, in your own interest:

1. The helper builds its environment from `QProcessEnvironment::systemEnvironment()` and only sets locale, so `IRLUME_SOCKET`, `IRLUME_STATE_DIR` and `ORT_DYLIB_PATH` pass through. Worth knowing that irlume's `secure_getenv` does **not** protect you here, because your helper is already root rather than setuid. Constructing a minimal environment and dropping every `IRLUME_*` and loader variable would be free hardening.
2. `Persistence=session` on the `auth_admin` actions means one administrator authentication covers the rest of the session. That is generous for authentication-stack and camera-firmware changes; one-shot would fit better.
3. Rollback is skipped when the apply response yields no transaction ID (`authhelper.cpp:373-384`). If irlume mutates PAM and then crashes or times out, the mutation stays and your helper cannot undo it. That one is genuinely mine to fix, not yours: apply has to be self-recovering inside irlume. It is the same class of problem as something I criticised in the larger stack, from the opposite direction, and it is the design question I want settled before any transaction API exists.

On versioning: a `v2.0.0` tag and a `Version: 2.0.0 Release: 1` RPM read to the outside world as a release, while your own CHANGELOG, `V2-IMPLEMENTATION.md` and `TEST-MATRIX.md` all say stable publication is blocked and V2 is not approved. I do not read that as a claim you are making falsely, since you documented the block yourself, but a pre-release label would match your intent better and would stop anyone treating it as a compatibility commitment on irlume.

Thanks for the report that started all of this. The lock-screen regression it led me to was real, it was mine, and it would have shipped.

PR #110, opened 2026-07-26
feat: add stable face record IDs

loofiboss-bit:
## What & why

Adds stable opaque identifiers to face profiles and scans as a prerequisite for the safe machine-facing profile mutations requested in #108 and proposed in #109.

- Fresh records receive random, type-prefixed 128-bit IDs.
- Legacy enrollments receive deterministic migration IDs derived from owner and record ordinals, never display names or biometric templates.
- Existing valid unique IDs survive rename and serialization; malformed or duplicate IDs are repaired deterministically.
- `ProfileSummary` exposes profile and positional scan IDs while retaining the existing `scans` field and defaulting new fields for old-daemon wire compatibility.
- No existing name-based CLI mutation behavior changes in this PR.

This PR is intentionally independent of #109 so the storage and daemon identity boundary can be reviewed on its own.

## Testing

Validated in a Fedora 44 container:

- `cargo test -p irlume-common` — 44 passed
- `cargo test -p irlume-core` — 98 passed, 20 hardware/fault-injection tests ignored
- `cargo test -p irlume-cli` — 272 passed, 12 model/camera tests ignored
- `cargo check --workspace --all-targets` — passed

The full workspace model-backed daemon suite was not run locally because the ONNX assets are not present; upstream CI fetches those assets. No PAM, authentication decision, or capture behavior is changed.

- [ ] Hardware-validated (if it touches PAM, the daemon, or capture)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,14 @@ All notable changes to irlume are documented here. This project adheres to
 
 ### Added
 
+- **Face profiles and scans now carry stable opaque IDs.** Fresh records receive
+  random 128-bit identifiers that survive display-name changes. Existing
+  enrollments are backfilled deterministically from user and record order,
+  never names or biometric templates, and the typed daemon profile summary
+  exposes the IDs without breaking older clients. This is the storage boundary
+  required for safe machine-facing rename, delete, and improve-recognition
+  operations.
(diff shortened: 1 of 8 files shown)


archledger:

PR #111, opened 2026-07-26
feat: add opaque profile mutations

loofiboss-bit:
## What & why

Adds ID-addressed profile mutations with typed before/after results for machine clients. This is stacked on #110, which supplies the stable profile and scan IDs, and advances the machine-contract work tracked by #108 and #109.

- Adds privileged `AddScanById`, `DeleteProfileById`, `DeleteScanById`, `RenameProfileById`, and `RenameScanById` daemon requests.
- Resolves every mutable target inside the trusted daemon instead of in an untrusted client.
- Returns a typed mutation kind, stable record IDs, explicit before/after display identity, and remaining scan count.
- Rejects malformed IDs and bounded/control-character display-name violations before mutation.
- Preserves the existing name-based request variants for current human CLI clients.
- Keeps the last-scan invariant and exact-profile targeting fail closed.

The first commit shown in this stacked PR is #110. Once #110 merges, GitHub will reduce this PR to the mutation commit.

## Testing

Validated in a Fedora 44 container:

- `cargo test -p irlume-common` — 44 passed
- targeted daemon mutation and display-name tests — 3 passed without loading camera models
- `cargo check --workspace --all-targets` — passed

The ONNX-backed daemon test process was also attempted with checksum-verified release models, but its engine initialization entered an upstream futex wait in the Fedora 44 container and was terminated. The new mutation primitives are therefore deliberately tested independently of model initialization; no PAM, authentication decision, or capture implementation is changed.

- [ ] Hardware-validated (if it touches PAM, the daemon, or capture)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,10 @@ All notable changes to irlume are documented here. This project adheres to
   exposes the IDs without breaking older clients. This is the storage boundary
   required for safe machine-facing rename, delete, and improve-recognition
   operations.
+- **Profile mutations can target opaque IDs and return typed results.** New
+  daemon requests add scans, rename profiles or scans, and delete profiles or
+  scans without resolving mutable display names in an untrusted client. Legacy
+  name-based requests remain wire-compatible for existing human CLI clients.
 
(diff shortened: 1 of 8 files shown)


archledger:

PR #112, opened 2026-07-26
feat: expose opaque profile machine API

loofiboss-bit:
## What & why

Extends the public machine API from #109 with the stable IDs from #110 and ID-addressed mutations from #111. This provides the first complete, safe profile-management slice requested by #108.

- `profiles list --json` now returns stored opaque profile and scan IDs and refuses an older daemon with missing/misaligned IDs.
- Advertises `profiles-json` and `profile-mutations-json` only because their public commands and stored identity now exist.
- Adds JSON profile/scan rename and delete commands using `--profile-id` / `--scan-id`.
- Validates bounded IDs and display names before contacting the daemon.
- Returns exact affected IDs, typed operation, explicit before/after identity, remaining scan count, `deleted`, and `mutated_other_profiles: false`.
- Keeps stdout to one JSON document and never exposes daemon prose, usernames, paths, templates, embeddings, or secrets.
- Retains all human name-based commands unchanged.

This is a stacked integration PR containing #109, #110, and #111 plus one profile-machine-API commit. It should be rebased after those prerequisites merge.

## Testing

Validated in a Fedora 44 container:

- `cargo fmt --all -- --check` — clean
- `cargo test -p irlume-common` — 44 passed
- `cargo test -p irlume-core` — 98 passed, 20 hardware/fault tests ignored
- `cargo test -p irlume-cli` — 283 passed, 12 model/camera tests ignored
- targeted daemon opaque-mutation/display-name tests — 3 passed
- end-to-end CLI/socket mutation test verifies the exact opaque ID and JSON before/after result
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` — clean

The full ONNX-backed daemon suite remains delegated to upstream CI; the new storage mutation tests deliberately do not initialize camera models. No PAM or authentication-decision behavior changes.

- [ ] `cargo test --workspace` passes
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` clean
- [x] Docs/CHANGELOG updated if user-facing
- [ ] Hardware-validated (if it touches PAM, the daemon, or capture)

PR #113, opened 2026-07-26
feat: add bounded enrollment event streams

loofiboss-bit:
## What & why

Adds the public, versioned JSONL contract required by desktop enrollment flows:

- bounded `enroll`, `auth test`, and ID-addressed `profiles add-scan` event streams;
- monotonic operation/session identity and exactly one terminal event;
- opt-in in-memory preview capped at 640x480, 96 KiB JPEG, 8 fps, and 256 KiB per JSON line;
- stable typed errors without daemon-prose parsing;
- prompt SIGINT/SIGTERM cancellation;
- exact-ID rollback in the client and atomic daemon-side rollback when a response cannot be delivered;
- auth test never releases credentials or mutates profiles.

This is stacked on the public machine API foundation and depends on #109, #110, #111, and #112. The commits remain separated so they can be reviewed/merged in order.

## Testing

- Fedora 44 CI container: all non-model workspace targets passed, including 206 CLI unit tests, 52 CLI integration tests, 45 common, 118 core (98 passed/20 hardware ignored), 67 camera (57 passed/10 loopback ignored), 30 liveness, 39 vision with 7 real-model tests excluded, and the new 9-test machine API suite.
- `irlume-auth` pure suite: 26 passed; preview size/aspect test included.
- daemon storage-only rollback and typed-error tests passed.
- The complete real-model workspace run reaches `add_scan_pre_camera_guards` and then stalls in an ONNX futex in this container. Hardware/model qualification is therefore deliberately not claimed here.

- [ ] `cargo test --workspace` passes (blocked by the documented ONNX futex stall in real-model tests)
- [ ] Hardware-validated (required before a stable consumer release)


archledger:

PR #116, opened 2026-07-27
fix: make the daemon control socket reachable again (0.7.0 blocker)

archledger:
Release blocker for 0.7.0. Found by the soak, not by a user: this never shipped.

## What broke

Adding `packaging/systemd/irlume.sysusers` created an `irlume` system group, which flipped the daemon into its `0660 root:irlume` socket branch. Nothing ever added a member to that group, and no documentation asked anyone to. The daemon-side branch itself predates this and shipped in 0.6.1; it stayed dormant there because packaging never created the group, so 0.6.1 installs got 0666 and worked.

Measured on Fedora 44 KDE, daemon healthy, face enrolled:

```
/run/irlume.sock                srw-rw---- root irlume
getent group irlume             irlume:x:965:        (no members)
connect() as the desktop user   EACCES
```

**The lock screen was broken.** `kscreenlocker_greet` is not setuid, so the `/etc/pam.d/kde` conversation runs as the user. Under strace, against the real stack:

```
connect(3, {sa_family=AF_UNIX, sun_path="/run/irlume.sock"}, 19) = -1 EACCES
```

Face unlock, and the keyring unseal that rides on it, silently fell through to the password. Alongside that, `irlume detect` exited 10 (partial) as a user and 0 (ready) as root on the same machine, and `status` reported "daemon NOT reachable" for a daemon serving requests. `sudo`, the login greeter and polkit were unaffected: all three run as root, which is why this survived earlier testing.

Our own `packaging/selinux/irlume.te` records the reasoning that missed it: "sudo and the KDE lock screen work without this module because both run in the user's UNCONFINED domain, which is already allowed." True of SELinux, silent on the permission bits.

## The fix

Socket is 0666; the group is gone from all three lanes. Reachability was never the authorization boundary. `SO_PEERCRED` is: kernel-supplied at connect time, unforgeable through protocol input (`unix(7)`), and every request already requires peer uid 0 or a target matching the peer uid.

This is the ordinary arrangement for a local system daemon. It is systemd's documented default for filesystem sockets (`SocketMode=`, `systemd.socket(5)`), pcscd ships the same, and fprintd, the closest analogue, keeps its endpoint reachable and authorizes per operation rather than by group.

Group membership could not have fixed it: supplementary GIDs are process credentials set at login, so adding a uid does not reach a running desktop (`newgrp(1)`), and greeter accounts differ per display manager (`sddm`, `plasmalogin`, `gdm`, `Debian-gdm`).

Reachability widened, authority did not. Requests stay bounded (64 KiB, read/write deadlines), connections stay isolated behind `catch_unwind`, and the unauthenticated dry-run emitter probe now shares the per-uid camera interval that already covered `Identify`. That interval is deliberately **not** applied to `Authenticate` (the real login path, throttled by consecutive-failure strikes) or `PositionSample` (the framing guide needs continuous samples).

Also fixes EACCES reporting, which was telling users a healthy daemon was down and sending them to `systemctl status`. `daemon_up() -> bool` became a typed `DaemonReach`. `detect` stays at exit 10 under EACCES, because readiness is genuinely unknown there and 0 would assert what it cannot see, but it names the real obstacle instead of guessing "not enrolled".

And packages `irlume-reconcile.timer` on Fedora: #103 installed it into the buildroot without a `%files` entry, which fails the rpm build on an unpackaged file, and left it out of `%preun`/`%postun`. The Packit check that would have caught this never ran on #103 (Copr backend: "Giving up waiting for copr_base repository").

## Hardware validation

Branch daemon on the real socket with the real enrolment, real PAM stacks driven with `pamtester`, packaged daemon restored on exit. Nothing on disk modified. **9/9 passed:**

| Check | Result |
|---|---|
| socket mode | `666 root:root` |
| real `/etc/pam.d/kde` as the user | `connect(...) = 0`, was EACCES |
| daemon served the lock-screen request | yes |
| `detect` as an unprivileged user | exit 0, "ready: daemon running and a face is enrolled" |
| `status` as an unprivileged user | "daemon: running", enrollment shown |
| `profiles` as an unprivileged user | lists profiles |
| real `/etc/pam.d/sudo` face path | `pamtester: successfully authenticated` |
| dry-run emitter probe, 1st then 2nd | XU controls listed, then "rate limited" |
| root exempt from the interval | yes |

Harness note: an earlier run reported this fix broken. A fixed 6-second sleep measured the stale 0660 socket file the stopped service leaves behind, before the branch daemon had finished loading models, so `connect` gave EACCES with no listener present. The harness now gates on the daemon listening and on our pid owning the socket.

Workspace: fmt clean, clippy `-D warnings` clean, all tests pass.

## Follow-ups, deliberately not in here

- `irlumed.service` still grants `CAP_CHOWN`, which existed only for the socket group chown. Nothing chowns anymore, so it can go, but reducing capabilities deserves its own validation pass rather than riding along in a blocker fix.
- Camera fairness across local uids is a real residual of wider reachability. The per-uid interval covers the unauthenticated probes; a general fair-share design for `PositionSample` and enrollment is separate work.
- Fleet confirmation is owed: a display manager whose greeter runs as a non-root account (SDDM as `sddm`, confined `xdm_t`) hit the same DAC gate, so greeter face login was likely broken there too. archhost and thinkpad were both unreachable while testing.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,47 @@ All notable changes to irlume are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- **Face unlock on the KDE lock screen works again, and `irlume detect` stops
+  lying to unprivileged callers.** An earlier change in this development cycle
+  had packaging create an `irlume` system group, which made the daemon restrict
+  its control socket to `0660 root:irlume`. Nothing ever added a member to that
+  group. `kscreenlocker_greet` is not setuid, so the lock screen's PAM
+  conversation runs as you, its `connect()` to the socket returned EACCES, and
+  face unlock silently fell through to the password along with the keyring
+  unseal that rides on it. `irlume detect` exited 10 (partial) as a user and 0
+  (ready) as root on the same healthy, enrolled machine, and `irlume status`
+  reported "daemon NOT reachable" while the daemon was serving requests
+  normally. `sudo`, the login greeter and polkit prompts were unaffected because
+  all three run as root.
+
+  The socket is now mode 0666 and the group is gone. Reachability was never the
+  authorization boundary: `SO_PEERCRED` is, it is supplied by the kernel at
+  connect time, and every request still requires either root or a target that
+  matches the calling uid. This is the ordinary arrangement for a local system
+  daemon. It is systemd's own documented default for filesystem sockets
+  (`SocketMode=` in `systemd.socket(5)`), pcscd ships the same, and fprintd
+  keeps its endpoint reachable and authorizes per operation. Group membership
+  could not have fixed it either: supplementary groups are process credentials
+  fixed at login, so adding your uid does not reach the desktop you are already
+  in, and greeter account names differ across display managers.
+
+  Requests remain bounded (64 KiB, read and write deadlines), each connection
+  stays isolated behind `catch_unwind`, and the unauthenticated dry-run emitter
+  probe now shares the per-uid camera interval that already covered `Identify`.
+  On Fedora the SELinux module is still the mandatory-access layer.
+
+- Connect failures now distinguish "nothing is listening" from "this uid may not
+  connect". EACCES no longer prints "irlumed is not running" or points you at
+  `systemctl status` for a healthy service, and it deliberately does not tell
+  you to run `sudo` or change the socket mode, because both hide whatever set it.
+
+- `irlume-reconcile.timer` is packaged on Fedora. It was installed into the
+  buildroot but never listed in `%files`, which fails the rpm build on an
+  unpackaged file, and it was missing from `%preun`/`%postun`, so an uninstall
+  left it enabled.
+
 ### Changed
 
 - **Releasing your keyring password now asks for a gesture, by default.** On a
--- a/crates/irlume-cli/src/commands.rs
+++ b/crates/irlume-cli/src/commands.rs
@@ -543,8 +543,37 @@ const WARN: &str = "\u{26a0}";
 const NO: &str = "\u{2717}";
 
 /// Reachability: a Ping that returns true iff `irlumed` answered.
+/// What a `Ping` established about the daemon.
+///
+/// This used to be a bare bool, and that lost the one distinction that matters:
+/// a user whose uid could not open the socket was told "NOT reachable
+/// (systemctl status irlumed)" and sent to inspect a service that was running
+/// normally. EACCES and "nothing is listening" are different answers and get
+/// different words.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum DaemonReach {
+    Running,
+    /// The socket is there but this uid may not connect. Carries no information
+    /// about whether the daemon is healthy, so never report it as "down".
+    AccessDenied,
+    /// Nothing is listening, or the reply was unusable.
+    Down,
+}
+
+/// Probe the daemon. Goes through the shared client directly rather than
+/// `daemon_request`, because the errno kind is the point here and the string
+/// conversion throws it away.
+pub(crate) fn daemon_reach() -> DaemonReach {
+    match irlume_common::client::request(&Request::Ping) {
+        Ok(Response::Pong) => DaemonReach::Running,
+        Ok(_) => DaemonReach::Down,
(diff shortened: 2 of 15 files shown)


archledger:
Added `a5e996a`, plus fleet measurements that narrow the blast radius and correct one claim I made in the PR description.

## The blast radius was narrower than I said, and I was wrong about greeters

I predicted a display manager whose greeter runs as a dedicated non-root account was blocked by the same gate. **That is false.** Checked against upstream source:

- SDDM authenticates in `sddm-helper` (`PamBackend`), started by the root daemon
- GDM in `gdm-session-worker`
- LightDM in `lightdm --session-child`, which explicitly tests `getuid() == 0` before `initgroups`
- greetd in its session worker, which drops uid/gid only in the inner session child

All four hold uid 0 through `pam_authenticate` and drop privileges only when starting the session, so a dedicated `sddm` or `gdm` account never reached this socket. The surfaces that broke are the ones where the user's own process drives PAM: the KDE lock screen, and the CLI. The bind-site comment now says this instead of listing greeter account names as if they mattered.

## Fleet: only the Fedora lane was affected, and there is a reason

**thinkpad (Ubuntu 26.04, .deb 0.7.0) was never broken.** Its socket is `666 root:root` and `irlume status` as an unprivileged user reports "daemon: running". The daemon log says why, every boot:

```
irlumed: could not chown socket to root:irlume (Operation not permitted (os error 1)); leaving 0666
```

and the kernel says why that failed:

```
apparmor="DENIED" operation="capable" class="cap" profile="irlumed" capability=0 capname="chown"
```

The AppArmor profile grants `dac_override` and `dac_read_search`, never `chown`. So on any AppArmor-enforcing box the chown failed, the daemon took its 0666 fallback, and the gate never engaged. That is also independent evidence that `CAP_CHOWN` was already dead weight on that lane.

**archhost (Arch) not yet measured.** Tailscale reports it `offline, last seen 11h ago` and neither LAN address I could find has sshd listening. Expectation is that it matches thinkpad, since AppArmor was put in enforce mode there earlier, but that is unverified.

## CAP_CHOWN removal, validated

Ran the installed daemon under a transient unit carrying the proposed set:

```
CapBnd: 000000000000000a
CapEff: 000000000000000a
  → cap_dac_override, cap_fowner        (CAP_CHOWN absent)
```

Under that set: socket bound at `666 root:root` with no chown error, and the existing root-owned 0600 enrollment read back through the TPM unseal path. `/etc/irlume` persistence is unaffected by construction, since `write_0600_atomic` creates with `create_new` and `.mode(0o600)` then fsyncs and renames, with no ownership call in the path.

`CAP_DAC_OVERRIDE` and `CAP_FOWNER` stay, because field installs may hold non-root-owned state from older workflows and dropping those needs a migration test rather than a guess. There is now a comment warning that the directive must not be deleted wholesale: for a uid-0 service, an absent `CapabilityBoundingSet=` grants the full root set, so "simplifying" it away is a privilege expansion.

## Installed and soaking

Built `irlume-0.7.0-2.fc44` from this branch and upgraded the daily driver, so the remaining soak covers the fixed code rather than the broken socket.

- reconcile timer came up **enabled and active** on a real upgrade, which exercises both the `%files` fix here and the arming marker from #103
- enrollment sha identical before and after (`317dcfb2c46a9858`), PAM counts unchanged (sudo 1, plasmalogin 4, kde 2, polkit-1 2), `rpm -V` clean
- real `/etc/pam.d/kde` as the user: `connect(...) = 0` and `pamtester: successfully authenticated`, so lock-screen face unlock works against the shipped package
- real `/etc/pam.d/sudo`: `pamtester: successfully authenticated`
- `irlume detect` exits 0 as an unprivileged user; `status` shows "running" with enrollment; `profiles` lists profiles

Rollback if needed: `sudo dnf downgrade irlume irlume-selinux`.

## Residual, filed rather than bolted on

#117 covers camera arbitration. The short version: the daemon has one synchronous request worker, the 15s deadlines bound socket I/O and not camera work, and `PositionSample`, `Enroll` and `AddScan` have no interval, so an authenticated local uid can delay a real login with legitimate self-scoped work. Wider reachability makes that residual more reachable, and it wants a priority queue with cooperative cancellation, not another rate check that would break the framing guide.

PR #130, opened 2026-07-27
ci: run the test suite under AddressSanitizer

archledger:
Closes the last open item from the security-CI research. The `pam_wrapper` item on that list turned out to be already done: `crates/irlume-pam/tests/pamwrap.rs` has 13 tests driving the real module through pamtester, and CI runs them.

## The unsafe surface this instruments

I measured it before building anything:

| Where | What |
|---|---|
| `irlume-common` | NSS `getpwnam_r`/`getpwuid_r` buffers, mlock/madvise page arithmetic, `secure_getenv` |
| `irlume-cli` / `irlume-camera` | a UVC XU control ioctl |
| `irlume-daemon` | the `SO_PEERCRED` path that authorizes callers |

Most of the rest is `geteuid()` and `isatty()`. So the job covers a specific handful of places: the ones where a mistake is a memory-safety bug in a program that authenticates people, and where clippy sees nothing.

**Not a required check, deliberately.** `-Zsanitizer` is unstable, so it needs nightly, and an unrelated nightly regression must not be able to block a PR. Triggers: pull requests touching `crates/**` or the manifests, pushes to main, and weekly.

## Verification

**The gate fires.** I planted a deliberate out-of-bounds read in `irlume-common` and confirmed the report, then removed it:

```
ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 1
    #0 ... deliberate_overflow_for_verification_only crates/irlume-common/src/memlock.rs:149:26
```

**Leak detection stays on.** Removing the suppression file still fails the run (`537 byte(s) leaked in 11 allocation(s)`), which is how I checked the suppression is scoped rather than blanket.

**Runtime:** about 20 seconds warm for the whole workspace, 25 test binaries.

## Sanitizer interactions with mlock and the pamwire tests

**The sanitizer defines its own `mlock`.** Confirmed from the shipped runtime's symbol table:

```
$ nm -g --defined-only librustc-nightly_rt.asan.a | grep -E ' (mlock|munlock|mlockall)$'
W mlock    W munlock    W mlockall
```

So the call never reaches the kernel, `RLIMIT_MEMLOCK` refuses nothing, and `mlock_refusal_warns_and_continues` could never observe its warning. The test now probes whether the lock was enforced and prints which assertion it skipped. Checked both ways: on stable the assertion still runs, under ASan it skips with a reason in the log.

**`pamwire`'s wiring tests leak on purpose.** `Svc.etc` is `&'static str` because the real paths are compile-time constants, so a test pointing at a tempdir has to `Box::leak`. The suppression names those test functions rather than the crate, so a leak anywhere else in `irlume-cli` still fails. The trade is stated in the file: a genuine leak inside those eight test functions would be hidden.

## Documentation

`docs/DEVELOPMENT.md` gains the reproduction, including the flag that is not optional. Without `--target`, build scripts and proc macros get instrumented too and the build fails.

### DIFF
--- a/.github/lsan-suppressions.txt
+++ b/.github/lsan-suppressions.txt
@@ -1,0 +1,16 @@
+# LeakSanitizer suppressions for the ASan job. Keep this file short: every entry
+# here is a leak nobody will be told about again.
+#
+# `pamwire`'s wiring tests build a `Svc`, whose `etc` and `vendor` fields are
+# `&'static str` because the real ones are compile-time paths. A test pointing at
+# a temporary directory has to produce a `&'static str` at runtime, which the
+# `leak()` helper does with `Box::leak`. That is a deliberate, bounded leak in
+# test scaffolding: eleven allocations totalling roughly half a kilobyte, all
+# freed when the test process exits.
+#
+# This is scoped to that test module rather than to the crate, so a leak anywhere
+# else in irlume-cli still fails the job. The trade it makes is that a genuine
+# leak inside those specific test functions would also be hidden; they allocate
+# almost nothing else, which is why the trade is acceptable here and would not be
+# at a wider scope.
+leak:irlume::pamwire::tests::
--- a/.github/lsan-suppressions.txt
+++ b/.github/lsan-suppressions.txt
@@ -8,6 +8,11 @@
 # test scaffolding: eleven allocations totalling roughly half a kilobyte, all
 # freed when the test process exits.
 #
+# A suppression matches the SYMBOL NAMES in the allocation stack. With no
+# symbolizer the stack is bare addresses, this entry matches nothing, and the
+# deliberate leak reappears as a failure that looks like a real one. The workflow
+# therefore resolves llvm-symbolizer and stops if it cannot find one.
+#
 # This is scoped to that test module rather than to the crate, so a leak anywhere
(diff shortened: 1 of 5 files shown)


archledger:
The first CI run of this job **failed**, while the same command passed locally. The cause is worth recording, because it is a way this kind of gate goes quietly wrong.

A LeakSanitizer suppression matches the **symbol names** in an allocation stack. The runner had no symbolizer, so the stack was bare addresses, the suppression matched nothing, and the deliberate test leak came back as a failure: `526 byte(s) leaked in 11 allocation(s)`, with no function names anywhere in the trace.

Reproduced locally rather than guessed at: `ASAN_OPTIONS=symbolize=0` produces the identical eleven allocations and the identical escape past the suppression.

Fixed by installing `llvm` and resolving `llvm-symbolizer` explicitly, and by **exiting before the suite runs if none is found**. A sanitizer report of bare addresses is not worth having even when nothing is suppressed, and a suppression that silently stops applying is worse than none at all. That is the same reasoning that makes `--strict` refuse to run without a validator in the conformance script.

`print_suppressions=1` now puts the evidence in the log, so this cannot regress silently again:

```
Suppressions used:
  count      bytes template
     11        537 irlume::pamwire::tests::
```

Issue #117, opened 2026-07-27
Camera arbitration: a local uid can delay a real login by monopolising the single request worker


archledger:
Follow-up from #116, deliberately left out of that PR. Researched rather than guessed, and the honest answer is that this needs a design, not another rate check.

## What is actually exposed

The daemon has one request worker: the main thread iterates `listener.incoming()` and calls `handle()` synchronously. A PAM caller opens its own connection, but that connection waits in the kernel accept backlog while an earlier request is served. It does not get a parallel camera worker.

The 15-second read/write deadline is set per accepted connection. It bounds socket I/O, not camera processing after the request has been read. A client can time out and disconnect while the daemon keeps working, and only learns of the disconnect when it writes the response.

Costs by request, for a non-root peer acting on its own account:

| Request | Bound today |
|---|---|
| `PositionSample` | none: opens the RGB camera, captures, runs detection, no interval, no operation deadline |
| `Identify` | 2s per uid, measured between *starts* rather than after completion |
| `Enroll` | none: default 10 scans, up to 10 assessment attempts per scan, camera sessions held open |
| `AddScan` | none |
| `SetupIrEmitter { dry_run: true }` | 2s per uid (added in #116) |
| `Authenticate` | retries within the 5s sudo/polkit or 15s login window; the failure throttle applies to outcomes, not presence retries |
| `TuneCaptureMode`, real emitter setup, `SelfTest`, ear calibration | root only |

So the exposure is broader than repeated preview calls: a local account can submit legitimate long self-enrollment work and, because dispatch is synchronous, delay root PAM traffic. There is no wall-clock cancellation anywhere for camera open, V4L2 dequeue, inference, or a wedged USB driver.

Before #116 the socket's permission bits happened to keep other uids out. That was never the authorization boundary and it broke the lock screen, so it is gone; this is the residual it was accidentally covering.

## Why the obvious fixes are wrong

- **A blanket per-uid interval breaks enrollment.** `PositionSample` is the framing guide and needs continuous samples for live feedback. Throttling it degrades exactly the flow it exists to serve.
- **A token bucket is an abuse-volume control, not a latency guarantee.** It still lets a request start immediately before a PAM request arrives, and it does not prioritise PAM.
- **Hard preemption is unsafe.** Killing a thread or unwinding across V4L2 or ONNX is not something to do on an authentication path; the code already treats a panic as grounds to rebuild the engine.
[...]

PR #132, opened 2026-07-27
feat(daemon): arbitrate camera work so an authentication is not queued behind it

archledger:
Addresses the arbitration half of #117.

## What was wrong

The daemon read one connection at a time and served it synchronously. A request nobody had read yet was invisible to it, so a local account running a legitimate self-enrollment or a preview loop held the only handler, and a lock-screen unlock arriving meanwhile waited in the kernel accept backlog with no way to overtake it.

Connections are now read and parsed on their own threads, and the parsed request is queued in a new arbiter. **One worker still owns the `Engine`**, so nothing changes about V4L2 and ONNX being driven from a single thread. What changed is that the daemon can now see what is waiting and choose.

Two rules do the work:

- An authentication, and the credential release that follows it, are taken before other camera work.
- Non-authentication camera work is **refused rather than queued** while an authentication is pending, and an unprivileged uid may hold at most one camera slot. Root is never capped, because the greeter and PAM are root and capping them would deny the login path.

Refusing beats queueing: a preview client told to retry has lost a frame, while one queued behind an enrollment holds a slot the login path may want.

## Measured on two machines

Eight preview clients from one unprivileged uid, looping `PositionSample`, then one `Authenticate`, timed from submission to answer. Three runs each.

| Machine | Camera | Before | After |
|---|---|---|---|
| Zenbook, Fedora 44 | real RGB+IR pair | 5.25s, 4.77s, 4.80s | 0.16s, 0.14s, 0.20s |
| archhost, Arch | v4l2loopback + ffmpeg | 1.72s, 1.72s, 1.72s | 0.08s, 0.00s, 0.00s |

The after figures are the bound the issue asked for: an authentication waits for at most the one camera operation already in flight. On the real IR pair a `PositionSample` costs about 0.4s, which is where the before figures come from: the authentication sat behind several of them.

Response tally from one arbiter run, which shows the rules working rather than the queue merely draining faster:

```
220421 preview responses: 218836 refused (slot), 1554 refused (auth priority), 31 served
```

## Stage 2: an enrolment in flight yields too

Reordering the queue left the case that started #117: an enrolment already
running held the camera for its whole ten-scan budget while an authentication
waited it out. Measured before this half landed, with an enrolment in flight, the
authentication waited **3.27s**.

The capture loop now asks whether something needs the camera more, at two points:
before the first frame, so an enrolment that already knows it is wanted elsewhere
never opens the device, and between whole captures, where nothing is
half-written. `Error::Preempted` is a distinct variant so a caller can tell a
yield from a failure.

Nothing is preempted by force: no thread unwound, no device closed under a
running capture, no ONNX session interrupted. Enrolment persists only after both
capture phases succeed, so returning early cannot leave a partial profile.

Measured on archhost against loopback feeders, an authentication submitted 1.2s
into an enrolment, three runs:

| | Result |
|---|---|
| Authentication answered | 0.14s, 0.17s, 0.14s |
| Enrolment returned | `preempted: an authentication needed the camera; nothing was saved, please retry` |
| Profiles written | 0 files, every run |

The CI loopback test covers the same path. Its stop signal answers "keep going"
once and "stop" after, so the yield lands on the in-loop boundary rather than the
cheap entry check, and it asserts the signal was consulted at least twice.

## What this still does not do

A capture already inside V4L2 or an inference session finishes; the yield lands
at the next boundary, which on the measured hardware is within a fraction of a
second. There is no wall-clock deadline on a wedged driver, which is a separate
problem from scheduling.

Worth knowing from the measurements: a refusal is cheap for the daemon (no camera) but still costs a connection thread, and a client spinning as hard as the test harness generated about 11k refusals a second. Concurrency is capped at 64 threads, but the rate is not. A refusal-side interval is a candidate follow-up if that ever matters in practice.

## Tests

Nine unit tests over the arbiter cover priority, both refusal reasons, slot release, root exemption, FIFO among plain work, wake-up, and drain-on-close. Two more drive a real `UnixStream` through `serve()`: one round-trips a request through a fake worker, and one asserts a queued authentication makes the connection thread refuse preview work **without a worker running at all**, so the refusal cannot be an artefact of a drained queue.

684 tests pass; fmt, clippy, `cargo doc -D warnings` and the machine-API conformance script are clean.

## Still owed before merge

A face unlock on real hardware. Everything above measures scheduling; none of it proves the authentication it prioritises still grants, because that needs a face in front of the camera.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -66,6 +66,14 @@ pub struct Engine {
     /// The gate treats `false` as "not seen yet", which is the fail-closed
     /// reading: the worst a stale `false` can do is ask for another gesture.
     gesture_seen_before_match: bool,
+    /// Asked between whole captures: "should this long operation stop now?".
+    ///
+    /// The daemon points this at its arbiter so an enrolment yields the camera
+    /// to an authentication. `None` (the CLI, tests) never stops. It is polled
+    /// only at a boundary where nothing is half-written, never mid-capture and
+    /// never mid-inference: stopping an operation is a scheduling decision, not
+    /// a way to abandon a device or a session.
+    stop_requested: Option<std::sync::Arc<dyn Fn() -> bool + Send + Sync>>,
 }
 
 /// Assurance tier of this engine, derived from the available camera hardware.
@@ -495,7 +503,23 @@ impl Engine {
             ir_dev: irlume_camera::DEFAULT_IR_DEVICE.into(),
             ir_available: irlume_camera::capabilities().ir_pair,
             gesture_seen_before_match: false,
+            stop_requested: None,
         })
+    }
+
+    /// Point the engine at a signal asked between whole captures, so a long
+    /// operation can yield the camera to an authentication.
+    ///
+    /// Only the daemon sets this. It is a request, not a kill: the check happens
+    /// at boundaries where nothing is half-written, so a stopped operation
+    /// persists nothing and the caller retries.
+    pub fn set_stop_signal(&mut self, signal: std::sync::Arc<dyn Fn() -> bool + Send + Sync>) {
+        self.stop_requested = Some(signal);
+    }
+
+    /// True when something has asked this operation to stop.
+    fn should_stop(&self) -> bool {
+        self.stop_requested.as_ref().is_some_and(|f| f())
     }
 
     /// Assurance tier from the hardware: `Secure` with a real RGB+IR camera,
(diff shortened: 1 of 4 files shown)


archledger:
## Hardware validation on the Zenbook (Fedora 44 KDE, real IR camera, enrolled face)

The branch daemon was put on `/run/irlume.sock` in place of the packaged one, with the same environment the systemd unit supplies.

**Face recognition still works through the restructured daemon.** `irlume identify` returned `wisbfime (profile 'BEN', score 0.776)` in 3.0s.

**All four real PAM stacks authenticate**, each on the first attempt:

| Service | Result |
|---|---|
| `sudo` | successfully authenticated |
| `kde` (lock screen) | successfully authenticated |
| `plasmalogin` (login manager) | successfully authenticated |
| `polkit-1` (app prompts) | successfully authenticated |

**The scenario the change exists for.** A real lock-screen unlock through `pamtester kde`, while eight unprivileged clients hammered `PositionSample` from another process:

| Daemon | Unlock under flood |
|---|---|
| Packaged 0.7.0 | **14.39s** |
| This branch | **4.27s** |

About 3s of each is the face capture and the nod gesture, so the queueing delay went from roughly 11s to roughly 1s. That 14s is what a user would experience today as a lock screen that appears to ignore them.

The machine is back on the packaged daemon: `/usr/bin/irlumed`, one process, `detect` reports ready, no errors in the journal.

## Still not covered

Cooperative cancellation of a running enrollment, which is the other half of #117. An enrollment already in flight still delays a login until it finishes; only queued work is reordered.

PR #154, opened 2026-07-28
fingerprint: don't count a PAM backup file as pam_fprintd wiring

archledger:
## The gap

`fingerprint enable` will not record `method=fingerprint` until it can see an active `pam_fprintd.so` line. That gate exists because standing face auth down while nothing drives the fingerprint prompt leaves the box silently password-only, with the user believing fingerprint is on. It was added to close B1 from the 2026-07-21 gap analysis.

The scan behind the gate read *every* entry in `/etc/pam.d`. `pam_start()` opens `/etc/pam.d/<service>`, and a service name carries no dot; every leftover in that directory does. Package managers write `.rpmsave`, `.rpmnew`, `.pacsave`, `.pacnew`, `.dpkg-old`, editors write `.bak`, and irlume's own wiring backs each stack up to `.pre-irlume` beside the original. A line in any of those satisfied the gate while PAM ignored the file.

## Not hypothetical

A Fedora 44 box running `irlume-0.7.0-1.fc44` from Copr carries six leftovers in `/etc/pam.d`:

```
kde.pre-irlume                        plasmalogin.pre-irlume
sudo.pre-irlume                       kde-fingerprint.pre-linhello
password-auth.pre-linhello-uninstall  system-auth.pre-linhello-uninstall
```

Three written by irlume, three by linhello. `system-auth.pre-linhello-uninstall` holds an active `pam_fprintd.so` line. On that box the live `system-auth` happens to carry one too, so the verdict is unchanged there; the defect is that the dead file alone would have produced it.

## The fix

Enumeration skips dotted names. **Matching the dot rather than a list of known backup suffixes is deliberate** — `.pre-linhello-uninstall` is not a suffix any such list would have held, and that is the one that actually had the line.

The rule governs *enumeration only*. A live stack can `include` a dotted file by name and PAM does follow that, which is why `fprintd_in_sudo` resolves include targets by name regardless. The cost is a false negative in that exotic case, which fails safe: method unchanged, face stays active.

`faillock_cohabits` shared the scan and the flaw, so it shared the fix. It was **not** misfiring on the observed box, whose backup carries no `pam_faillock` line; that half is a latent defect closed alongside, not a reported one.

## Testing

The regression test was written first and run against unmodified code to confirm it caught the bug:

```
test fingerprint::tests::pam_scans_ignore_files_pam_never_loads ... FAILED
panicked: a backup file is not wiring; enable must not record method=fingerprint from one
```

After the fix, 9/9 fingerprint tests and 705 workspace tests pass, `cargo fmt --check` and `cargo clippy --workspace --all-targets -D warnings` clean. The test uses the real leftover names observed in the field, and asserts the live stack is still read so the skip narrows which files count, not what a real stack may say.

## Not addressed

An active line in a live file that is not a login or sudo stack still counts as wired. That is a weaker check than "wired into a stack that runs at login", but it is a separate change with real breakage risk for the Arch flow, where the admin is told to add the line to `system-local-login`.

Hardware validation of the full `enable` path on a box with a real reader has not been done; `fp::available()` gates the flow before the wiring check, so a container cannot reach it.

### DIFF
--- a/crates/irlume-cli/src/fingerprint.rs
+++ b/crates/irlume-cli/src/fingerprint.rs
@@ -437,16 +437,42 @@ fn has_active_line(text: &str, needle: &str) -> bool {
     })
 }
 
-/// True when any file in `pam_dir` carries an ACTIVE (non-comment) line
-/// referencing `pam_fprintd.so`, i.e. something will actually drive the
+/// The contents of every file in `pam_dir` that PAM will actually load.
+///
+/// `pam_start()` opens `/etc/pam.d/<service>`, and a service name carries no
+/// dot: it is the application's own name (`login`, `sudo`, `plasmalogin`).
+/// Leftovers in that directory all do carry one. Package managers write
+/// `.rpmsave`/`.rpmnew`/`.pacsave`/`.pacnew`/`.dpkg-old`, editors write `.bak`,
+/// and irlume's own wiring backs each stack up to `.pre-irlume` BESIDE the
+/// original. Reading those as if they were live stacks is how a scan concludes
+/// "wired" about a file nothing loads.
+///
+/// Keying on the dot instead of a list of known backup suffixes is deliberate:
+/// a Fedora box in the field carried an active pam_fprintd line in
+/// `system-auth.pre-linhello-uninstall`, a name no such list would have held.
+///
+/// This governs DIRECTORY ENUMERATION only. A live stack may `include` a dotted
+/// file by name, and PAM does follow that; [`fprintd_in_sudo`] resolves include
+/// targets by name for exactly that reason. The cost here is a false negative in
+/// that exotic case, which leaves the method unchanged and face active.
+fn live_pam_stacks(pam_dir: &std::path::Path) -> Vec<String> {
+    let Ok(entries) = std::fs::read_dir(pam_dir) else {
+        return Vec::new();
+    };
+    entries
+        .flatten()
+        .filter(|e| !e.file_name().to_string_lossy().contains('.'))
+        .filter_map(|e| std::fs::read_to_string(e.path()).ok())
+        .collect()
+}
+
+/// True when a PAM stack PAM actually loads carries an ACTIVE (non-comment)
+/// line referencing `pam_fprintd.so`, i.e. something will really drive the
 /// fingerprint prompt. Unreadable dirs/files count as not wired.
 fn pam_fprintd_wired(pam_dir: &std::path::Path) -> bool {
-    let Ok(entries) = std::fs::read_dir(pam_dir) else {
-        return false;
-    };
-    entries.flatten().any(|e| {
-        std::fs::read_to_string(e.path()).is_ok_and(|s| has_active_line(&s, "pam_fprintd.so"))
-    })
+    live_pam_stacks(pam_dir)
+        .iter()
+        .any(|s| has_active_line(s, "pam_fprintd.so"))
 }
 
 /// True when one PAM service file stacks BOTH pam_faillock and pam_fprintd.
@@ -455,14 +481,9 @@ fn pam_fprintd_wired(pam_dir: &std::path::Path) -> bool {
 /// faillock failure (fprintd#209/#215). Doctor surfaces it with the
 /// `faillock --reset` remedy.
 pub(crate) fn faillock_cohabits(pam_dir: &std::path::Path) -> bool {
-    let Ok(entries) = std::fs::read_dir(pam_dir) else {
-        return false;
-    };
-    entries.flatten().any(|e| {
-        std::fs::read_to_string(e.path()).is_ok_and(|s| {
-            has_active_line(&s, "pam_faillock.so") && has_active_line(&s, "pam_fprintd.so")
-        })
-    })
+    live_pam_stacks(pam_dir)
+        .iter()
+        .any(|s| has_active_line(s, "pam_faillock.so") && has_active_line(s, "pam_fprintd.so"))
 }
 
 /// True when the sudo PAM service reaches pam_fprintd, either directly or via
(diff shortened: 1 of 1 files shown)


archledger:
## Hardware validation on a second box

Ubuntu 26.04, ThinkPad X13 Yoga Gen 4, Synaptics reader, `irlume 0.7.0` installed from the PPA. Both binaries run against the same `/etc/pam.d` in the same session.

### The dead-file class is not irlume's own mess

This box carries five `.pam-old` leftovers written by **Ubuntu's `pam-auth-update`**, and one of them holds an active line while the live stack does not:

```
$ grep -nE '^[[:space:]]*[^#].*pam_fprintd\.so' /etc/pam.d/common-auth.pam-old
17:auth	[success=3 default=ignore]	pam_fprintd.so max-tries=1 timeout=10 # debug

$ grep -n fprintd /etc/pam.d/common-auth
(no fprintd line in the live stack)
```

That is a third naming convention, after irlume's `.pre-irlume` and linhello's `.pre-linhello-uninstall`, and this one was written by the distro's own PAM tool. Any deny-list of known backup suffixes would have had to carry all three.

### A/B with a control

The probe file is named `zz-irlume-b1-probe`, not `common-auth.pre-irlume`, on purpose: `pamwire.rs`'s `backup()` is `if !bak.exists() { copy }`, so a leftover using a real backup name would make a later `login enable` treat the stub as the pristine original.

| condition | installed 0.7.0 | built from this branch |
|---|---|---|
| `zz-irlume-b1-probe.dead` (dotted) | **warns** | silent |
| `zz-irlume-b1-probe` (live name, identical content) | — | **warns** |

```
===== INSTALLED 0.7.0 (packaged, from the PPA) =====
  ⚠ pam_faillock and pam_fprintd share a PAM stack: a touch-sensor misread can burn every
    fingerprint retry in seconds, and each one counts toward the account lockout.
  exit=0  (0 = warned)

===== BUILT from fix/pam-scan-ignores-dead-files =====
  exit=1  (1 = silent, correct)
```

The control is the load-bearing half. Same bytes, same directory, same binary, only the filename differs, and the check still fires. Without it, "the warning stopped" is equally consistent with having broken the check.

Baseline before planting was silent, and no live file on this box stacks both modules, so nothing else could have produced the warning. Both probes were removed and `/etc/pam.d` confirmed clean.

### Out of scope, and now with a concrete instance

`gdm-fingerprint` is standalone (nothing includes it; GDM names it via `pam_start`) and holds the only live `pam_fprintd.so` line on this box, while `sudo` does `@include common-auth` and that stack has no fprintd at all. So `pam_fprintd_wired()` answers true here, legitimately, from a service that covers the greeter and nothing else. The check answers "wired somewhere", not "wired where the recorded method claims to work". Filed separately rather than folded in.

### Not validated

`fingerprint enable` end to end. No finger is enrolled on this box, so `enable` drops into enrollment and waits for physical touches on the reader. That needs someone at the machine.

archledger:
## The dot rule, checked against package databases

The fix rests on one assumption: no distro ships a PAM service file whose name contains a dot. Linux-PAM never enumerates `/etc/pam.d`; it opens the exact file `pam_start()` names, so a dotted file is reachable only if something names it explicitly. Verified against package databases rather than by inspecting directories, since a directory shows only what happens to be installed.

| distro | packaged `/etc/pam.d` files | with a dot in the basename |
|---|---|---|
| Fedora 44 (`rpm -qal`) | 37 | 0 |
| Ubuntu 26.04 (`/var/lib/dpkg/info/*.list`) | 34 | 0 |
| Arch (`pacman -Qlq`) | 25 | 0 |

96 packaged service names, none dotted. Each query was proved non-empty by printing real names from it. My first Ubuntu attempt returned zero rows because the query was wrong, not because nothing matched, and a vacuous check is not evidence.

Every dotted entry on all three boxes is unowned by any package:

```
$ rpm -qf /etc/pam.d/system-auth.pre-linhello-uninstall
file /etc/pam.d/system-auth.pre-linhello-uninstall is not owned by any package
```

## Four leftover conventions, one per tool

| convention | written by | seen on |
|---|---|---|
| `.pre-irlume` | irlume | all three |
| `.pre-linhello`, `.pre-linhello-uninstall` | linhello | Fedora, Arch |
| `.pam-old` | Ubuntu `pam-auth-update` | Ubuntu |
| `.pacnew` | pacman | Arch |

Two of the four come from distro tooling, not from this project. That is the case for matching the dot instead of enumerating suffixes.

Arch contributes no bug instance: no active `pam_fprintd.so` line in any file there, live or dead.

If the assumption is ever wrong on some distro, the failure is a false negative: the gate refuses, the method stays unchanged, and face auth stays active.

archledger:
## End-to-end on `fingerprint enable`, the gate itself

The earlier validation used `doctor` as a proxy, because `enable` needs a reader and an enrolled finger. Both are reachable without hardware: libfprint ships virtual drivers, and Ubuntu compiles them in.

**Setup.** Ubuntu 26.04 container, `fprintd` 1.94.5-4, `libfprint-2-2` 1:1.95.1+tod1. A finger was enrolled against `virtual_device_storage` with no hardware, driven over its UNIX socket. The protocol was read from `tests/virtual-device.py` in Ubuntu's own source package rather than guessed: one AF_UNIX connection per command, `"<CMD> <args>"` as UTF-8, close.

```
$ fprintd-enroll tester      # while SCAN commands are fed to /tmp/fp.sock
Enroll result: enroll-stage-passed   (x5)
Enroll result: enroll-completed

[fingerprint] reader   : Virtual device with storage and identification for debugging
[fingerprint] enrolled : 1 (right-index-finger)
```

Two things had to be right for this to work at all: fprintd exits when idle and D-Bus re-activates it, so the env var has to live in the activation file, not in the shell that started it. Otherwise the replacement process has no virtual device and the reader silently disappears.

**Why a container and not the ThinkPad.** The Fedora and Debian branches run `authselect` / `pam-auth-update`, which write a genuine line before the gate is consulted, so a dead file never decides anything there. The gate is only load-bearing on the `Arch | Other` branch, and reaching that means spoofing `/etc/os-release`.

**Result.** Only one file carries the `pam_fprintd.so` line in each run.

| binary | file with the line | outcome | exit | `/etc/irlume/method` | live stack driving a prompt |
|---|---|---|---|---|---|
| shipped 0.7.0 | `common-auth.pam-old` (dead) | `using it` → `face is disabled` | 0 | `fingerprint` | **none** |
| this branch | same dead file | refuses, prints wiring instructions | 1 | absent | none |
| this branch | `zz-b1-live` (live name) | accepts | 0 | `fingerprint` | `zz-b1-live` |

```
== SHIPPED 0.7.0, ONLY a dead file has the line ==
   [fingerprint] found an active pam_fprintd.so line in /etc/pam.d; using it
   [fingerprint] ✓ enabled (fingerprint-only): irlume face is disabled, pam_fprintd drives...
   EXIT=0
   method=fingerprint   live-fprintd-files=[]
```

That is B1 completed on the shipped release: face auth off, method recorded, nothing wired that can drive a prompt. Silently password-only.

The third row is the control. Same bytes, live filename, and the fixed build still accepts, so the refusal in row two is the dot rule and not a broken command.

## Two corrections to my own harness

**Exit codes.** My first run printed `exit=$?` after a pipe into `sed`, so every row reported `0` including the refusal. Captured before the pipe on the re-run; the refusal is exit 1, which matters because failing closed is half the claim.

**`/etc/os-release` is a symlink** to `../usr/lib/os-release`. `cp -a` copied the link and `printf >` wrote through it, clobbering the target, so my restore step silently did nothing and the container still read `ID=probelinux` afterwards. Fixed by replacing the link with a regular file during the spoof and restoring the link after. Verified `ID=ubuntu` at the end.

Issue #167, opened 2026-07-29
Use the camera's own illumination metadata instead of guessing from brightness


archledger:
Found while researching how Windows drives these cameras, after #159.

## The problem

irlume decides which IR frames are illuminated by taking the mean pixel value of a burst and comparing it to a hardcoded threshold of 40/255. That guess is what triggered the destructive extension-unit search fixed in #159: a dark room, an empty chair, or a camera with no IR sensor all produce the same "dark" reading. Measured on the development machine at 38.9 against a threshold of 40, so the trigger sat on a knife edge.

The camera can simply tell us.

## What the hardware already provides

Microsoft's UVC 1.5 extensions define `MetadataId_FrameIllumination`, a 16-byte record appended to the UVC payload header:

```c
struct {
    uint32_t MetadataId;  // 6
    uint32_t Size;        // 16
    uint32_t Flags;       // bit 0 = illumination was on for this frame
    uint32_t Reserved;
};
```

In alternative-frame-illumination mode (D1), which is the mode irlume selects, the camera is required to strobe the illuminator across successive frames and mark every frame with this record. Classifying by brightness is not the contract.

Linux already carries the transport. `V4L2_META_FMT_UVC` (`UVCH`) truncates the header to at most 12 bytes and structurally cannot deliver Microsoft's appended records, which is why `V4L2_META_FMT_UVC_MSXU_1_5` (`UVCM`) exists. Upstream `uvcvideo` also already finds the Microsoft XU, enables metadata production via selector `0x09`, and offers `UVCM` when that succeeds, so irlume must not duplicate that write.

Confirmed on the development machine (kernel 7.1.5):

```
$ v4l2-ctl -d /dev/video3 --list-formats-meta
    [0]: 'UVCH' (UVC Payload Header Metadata)
    [1]: 'UVCM' (UVC MSXU Metadata)

MSXU_CONTROL_METADATA (unit 14, selector 9): GET_CUR = 01, GET_MAX = 01
```

The metadata node is present, the MSXU format is offered, and metadata production is already enabled.

## Shape

Open the metadata node paired with the IR streaming interface, select `UVCM`, stream it alongside the image node, correlate by sequence and the copied PTS/SCR rather than dequeue order, skip the standard UVC header bytes, walk the concatenated little-endian item headers, and read `Flags & 1` from item 6.

[...]

PR #171, opened 2026-07-29
fix(ir): stop writing guessed values to camera extension units (#159) + 0.7.1

### LINKED ISSUE #159: [hw] IR emitter auto-discovery can wedge a SunplusIT 174f:11b4 camera by continuing XU writes after timeout
### Machine and camera

Lenovo ThinkPad P14s Gen 6 AMD
Camera: `174f:11b4`

### Distro and desktop

CachyOS Niri (Wayland)

### Output of `irlume doctor`

```text
❯ irlume doctor
[doctor] platform: Arch-family
[doctor] install origin: pacman package (AUR / makepkg)
[doctor] TPM 2.0: /dev/tpmrm0 ✓
[doctor] Secure Boot: enabled ✓
[doctor] boot mode: UKI (signed PCR-11 eligible)
[doctor] signed PCR policy: systemd PCR-11 signature present ✓; kernel updates won't need re-seal
[doctor] pcrlock: not provisioned: seals use the literal PCR-7 policy + recovery passphrase (re-arm/restore after firmware updates); `systemd-pcrlock make-policy` enables Tier 2
[doctor] camera nodes (classified by pixel format):
  (none found under /dev/video0..9)
[doctor] models:
  loaded by the daemon ✓
[doctor] ORT_DYLIB_PATH: (unset)
[doctor] third-party PAD model: none (default; see `irlume models`)
[doctor] fingerprint reader: Synaptics Sensors ✓ (manage with `irlume fingerprint`)
[doctor] templates (kozika): unknown (daemon not reachable; run `irlume recovery status`)
[doctor] polkit app prompts: not wired (opt-in: sudo irlume login enable --with-polkit --apply)
[doctor] login keyring [gnome-keyring-d]: unlocked ✓ (Secret Service apps like Bitwarden can read secrets after a face login)
```

### Output of `irlume detect` and `irlume version`

```text
❯ irlume detect
partial: installed but not ready (daemon up, not enrolled)
❯ irlume version
irlume 0.7.0
```

### What happened, and what you expected

The camera initially enumerated and worked normally, including its IR stream. After `irlumed` started, it automatically probed UVC Extension Unit controls.
Unit 4 / selector 13 began timing out, but probing continued with additional `SET_CUR` requests. After reboot, the camera could no longer enumerate over USB. A full shutdown and emergency-reset-hole power reset did not recover it.

## Timeline

The camera initially enumerated successfully:

```text
10:43:41 usb 1-1: New USB device found, idVendor=174f, idProduct=11b4, bcdDevice=10.15
10:43:41 usb 1-1: Product: Integrated RGB Camera
10:43:41 usb 1-1: Manufacturer: SunplusIT Inc
10:43:47 uvcvideo 1-1:1.0: Found UVC 1.50 device Integrated RGB Camera (174f:11b4)
10:43:48 uvcvideo 1-1:1.2: Found UVC 1.50 device Integrated RGB Camera (174f:11b4)
```

irlumed then started and selected the camera pair:
```
10:43:57 irlumed: cameras rgb=/dev/video0 ir=/dev/video2 (secure tier)
```

Shortly afterward, the following sequence occurred:

```
10:44:23 uvcvideo 1-1:1.2: Failed to query (SET_CUR) UVC control 13 on unit 4: -110 (exp. 4).
10:44:24 uvcvideo 1-1:1.1: Failed to set UVC probe control : -110 (exp. 48).
10:44:28 uvcvideo 1-1:1.2: Failed to query (SET_CUR) UVC control 13 on unit 4: -110 (exp. 4).
10:44:33 uvcvideo 1-1:1.2: Failed to query (SET_CUR) UVC control 13 on unit 4: -110 (exp. 4).
10:44:39 uvcvideo 1-1:1.2: Failed to query (SET_CUR) UVC control 13 on unit 4: -110 (exp. 4).
10:44:44 uvcvideo 1-1:1.2: Failed to query (SET_CUR) UVC control 13 on unit 4: -110 (exp. 4).
10:44:49 uvcvideo 1-1:1.2: Failed to query (SET_CUR) UVC control 13 on unit 4: -110 (exp. 4).
10:44:54 uvcvideo 1-1:1.2: Failed to query (SET_CUR) UVC control 13 on unit 4: -110 (exp. 4).
10:44:59 uvcvideo 1-1:1.2: Failed to query (GET_LEN) UVC control 14 on unit 4: -110 (exp. 2).
10:45:04 uvcvideo 1-1:1.2: Failed to query (GET_LEN) UVC control 15 on unit 4: -110 (exp. 2).
```

The daemon then reported:
```
10:45:04 irlumed: IR emitter check skipped: hardware: could not auto-enable the IR emitter: no extension-unit control brightened the IR image.
```

On the next boot, the device descriptor could no longer be read:
```
usb 1-1: device descriptor read/64, error -110
usb 1-1: device descriptor read/64, error -110
usb usb1-port1: attempt power cycle
usb 1-1: device descriptor read/8, error -110
usb 1-1: device descriptor read/8, error -110
usb usb1-port1: unable to enumerate USB device
```

On subsequent boots, the camera stopped appearing on the USB bus entirely.

## Correlation with the discovery loop

The seven SET_CUR timeouts on unit 4 / selector 13 appear to match:

- six payloads returned by candidate_payloads()
- one attempted restoration of the original value

The loop then continued to selectors 14 and 15 despite the earlier ETIMEDOUT.

Relevant code:

- crates/irlume-daemon/src/main.rs: unconditional startup call to ensure_ir_emitter()
- crates/irlume-camera/src/lib.rs: dark-frame threshold triggers setup_ir_emitter()
- crates/irlume-camera/src/ir_emitter.rs: enumeration of units 0..=31 and selectors 0..=15
- xu_query() reduces all ioctl failures to bool, so ETIMEDOUT, EPIPE, and ENODEV cannot be distinguished
- probing continues after timeout
- restoration failures are ignored

The camera already controls its IR emitters automatically when the IR stream is opened. It may have been incorrectly classified as needing an emitter
control because the measured frame mean was below IR_LIT_MEAN.

## Expected behavior

Automatic startup should not send arbitrary vendor-specific SET_CUR payloads to an unknown camera.

At minimum:

1. Emitter auto-discovery should require explicit user confirmation.
2. Unknown devices should default to emitter control disabled.
3. Discovery should use a VID:PID allowlist.
4. The first timeout or device error should immediately abort all probing.
5. A control should not be modified unless its current value was read successfully.
6. Restoration failure should be treated as fatal.
7. GET_INFO should be checked before attempting SET_CUR.
8. Generic "brightness boost" probing should not write arbitrary values to unknown controls.

## Impact

This was not merely a temporary capture failure. The camera became unavailable at USB-enumeration level and could not be recovered with a complete shutdown
or the laptop's emergency reset hole. The laptop now requires another hardware service visit.

archledger:
Closes #159. Cuts 0.7.1.

## What happened

irlume searched for the control that lights a Windows-Hello illuminator by writing guessed payloads to every UVC extension unit from 0 to 31 and every selector from 0 to 15, until the infrared image got brighter. On a Lenovo ThinkPad camera (USB `174f:11b4`) those writes reached an undocumented vendor unit, the camera stopped answering, and by the next boot it no longer enumerated on the USB bus. Neither a full shutdown nor the laptop's emergency reset hole recovered it.

Nobody asked for the search. From 0.1.0 enrolling a face could start it; from 0.3.0 so could every daemon start. The trigger was a mean infrared brightness below a threshold, which is also what an unlit room or an empty chair looks like. Measured on the development machine at 38.9 against a threshold of 40.

## What changed

The search is gone. Nothing guesses.

Everything needed to avoid this was already in the camera's USB descriptor. Discovery now resolves identity from the file descriptor that will receive the write, considers only the unit whose GUID is Microsoft's documented camera-control extension and only if there is exactly one, touches only selectors that unit advertises, requires `GET_INFO` to permit the write, and builds the value from the camera's own answers. A change is only believed if the picture brightens when the control is set and dims again when it is put back. Nothing is written whose previous value could not be read first, and the first failed request ends the operation.

Discovery runs only on `irlume ir-setup`, never on a daemon start or an enrolment.

Upgrading also stops writes that were already happening: the config now records which camera a control was found on and stores no payload, so pre-0.7.1 files are refused.

## Validation

Three cameras, three vendors, three distributions.

| | writes | outcome |
|---|---|---|
| ASUS `3277:0059`, Fedora | 4 × `01 03 02` during a real authentication | `archledger, profile 'BEN', score 0.722` |
| NexiGo `3443:c803`, Arch | table value and runtime-derived value byte-identical | both light the emitter |
| Chicony `04f2:b7bf`, Ubuntu, RGB-only | **0** | old code would have sent 119 guessed writes to 17 responding controls |

Stress: 12 concurrent identifies, 12 daemon restarts, 10 `ir-setup` runs, 114s continuous capture, across two machines. `NRestarts 0`, zero watchdog kills, zero LSM denials. Every byte sent to a camera in the entire run was either the intended value or a restore read back from the device.

The descriptor parser is tested against 1301 bytes captured from a real camera, including a module exposing two independent VideoControl functions, which is why parsing is scoped to one interface: a unit number alone is not an address.

Eight adversarial review rounds. The last says ship.

## Also in this release

`scripts/diagnose-missing-camera.sh`, which collects evidence from an affected machine and sends the camera nothing. Install-time messages on the Debian, PPA and Arch lanes no longer claim enrolment configures the emitter. `IRLUME_LOG_EMITTER_WRITES=1` prints every emitter write.

Riding along, previously unreleased: #150, #151, #153, #154, #156, #157.


archledger:
Good catch on the message, fixed.

You are right that reporting "camera stopped responding" when the errno was `Unsupported` is untrue, and that error text is exactly what someone reads when their camera is not working. `Unresponsive` now renders two different sentences: a refusal says the unit refused a request its own descriptor claims it implements, and only a genuine non-answer says the camera stopped responding. Test added so the two cannot drift back together.

On continuing to the next selector after an `Unsupported` query: keeping the stop is deliberate, and I would rather explain than quietly disagree.

Only selectors the camera advertises in `bmControls` are reached at all. A refusal there is the camera contradicting the descriptor it just published, which is not a state to keep issuing requests in. The errno also cannot carry the weight the fallthrough would need: uvcvideo maps the device's "wrong state" to `EACCES` and "invalid unit, control or request, or a hardware error" to `EIO`, so a polite "not now" and a device in trouble are not cleanly separable from userspace. An earlier revision of this branch classified those two errnos in exactly the wrong direction.

The cost of stopping is that a camera advertising both IR Torch and Face Authentication, where the first refuses, will not reach the second in one run. The cost of continuing is that irlume keeps talking to a camera that has started behaving unexpectedly, which is the mechanism that destroyed a reporter's camera in #159 and the reason this PR exists. For this release I would rather pay the first.

Worth revisiting once irlume reads the camera's own per-frame illumination metadata (#167), which removes the guesswork this path is built on.

PR #184, opened 2026-07-31
fix(camera): follow the documented face-auth lifecycle, set once and unset after

### LINKED ISSUE #168: Follow Microsoft's face-auth lifecycle: set before streaming, unset after
Found while researching how Windows drives these cameras, after #159.

## What Windows does

Microsoft's HLK qualifying sequence for a Hello camera is:

1. identify the Hello-capable IR stream
2. select the profile media type
3. determine whether the stream supports alternative illumination or background subtraction
4. **set** the face-auth property
5. start streaming
6. validate frames and metadata
7. stop streaming
8. **unset** the property

Microsoft explicitly tests that the control is unset after streaming, and requires a general-purpose interface to default to D0. Both cameras validated here report `GET_DEF 01 03 01`, which is D0, matching that requirement.

## What irlume does

Two departures:

**It never unsets.** irlume sets the face-auth mode and leaves it. On the ASUS module the control was observed back at `01 03 01` once streaming stopped, so that camera reverts on its own, but the NexiGo was observed at `01 03 02` outside a capture. The behaviour is camera-dependent, and relying on a camera to undo something irlume did is not a design. D0 should be restored on normal completion, failure, cancellation, timeout, and daemon shutdown.

**It rewrites the control mid-burst.** `capture_ir` re-fires the emitter control halfway through a burst, on the theory that the control may self-clear. That is not part of the documented sequence, which sets the mode once before the stream starts. There is no public guarantee that changing the mode during streaming is frame-atomic or has a defined first affected frame. Measured on a real authentication: four `SET_CUR` writes per authentication where the documented sequence implies one per stream.

Neither is dangerous in the way #159 was. Both write a validated value to a documented control on a unit the descriptor confirms. But writing to camera firmware more often than the hardware's own specification calls for is not a habit this project should keep.

## Shape

Set once immediately before starting the stream. Do not rewrite mid-burst. Restore D0 when the stream stops, including on every error path. If frames do not conform, stop the stream, restore D0, and start a fresh session rather than rewriting underneath a running stream.

Worth doing alongside #167, since illumination metadata removes the reason the mid-burst re-fire was added: it exists to make sure a lit frame lands in the burst, which metadata answers directly.

archledger:
Closes #168.

## What Microsoft documents

The [camera driver bring up guide](https://learn.microsoft.com/windows-hardware/drivers/stream/windows-hello-camera-driver-bring-up-guide) gives the qualifying sequence for a Hello camera: select the stream, **set** the face-auth property, start streaming, validate, **stop streaming**, **unset the KS control**. Steps 4 and 5 are spelled out, and the HLK suite tests them.

irlume did the first four and left the control set. It also rewrote it underneath the running stream, which is not part of the sequence at all.

## Two departures, both removed

**It rewrote the control during streaming.** Three places did it: `capture_ir_streaming` and `capture_ir_sequence` re-applied every eighth frame, and `capture_with_stats` re-applied halfway through each burst. At the default consent budget of 80 frames the first is ten extra writes per watch on its own.

The cost was not only firmware writes. `enable` is not a bare ioctl: every call re-reads the camera's USB descriptors from sysfs and takes a lock to scan the undo journal on disk, so those ten sat in the authentication path as ten descriptor reads and ten journal scans too.

The justification was a comment reading "some controls self-clear". That is a hypothesis, and since #167 the camera answers it directly, per frame, through illumination metadata; `capture_with_stats` already picks its frame with `brightest_lit` rather than by guessing.

**It never unset.** `StreamMode` now holds the mode for one stream and restores it on drop, covering the paths no statement does: an error taken by `?`, a panic in the decoder, a cancelled request. Same reason the undo record in #183 restores from `Drop`.

What gets written back is the camera's own `GET_DEF`, read before anything is applied, rather than a constructed "off" value. For a dual-purpose interface the specification requires that default to carry D0, general purpose; for a face-auth-only interface it is whatever that camera says it starts in. A camera whose default cannot be read is not driven at all, because there would be no way back.

## Measured on hardware

NexiGo HelloCam N930W, Microsoft XU unit 4 selector 6, same workload both sides:

| | total `SET_CUR` | applied | restored |
|---|---|---|---|
| `origin/main` | 14 | **14** | **0** |
| this branch | 14 | **7** | **7** |

Every mode set for a stream is put back when that stream ends. The total is unchanged because the restore is a second write by design; what went away is the re-firing that scaled with frames rather than streams.

`scripts/emitter-selfclear-test.sh` asserts the pairing rather than a threshold, because a flat count cannot tell "twice per stream" from "once every eighth frame". **It fails against `origin/main` on two assertions and passes here**, which is the only reason to believe it. It does NOT confirm that the value survived while the stream was running. An earlier version of this description said it did, and that was wrong: the script read the control back after killing the daemon, which drops the guard and restores the default, so it was inspecting the post-restore state. Its section 3 now reports itself as not exercised and names what would answer it, a `GET_CUR` taken from inside a live stream.

## Review notes

`StreamMode` is `#[must_use]`. That is how the six remaining call sites were found rather than assumed: discarding the guard restores immediately and leaves the stream unlit, so the compiler had to be made to say so.

On a stream restart the guard is disarmed before the new one is armed. Reassigning without that would evaluate the new `enable` and then drop the OLD guard, writing the default straight over the value just set.

One test came from a surviving mutant: `Drop` checks `armed` before calling `restore`, so the identical check inside `restore` is unreachable through a drop and nothing exercised it. `restoring_twice_writes_once` covers it.

## What this does NOT prove

- **The ASUS module could not exercise the guard.** In this room its IR mean sits at 144 and discovery correctly reports no brightness lift, so it never enables an emitter to restore. The write-count halving was measured on it; the set/restore pairing was measured on the NexiGo only.
- **IR Torch (`0x0A`) is untouched here.** Its "off" is the D0 bit in `dwMode` rather than `GET_DEF`, which the specification requires to be an active mode, so it needs different restore semantics. No camera here advertises it.
- The three capture paths still open `Device::with_path` and `SafeStream` directly rather than going through `IrCamera`/`IrSession`. Consolidating them is worth doing and is deliberately not in this PR, so a regression here stays attributable.
- **A crash can defeat this lifecycle permanently, and that is tracked in #188.** `StreamMode` decides whether it may restore by asking whether the control still holds what irlume applied, and that question cannot separate "another tool set this" from "irlume set this and was killed". This PR takes the conservative side: a value found already in place is left alone. So a `SIGKILL` mid-capture leaves the mode applied, every later session reads it as somebody else's, and it is never put back. One crash outlives arbitrarily many clean captures.

  Both directions are wrong without a record of what irlume itself wrote, which is why the fix is its own change rather than another commit here. Landing this first is deliberate: the lifecycle is a strict improvement on `main`, where the mode was set on every session and never restored at all, and #188 removes the remaining case.
- **Stream writes are not journalled.** `StreamMode` keeps its undo data in memory only, so a `SIGKILL`, a watchdog kill or a power loss mid-capture leaves the mode applied with nothing on disk describing it, and `recover_pending_write` finds nothing because only the discovery path writes records.

  This is a deliberate trade, not an oversight, and it is raised here because two review rounds found it. Journalling every capture means a durable write with an `fsync` on the authentication path, once per unlock, against a window that ends when the process does. It is also not a regression: before this PR irlume set the mode on every session and never restored it at all, so the crash-time state is what shipped in 0.7.2 and the in-process paths are strictly better. Closing it properly means a per-camera record with a lifetime shorter than a capture, which is a different design than #183's and belongs in its own change.
- Draft until the full mutation suite, ASan and a Codex round have run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,7 @@ SHA256SUMS.asc
 # Bytecode from the scripts/*.py tooling
 __pycache__/
 *.pyc
+
+# Mutation-harness backups. One was committed by a `git add -A` run while the
+# harness had files out, which put a 1938-line copy of a source file in a commit.
+*.rs.bak
--- a/crates/irlume-camera/src/emitter_journal.rs
+++ b/crates/irlume-camera/src/emitter_journal.rs
@@ -419,7 +419,9 @@ pub(crate) fn record_applies(record: &PendingWrite, id: &CameraIdentity) -> Resu
             supported: SCHEMA_VERSION,
         });
     }
-
+    if record.descriptor_sha256 != fingerprint(id) {
+        return Err(Mismatch::DifferentCamera);
+    }
     // The digest already pins the descriptors, and these fields come from the
     // same sysfs read, so a disagreement means the record was assembled from two
     // different cameras or edited by hand. Either way it is not a description of
--- a/crates/irlume-camera/src/emitter_journal.rs
+++ b/crates/irlume-camera/src/emitter_journal.rs
@@ -419,9 +419,7 @@ pub(crate) fn record_applies(record: &PendingWrite, id: &CameraIdentity) -> Resu
             supported: SCHEMA_VERSION,
         });
     }
-    if record.descriptor_sha256 != fingerprint(id) {
-        return Err(Mismatch::DifferentCamera);
-    }
+
     // The digest already pins the descriptors, and these fields come from the
     // same sysfs read, so a disagreement means the record was assembled from two
     // different cameras or edited by hand. Either way it is not a description of
--- a/crates/irlume-camera/src/ir_emitter.rs
+++ b/crates/irlume-camera/src/ir_emitter.rs
@@ -946,12 +946,17 @@ impl StreamMode {
     /// Give up the restore without writing anything.
     ///
     /// For one case only: the stream this guard was set for has been torn down
-    /// and reopened. These modules reset the control per open, so there is
-    /// nothing left to put back for the old stream, and a fresh guard covers the
-    /// new one. Restoring here instead would be a write to a control the reopen
-    /// already reset, and reassigning without it would drop the OLD guard after
-    /// the new `enable` had run, writing the default straight over the value
-    /// just set.
+    /// and reopened, and a FRESH guard has taken over the same control.
+    ///
+    /// Without this, reassigning would drop the old guard after the new `enable`
+    /// had already run, writing the default straight over the value just set.
+    ///
+    /// An earlier version of this comment justified it with "these modules reset
+    /// the control per open". That is not what this project measured: the record
+    /// in `IrCamera::session` says the control survives stream close and process
+    /// exit on both cameras here. So the caller gives the old guard up only when
+    /// the replacement is armed, and an unarmed replacement leaves the old guard
+    /// to put the value back.
     pub fn disarm(&mut self) {
         self.armed = false;
     }
@@ -1274,10 +1279,15 @@ fn override_memo() -> &'static OverrideMemo {
 /// camera per process, and only with the evidence every other write here
 /// requires.
 ///
-/// The record bounds the WRITES. It is not used as an answer about the camera's
-/// present state: a remembered success re-reads `GET_CUR` and reports whether
-/// the control still holds the payload, because "we wrote this once" and "this
-/// is set now" stop being the same statement the moment a camera resets.
+/// The record bounds the CHECKS, not the writes. A remembered success means this
+/// payload was already validated against this camera's descriptors, so the
+/// checks and the refusal message do not run again; the value is still applied
+/// for each new stream, because the mode is restored when the previous one
+/// ended and the control is sitting at the camera's default.
+///
+/// It used to answer from a `GET_CUR` read-back instead, which was right while
+/// irlume left the mode set for the life of the process. Under a per-stream
+/// lifecycle that reported every stream after the first as dark.
 ///
 /// The override is still an escape hatch: the unit may be a vendor's, and the
 /// bytes are the person's rather than the camera's own `GET_DEF`. What it no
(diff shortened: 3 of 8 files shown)


archledger:
Pausing here with the branch in a known state, so picking this up later does not need re-derivation.

## Done

Five commits. The mid-stream re-firing is gone from all three places it lived, and `StreamMode` restores the camera's own `GET_DEF` when a stream ends. Three review defects are fixed, two of which this branch introduced: locals drop in reverse declaration order so the restore was firing before `STREAMOFF`; the override memo turned the per-stream restore into a one-unlock daemon; and disarming unconditionally on a stream restart could strand a set control with nothing to put it back.

Measured, NexiGo and ASUS: `main` 14 applied / 0 restored against 7 / 7 here, and the override now applies on every stream rather than only the first (1 across three requests before the memo fix, 7 after). 175 tests, clippy clean, 58/58 mutants, 873 tests under ASan with no unsuppressed leak.

## Still open, in order

1. **`IrCamera::session` sets the mode after `STREAMON`.** `warm_up_stream`'s first dequeue starts the stream, and `enable` runs after it, so the normal authentication path still writes under a running stream. Pre-existing rather than introduced here, but this PR claims to fix the lifecycle, so it belongs in it. Needs a hardware pass to confirm the emitter still lights when the mode is set before the first dequeue.
2. **Do not arm the guard when irlume did not change anything.** `check_and_apply_override` returns `Ok(true)` without writing when the control already holds the payload, so the stream end clears state irlume never set, against the principle established in #183. On IR Torch, `intended_value` derives from `GET_DEF`, so the applied bytes equal the restore bytes and every stream end writes an identical value.
3. Another review round, then out of draft.

## Deliberate gaps

`emitter-selfclear-test.sh` section 3 reports itself as not exercised rather than pretending. Answering "did the value survive while the stream was running" needs a `GET_CUR` taken from inside a live stream, before `STREAMOFF` and before the guard runs, which belongs in `crates/irlume-camera/examples/ir_refire_probe.rs` since that already owns one uninterrupted stream.

The three capture paths still open `Device::with_path` and `SafeStream` directly instead of going through `IrCamera`/`IrSession`. Consolidating them is worth doing as a separate PR, so a regression there stays attributable.

archledger:
Review has now found the same gap from two directions, and they are one problem rather than two.

## The two findings are the same question

**No record of stream writes.** A kill mid-capture leaves the mode applied with nothing on disk describing it.

**A crash-created mode is never restored.** The "do not claim a value irlume did not set" rule uses `current == wanted`, which cannot tell *another tool set this* from *irlume set this and was killed*. After a crash the value is still there, every later session reads it as already held, stays unarmed, and never puts it back. One crash makes the mode permanent.

Those are the same missing fact: **whether irlume put the value there**. Without it, any policy is wrong in one direction. Claiming always means clearing another writer's state, which is what the earlier round objected to. Never claiming means irlume's own leftovers survive forever, which is what this round objects to. Both objections are correct.

## Correcting the cost I gave for fixing it

I argued against recording stream writes on the grounds that it means an `fsync` per unlock on the authentication path. That reasoning was wrong, and it was wrong in a way that matters.

The event this needs to survive is PROCESS death: a `SIGKILL`, the watchdog, a panic. The page cache survives all of those, so a plain write with no `fsync` is enough to let the next process read the record. Durability against power loss is what #183 needed, because a control changed by discovery must be recoverable across a reboot; a capture-scoped record does not carry that requirement, because a power cut also takes the camera's own state with it on any module whose control does not survive one, and where it does survive, the same record is still readable after boot.

So the honest comparison is a `write` and a `rename` per stream, not an `fsync` per unlock. That is a different decision and I should have separated the two durability questions before dismissing it.

## What this means for the PR

I am not going to fix this by picking one side of the tension, because both sides are defects. The shape that resolves it is a per-stream record holding the coordinates and the applied bytes, written before the write and removed after the restore, consulted when the control is found already holding the wanted value:

- already held **and** a record from this camera exists: irlume's own leftover, so claim it and restore
- already held **and** no record: somebody else's, so leave it alone

That is a real design change with its own tests and its own hardware validation, and putting it in a branch that has already been through eight review rounds would make the regression surface harder to attribute rather than easier.

Unless there is an objection, I would rather land the lifecycle work here with the limitation stated plainly, and take the record into its own change immediately after. What must not happen is this PR merging with the limitation undocumented, which is the state it was in when review found it twice.

Issue #101, opened 2026-07-26
Credential-release nod gate: no signal for when to nod, so a single nod misses the window


archledger:
Measured on real hardware (Zenbook, ASUS FHD built-in, branch daemon, real enrolment, driving the actual `plasmalogin` PAM stack) on 2026-07-25. Three trials:

| trial | user did | result |
|---|---|---|
| A | held still | refused: `denied (live=true, score 0.9870: nod your head to approve)` |
| B | nodded once, at the start of the attempt | refused, same message, score 0.9275 |
| C | nodded continuously until the result printed | `password unsealed`, 6s total |

Detection works. Trial B is the problem: **the user has no signal for when to nod.**

The gesture watch opens only after the face capture and the match complete, several seconds into the attempt. A single nod made when the prompt appears lands before anything is watching, so it is discarded and the window then times out. From the user's side this is indistinguishable from the gate being broken: the face matches, the login proceeds, and the keyring password silently never gets released.

This fails closed, so it is not a security hole. It is a usability defect with a likely support cost: the natural bug report is "face login works but my keyring never unlocks", with the user having done nothing wrong.

## Options

1. **Start the gesture watch earlier**, so any nod during the attempt counts rather than only one made after the match. The frames are already streaming, so this costs no extra time. Most robust, and needs review because it changes when the gate's evidence is collected relative to the match.
2. **Emit the instruction when the watch actually opens** rather than before the capture. Harder: the daemon cannot speak to the PAM conversation mid-flow.
3. **Say "keep nodding"** in the greeter text. Cheapest, and honest about the current behaviour, but it asks the user to work around the timing rather than fixing it.

Option 1 looks right, with option 3 as an immediate mitigation.

## Note on scope

The gate itself behaved correctly throughout: it refused release without a gesture while the face scored 0.99, and login still succeeded on the face match with only the sealed password withheld, which is the intended split.

archledger:
Option 1 is implemented and it fixes the reported case, but measuring it turned up something worse underneath. Not pushed.

## The fix works

The watch now runs once before the first capture, taking a third of the same budget rather than adding to it, and runs per authentication rather than per grace-window retry.

| trial | user did | result |
|---|---|---|
| nod once, early, then hold still | the case from the report | `pre-match watch saw the gesture` → released, 6s |

That is the exact scenario that used to be refused.

## The problem it exposed

[...]

Issue #25, opened 2026-07-19
Landmark-anchored IR relief depth check (local geometry vs global center-edge ratio)


archledger:
Splitting this out of #4 so it isn't lost in a closed thread. Credit to @Magniquick for the idea.

**Problem.** The depth gate uses a single global `ir_center_edge_ratio`. It can't separate a real 3D face from a flat sheet tilted so its 1/r^2 IR falloff mimics a face, which is the overlap the self-test measured against the vinyl banner (banner 1.02-1.58 vs genuine 1.37-1.40).

**Idea.** Sample IR brightness at FaceMesh landmarks irlume already has (nose tip, eye sockets, cheeks) and check the pattern against real facial relief: nose brightest and convex, eye sockets recessed, cheeks between. A flat print can fake a monotonic center-to-edge ramp but can't reproduce nose-brighter-than-cheek-brighter-than-eye-socket. Stays fully algorithmic, no dataset. It composes with ambient subtraction (`IRLUME_IR_AMBIENT_SUBTRACT`): subtract the emitter-off frame first, then test the residual against the landmark geometry.

**Starting tooling, already in the repo.** `landmark_dump` (example in `crates/irlume-auth`, walkthrough in `docs/DEVELOPMENT.md`) captures a raw IR strobe burst and writes, per frame, the raw PGM plus `idx,x,y,brightness` for all 478 landmarks. So the relief pattern can be measured on real captures before any gate code is written.

**Scope and constraints.**
- Deny-only, like the rest of the liveness stack: a relief mismatch may reject, never approve what the physics gate rejected.
- Has to hold up in the same envelope the current gate is measured in (dim/night IR-emitter-only, direct-sun saturation). Numbers land in `docs/pad-results` or the self-test, same standard as everything else.
- Per-user calibration is acceptable; the existing depth floor is already per-enrollment.

**Status:** no code yet, tooling in place, open to anyone.

archledger:
Evidence from #101 that raises the value of this issue, and narrows what it has
to beat.

Measured 2026-07-27 on one camera and one user, driving the real consent path in
three conditions and recording the raw pose trajectories: sitting still, a
hand-held printed face, and deliberate nodding.

| feature | still | nod | hand-held print |
|---|---|---|---|
| pitch range | 0.0132-0.0521 | 0.0747-0.1110 | 0.0243-**0.0916** |
[...]

PR #205, opened 2026-08-01
Record #101's candidate discriminator with the consent evidence, gating nothing

archledger:
Advances #101 and deliberately does not close it. Based on `main`, independent of the other open PRs.

## Where #101 stands, per its own thread

The original complaint (no signal for when to nod) is fixed by the pre-match watch, the "keep nodding" wording, #136, and #140. Print resistance moved to #25 after measurement showed a hand-held print physically performs the gesture. What keeps #101 open is consent without intent: the gate fired on a still, non-consenting user 2 times in 18.

The thread measured a candidate discriminator for that case. Mean per-frame pitch step separated still from nod by 2.3x (still at most 0.0064, nods at least 0.0149) where the gating `pitch_range` manages 1.44x. The thread then declined to ship a threshold from one user's single session, citing measured cross-session drift: genuine nods at 0.057 in one campaign against 0.082 weakest in another. The recorded blocker is more subjects and more sessions.

## What this PR is

The collection path for that data, not the threshold.

`NodEvidence` gains `mean_step`, computed over consecutive frame pairs: both pitch readings must exist, be finite, and sit at `idx` values exactly one apart. The review round showed the original adjacency rule was positional, so two samples left neighbouring by an upstream drop (idx 10 beside idx 12) read a two-frame move as one frame's motion; pairing now keys on the recorded frame index and a sparse series measures 0.0. The consent evidence debug line carries the value on accepts and refusals under `IRLUME_LOG=debug`, the same opt-in as every other diagnostic line; ordinary authentication runs compute and discard it, and the docs now say that instead of claiming every watch records it. `blinkcap replay` summarizes it per label, so existing recorded corpora count as sessions too.

Nothing reads the field in a verdict. Its docs quote the one-session populations as provenance, state they are not a contract, and name the threshold decision as #101's to make on the accumulated data.

## What the review round did to replay (`b490002`)

The corpus is only as honest as its reader and writer, and the round found both lying in the tool this PR turns into the collection instrument:

- Replay marked a file as a session the moment its header parsed, silently dropped unparseable record lines, and never checked the header's declared `frames` count. A crash-truncated capture replayed as `mean_step 0.0000` and sat in the per-label minimum exactly like a still head. The body is now parsed strictly, the count is enforced, and any damage fails the whole replay loudly instead of shrinking the corpus.
- `replay <file.jsonl>` is in the usage line but never reached the pose detector (directory-only), and a pose-only directory printed its summary then exited FAILURE. Both forms now work and a pose-only corpus exits 0.
- Captures were written straight to the destination, so a crash mid-write could leave the valid-header-over-partial-body shape the reader now refuses. Captures stage to a `.tmp` file, sync, and rename into place.

## Testing

Liveness unit tests pin the arithmetic (a known series averages to 0.02), the `None` gap rule, the sparse-index rule (idx 10 next to idx 12 yields 0.0), overflow safety at `idx == usize::MAX`, the no-pairs cases (0.0, never NaN), and relative ordering on the gate's own model shapes (nod above still), with no absolute floor asserted. Black-box tests spawn the real binary: a single pose file and a pose-only directory both exit 0 with the exact `mean_step 0.0200..0.0200` figure in the summary; a header-only truncated file and a malformed record both exit 1 naming the damage, with no summary printed over a damaged corpus.

Five hand-applied mutants each made exactly one of those tests fail: positional pairing restored, the count check disabled, the silent record skip restored, the pose-only SUCCESS exit removed, and the single-file arm removed. The existing detector-versus-evidence agreement test is unchanged and green, so verdicts are provably untouched.

- [x] `cargo test`: `irlume-liveness` 37, `irlume-cli` all six targets green on `b490002`
- [x] clippy `-D warnings`, `fmt --check`
- [x] CHANGELOG updated, including the debug-gated recording contract

## What the next hardware session gets

Run consent watches with `IRLUME_LOG=debug`, or `blinkcap replay` over old recordings, and the per-label `mean_step` ranges land in the output beside the accept rates: the cross-session spread #101 needs before `mean_step` can move from recorded to gating.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ All notable changes to irlume are documented here. This project adheres to
 
 ### Added
 
+- **The consent watch records the #101 candidate discriminator on every run.**
+  `mean_step` — mean absolute pitch change per adjacent frame pair, with
+  face-lost gaps contributing nothing — now rides in the nod evidence, the
+  consent debug line, and `blinkcap replay`'s per-label summary. It gates
+  nothing: #101 measured it separating a still head from a deliberate nod by
+  2.3x where the gating pitch range manages 1.44x, then deliberately declined
+  to set a threshold on one user's single session, because this class of
+  signal is known to drift between sessions. The blocker is cross-session
+  data; recording the metric with the rest of the evidence means every consent
+  watch and every replayed capture accumulates it, instead of only runs where
+  someone remembered the pose-series dump flag.
+
 - **A dark or blinded IR capture reports what its evidence supports.** A dark
   burst used to get one hint ("no active emitter; run `sudo irlume ir-setup`")
   even though shutters, covers, range, exposure and emitter failures all
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,17 +7,22 @@ All notable changes to irlume are documented here. This project adheres to
 
 ### Added
 
-- **The consent watch records the #101 candidate discriminator on every run.**
-  `mean_step` (mean absolute pitch change per adjacent frame pair, with
-  face-lost gaps contributing nothing) now rides in the nod evidence, the
-  consent debug line, and `blinkcap replay`'s per-label summary. It gates
-  nothing: #101 measured it separating a still head from a deliberate nod by
-  2.3x where the gating pitch range manages 1.44x, then deliberately declined
-  to set a threshold on one user's single session, because this class of
-  signal is known to drift between sessions. The blocker is cross-session
-  data; recording the metric with the rest of the evidence means every consent
-  watch and every replayed capture accumulates it, instead of only runs where
-  someone remembered the pose-series dump flag.
+- **Debug-enabled consent watches report the #101 candidate discriminator.**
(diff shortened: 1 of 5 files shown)


archledger:
The hardware pass found a defect the review round's own prescription introduced, now fixed in `280eeb9`.

IR modules strobe alternate frames: a live 75-frame take on the ASUS had 38 usable samples, all 37 usable pairs at index gap 2 (verified from the recording itself). The gap-1-only rule therefore read every real capture as `mean_step 0.0000`, including an accepted nod: structurally zero on exactly the cameras the metric measures. Pairs now count at gap 1 or 2 with |Δpitch| divided by the gap, so a strobe pair is a per-frame rate and no gap can inflate the metric (the round's original concern); longer gaps contribute nothing.

Live corpora through the fixed binary: nod `0.0116`, still `0.0018`, a 6.4x separation, exit 0 for both single-file and directory replay. The figures match an independent recalculation of the same recordings outside irlume. Three mutants (normalization dropped, gap-1-only restored, any-gap allowed) each fail exactly one new assertion.

PR #206, opened 2026-08-01
Honour the Windows exclusive-control model at the emitter write; state the rest

### LINKED ISSUE #169: Divergences from the Windows camera contract these modules are built to
Research into how Windows 11 drives these cameras, after #159. Separate from #167 (use the camera's illumination metadata) and #168 (set once before streaming, unset after), which came out of the same work.

These modules are designed and certified against Windows, so the Windows contract is the closest thing to a specification for the hardware. Three further places irlume diverges.

## Exclusive control while another application is streaming

Windows permits several frame consumers but allows only one of them to modify device configuration; shared consumers are read-only and inherit the controlling application's settings. irlume changes an extension-unit control whenever it captures, without regard for whether something else currently owns the camera.

In practice irlume takes the camera exclusively, so this is mostly theoretical today, but the ownership model is worth honouring rather than relying on that.

## Privacy indication for infrared-only capture

Windows hardware certification ties visible privacy indication to sensor capture, and says so explicitly for infrared-only capture with an illuminator the user cannot see. irlume can illuminate a person with 850nm light during authentication. On most laptops the indicator is wired to sensor power and will follow, but that is the hardware's doing, not irlume's, and it is not something the project currently reasons about at all.

Worth stating a position in the threat model even if no code changes: what the user is told is happening, and when.

## Control state across power transitions

Windows powers integrated camera components fully off in D3cold and preserves no control context. It defines an optional cache protocol for restoring a specific list of ordinary UVC controls, and replaying arbitrary extension-unit state after resume is not part of it.

irlume re-applies the emitter control at every capture, which is a reasonable response to the same power behaviour, but it is worth being deliberate about rather than incidental. Overlaps with #168.

## Context worth recording

Two things Linux cannot reproduce, and should not pretend to:

- Windows routes Hello controls through a vendor DeviceMFT that can translate or refuse them. Writing an extension unit directly bypasses whatever policy that component implements. There is no Linux equivalent.
- Enhanced Sign-in Security is not "the same camera with a better application". It requires supported camera firmware, the inbox driver, compatible chipsets, secure device enumeration, VBS-protected memory, and an OEM-provided ACPI SDEV table. Opening the same IR node on Linux does not approximate that trust path, and documentation should not imply otherwise.

## Where Windows and irlume now agree

Worth recording, since it was not true before 0.7.1: Windows' own UVC driver probes only descriptor-advertised controls and disables a control whose initialization queries fail, rather than sweeping unit and selector combinations or continuing to use a selector that failed. That is the model 0.7.1 adopted.

## Not established

Microsoft does not publish how many frames Hello consumes per authentication, its recognition window, or its stopping threshold. Nothing here should be read as a claim about those.

archledger:
Closes #169. Based on `main`, independent of the other open PRs. Each claim in the issue was re-verified against the Microsoft sources before acting; the citations live in the threat model.

## Exclusive control, the one code change

Verified: Windows permits many frame consumers but exactly one controlling instance. Sharing consumers must use the controlling application's media type and "cannot change KSPROPERTYSETID_ExtendedCameraControl controls", the class Windows routes Hello's camera control through. The same page does let sharing consumers change legacy `VIDEOPROCAMP`-style and vendor-specific controls, so the stand-down below is deliberately stricter than the Windows rule ([MediaCaptureSharingMode](https://learn.microsoft.com/uwp/api/windows.media.capture.mediacaptureinitializationsettings.sharingmode), [MF FrameServer share modes](https://learn.microsoft.com/windows/win32/medfound/mf-devsource-attribute-frameserver-share-mode)).

`ir_emitter::enable` now scans `/proc` for other processes holding the target node and stands down from the control write when one exists, returning the existing inert guard. The foreign application's configuration is left untouched and the capture proceeds or fails on its own terms.

Properties, each with its reason in the code:

- Fail-safe direction: an unlit emitter degrades one capture toward the password; a write under a foreign owner mutates a stream that application is using. The previously reachable outcome in this scenario was an EBUSY failure.
- Only positive evidence of a foreign consumer stands the emitter down, and a scan that could not see everything says so. The review round found the original scan silently failing open in production: reading another process's `/proc/<pid>/fd` is gated by `PTRACE_MODE_READ_FSCREDS` (proc_pid_fd(5), ptrace(2)), the packaged unit keeps only `CAP_DAC_OVERRIDE` and `CAP_FOWNER`, so every cross-uid holder (PipeWire, a browser) was invisible and `.is_ok_and` collapsed the denial into "no consumer". The scan now returns `ConsumerScan { consumers, permission_denied }`; a permission denial marks the scan incomplete, the daemon logs that degradation once per process and proceeds, and the threat model states the limit. Standing down on the blind spot instead would make every packaged capture inert. Granting `CAP_SYS_PTRACE` or adding a separately confined observer is #207. A missing `/proc` or an fd-less pid dir is ordinary churn, not a blind spot, and stays silent.
- Racy by nature and documented as such: this honors a sharing contract for consumers that exist at decision time. It is not mutual exclusion; the kernel lock keeps that job for irlume's own processes.
- Self pid excluded, since the daemon holds the node when `enable` runs.

Tested against a constructed proc tree: detection with comm names, self-exclusion, non-pid entries, fd-less dirs, missing root, empty device path. Also probed live against the real `/proc` of the test container: a process holding a file is reported with its comm name, and the report empties when it exits.

## Privacy indication and D3cold, positions in the threat model

The threat model gains a "Windows camera contract" section:

- IR-only privacy indication ([camera privacy controls](https://learn.microsoft.com/windows-hardware/drivers/stream/camera-privacy-controls)): certification ties visible indication to sensor capture including IR-only streams. A visible-wavelength 850 nm illuminator may itself be the indicator (it glows faint red); an invisible 940 nm one requires a separate LED. The obligation is phrased against the module's ISP capturing sensor data, not against any host application, so it follows irlume's captures without irlume's involvement. irlume's stated position: it never suppresses indication, adds no software indicator, and inherits whatever the module provides. Users of uncertified or external cameras should know IR capture may be visually silent.
- D3cold and control state ([UVC control cache](https://learn.microsoft.com/windows-hardware/drivers/stream/camera-device-uvc-control-cache)): the cache protocol restores a fixed `VIDEOPROCAMP` list, and extension-unit state is not on it. Re-applying the control at every capture is recorded as the deliberate answer to the same power behavior; when-within-a-capture stays with #168.
- What Linux does not reproduce: the DeviceMFT policy layer, and Enhanced Sign-in Security ([ESS requirements](https://learn.microsoft.com/windows-hardware/design/device-experiences/windows-hello-enhanced-sign-in-security): ESS camera firmware under the inbox driver, a VBS-protected memory pathway, an SDEV ACPI table). Stated so nothing implies opening the same IR node approximates that trust path.
- The 0.7.1 agreement (descriptor-advertised probing, disable on failed init, per the [extension-unit device requirements](https://learn.microsoft.com/windows-hardware/drivers/stream/device-requirements-for-usb-video-class-extension-units)) and the issue's not-established list (frames per auth, window, threshold) are both carried.

A deep-review pass re-fetched all five Microsoft pages and checked every claim against them. Two overstatements were found and corrected in the threat model and here: "sharing consumers are read-only" (they can change legacy and vendor controls) and "indication is wired to sensor power in hardware" (the requirement binds the ISP, not a wire).

## The review round (`6e5d537`)

The single Codex round found the scan failing open in production, described under the properties bullet above; the fix is in `6e5d537`. Also from the round: `enable`'s doc comment had been captured by the helper sitting between the doc and the function, the changelog called sharing consumers read-only where Windows only forbids extended-control changes, and the stand-down message carried fourteen accidental spaces. All fixed in the same commit.

New tests, each proven by a hand-applied mutant that made exactly that test fail: an unlistable fd dir sets `permission_denied` with `consumers` empty (mutant: restore the silent collapse); a denied readlink does the same through the other arm (mutant: drop that arm's flag); a blind spot alone never stands the emitter down (mutant: the fail-closed variant). Both denial tests construct the denial with directory permissions and skip loudly under a privileged uid instead of passing vacuously.

## Testing

- [x] `cargo test -p irlume-camera`: 224 pass after the round's fixes (up from 220; the runs-as-root failure is pre-existing and identical on `main`)
- [x] `cargo test -p irlume-cli`: all six targets green
- [x] clippy `-D warnings`, `fmt --check`, `doc -D warnings` (re-run on `6e5d537`)
- [x] CHANGELOG and THREAT_MODEL updated, including the production blind spot
- [ ] Hardware: one capture on the reference Zenbook to confirm the happy path (emitter drives as before, degradation line appears once in the journal), a same-uid holder to see the stand-down line, and a cross-uid holder under the packaged capability set to see the incomplete-scan path, the case the round proved the original code could not detect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ All notable changes to irlume are documented here. This project adheres to
   settings. irlume wrote the extension-unit control whenever it captured,
   relying on its capture failing busy afterwards anyway. `ir_emitter::enable`
   now stands down from the write when another process already holds the camera
-  node, leaving that application's configuration untouched — the fail-safe
+  node, leaving that application's configuration untouched, the fail-safe
   direction, since an unlit emitter degrades one capture toward the password.
   The threat model gains a sourced section on the whole Windows camera
   contract: the exclusive-control model, the certification tie between IR
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ All notable changes to irlume are documented here. This project adheres to
 
 ### Added
 
+- **The emitter write honours the Windows exclusive-control model.** Windows
+  permits many camera consumers but only one may modify device configuration;
+  shared consumers are read-only and inherit the controlling application's
+  settings. irlume wrote the extension-unit control whenever it captured,
+  relying on its capture failing busy afterwards anyway. `ir_emitter::enable`
+  now stands down from the write when another process already holds the camera
+  node, leaving that application's configuration untouched — the fail-safe
+  direction, since an unlit emitter degrades one capture toward the password.
+  The threat model gains a sourced section on the whole Windows camera
+  contract: the exclusive-control model, the certification tie between IR
+  capture and visible indication (and irlume's position on it), why
+  re-applying the control every capture is the deliberate answer to D3cold,
+  and what Linux does not reproduce (DeviceMFT, Enhanced Sign-in Security) and
+  irlume does not claim to. Closes #169.
+
 - **A dark or blinded IR capture reports what its evidence supports.** A dark
   burst used to get one hint ("no active emitter; run `sudo irlume ir-setup`")
   even though shutters, covers, range, exposure and emitter failures all
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,18 @@ All notable changes to irlume are documented here. This project adheres to
 ### Added
 
 - **The emitter write honours the Windows exclusive-control model.** Windows
-  permits many camera consumers but only one may modify device configuration;
-  shared consumers are read-only and inherit the controlling application's
-  settings. irlume wrote the extension-unit control whenever it captured,
-  relying on its capture failing busy afterwards anyway. `ir_emitter::enable`
-  now stands down from the write when another process already holds the camera
-  node, leaving that application's configuration untouched, the fail-safe
-  direction, since an unlit emitter degrades one capture toward the password.
+  permits many camera consumers but only one controlling instance; sharing
+  consumers cannot change extended camera controls and inherit the controlling
+  application's media type. irlume wrote the extension-unit control whenever
+  it captured, relying on its capture failing busy afterwards anyway.
+  `ir_emitter::enable` now stands down from the write when it sees another
+  process holding the camera node, leaving that application's configuration
+  untouched, the fail-safe direction, since an unlit emitter degrades one
+  capture toward the password. The scan has a stated blind spot: reading
+  another user's `/proc/<pid>/fd` is ptrace-gated and the packaged daemon does
+  not hold `CAP_SYS_PTRACE`, so a cross-uid holder can go unseen. The scan
+  reports that instead of hiding it, the daemon logs the degradation once and
+  proceeds, and #207 tracks closing the blind spot.
   The threat model gains a sourced section on the whole Windows camera
   contract: the exclusive-control model, the certification tie between IR
   capture and visible indication (and irlume's position on it), why
--- a/crates/irlume-camera/src/ir_emitter.rs
+++ b/crates/irlume-camera/src/ir_emitter.rs
@@ -1056,8 +1056,8 @@ fn write_if_different(
 /// PIDs (with `comm`) of OTHER processes holding `dev` open, read from
 /// `proc_root` (`/proc` in production; a constructed tree in tests).
 ///
-/// The check is definitionally racy — a consumer can arrive one instant after
-/// the scan — and that is acceptable for what it guards: honouring the
+/// The check is definitionally racy (a consumer can arrive one instant after
+/// the scan), and that is acceptable for what it guards: honouring the
 /// read-only-sharer contract for consumers that exist at decision time, not a
 /// mutual-exclusion primitive (the kernel lock does that for irlume's own
 /// processes). Unreadable entries are skipped silently: as root (the daemon)
--- a/crates/irlume-camera/src/ir_emitter.rs
+++ b/crates/irlume-camera/src/ir_emitter.rs
@@ -1053,8 +1053,51 @@ fn write_if_different(
 /// hardware-target substitution through the API of the type that exists to
 /// prevent it (#189). Holding the `Arc` makes the descriptor outlive the guard
 /// by construction.
+/// PIDs (with `comm`) of OTHER processes holding `dev` open, read from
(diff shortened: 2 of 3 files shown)


archledger:
Kernel-level confirmation of the blind-spot mechanism on the reference Zenbook, under the packaged unit's exact capability set (`systemd-run -p CapabilityBoundingSet="CAP_DAC_OVERRIDE CAP_FOWNER"`, root) against a uid-1000 process:

- listing `/proc/<pid>/fd` succeeds (CAP_DAC_OVERRIDE covers the directory's DAC check);
- `readlink` on an entry fails with errno 13 EACCES (the PTRACE_MODE_READ_FSCREDS gate), which Rust maps to `ErrorKind::PermissionDenied`;
- plain unrestricted root readlinks the same entry fine.

So in production the scan reaches the per-entry `read_link` arm and that arm's `permission_denied` flag is the one that fires; the unlistable-directory arm is the fallback shape. Both arms are mutant-proven in the unit tests. Still owed from a live capture: the once-per-process degradation line in the journal and the same-uid stand-down line, in the hardware batch.

archledger:
NexiGo validation on archhost, `ir_refire_probe` built from `6e5d537`:

- Unprivileged run: the consumer scan hit cross-uid processes and printed the degradation line exactly once, citing #207, then proceeded. The write itself was refused for an unrelated reason: a stale emitter-journal record from an interrupted setup was pending on this box, and the guard correctly refused until it was resolved.
- Root run (after recovery applied that record): three 150-frame streams, emitter lit through each (lit level 25 to 28 against a 0.0 floor), no stand-down, no degradation line, which matches unrestricted root passing the ptrace gate.

Together with the capability-set probe above, every scan outcome has now been observed on real hardware except the two that need the packaged daemon on the Zenbook: the once-per-daemon degradation line during a real capture, and the same-uid stand-down line. Those are in the user's hardware batch.

archledger:
Final cell of the behavior matrix, live on archhost: with another process holding `/dev/video2`, the root probe printed

```
irlume: not driving the IR emitter on /dev/video2: another application holds this camera (sleep(1043753)); its configuration is left untouched
```

and the 150-frame run shows `lit 0.0` throughout after the module's own start-strobe faded: irlume wrote nothing. Every scan outcome has now been observed on hardware: stand-down with the holder named (this run), degradation-once-then-proceed under the packaged capability set (Zenbook daemon startup journal) and as an unprivileged caller (earlier archhost run), clean emitter drive with no holder (3x150 frames lit), and the EACCES-at-readlink mechanism under the exact packaged `CapabilityBoundingSet`.

Issue #235, opened 2026-08-02
A flat vinyl print passes every native cross-spectrum cue; only the third-party PAD model denies it


archledger:
Measured on the Zenbook, 2026-08-02 evening (setup: ASUS FHD module IR pin, GREY8 640x400, daemon built from main at 1289ae4, `IRLUME_LOG=debug`, dark room). A vinyl print of the enrolled user's face (a graduation banner) was held facing the camera and `irlume identify` was run four times. All four attempts produced this verdict from the native gate:

```
liveness(cross-spectrum): Live (live: face in RGB+IR, co-located, frontal, IR-reflective, 3D)
```

The denial came only from the third-party PAD stage each time:

```
thirdparty-pad('flir'): p_fake 0.993 >= 0.90; downgrading Live to Spoof
```

The four cross-spectrum readings for the print:

| attempt | ir_bright | centre/edge | glint | face_frac | ir_clipped | flir p_fake |
|---|---|---|---|---|---|---|
| 1 | 116 | 1.12 | 130 | 0.315 | 1.9% | 0.993 |
| 2 | 76 | 1.12 | 84 | 0.302 | 0.2% | 1.000 |
| 3 | 98 | 1.17 | 107 | 0.356 | 0.6% | 0.999 |
| 4 | 70 | 1.17 | 84 | 0.187 | 0.0% | 0.995 |

What each native cue did against the print:

- **IR reflectivity:** the vinyl reproduces the whole face in IR at full contrast (burst frames on #222), so "a real face reflects 850nm" holds for vinyl too.
- **Centre/edge ratio (the 3D cue):** 1.12 to 1.17, above the 1.03 floor in every attempt. A flat print defeats the floor as tuned.
- **Glint:** 84 to 130 against the pegged 255.00 of nine out of ten genuine logins tonight. The cue separated print from cornea in this data, but the earlier glossy-print result (69/70) shows a surface with the right gloss can peg it, so it is not a floor to lean on.
- **Co-location, frontality:** both satisfied by a print held square to the camera.

Why the floor cannot simply be raised: #221's measurements put a saturated genuine capture at ratio 1.11, below the print's 1.12, while unclipped genuine captures read 1.35 to 1.54. The genuine and spoof distributions only separate after clipped frames stop being gated. Sequence that follows: land least-clipped-frame selection (#221), re-measure the genuine floor, then retune the 1.03 threshold with the print data as the attack reference.

[...]

Issue #197, opened 2026-08-01
A covered IR camera under an active emitter saturates flat (mean 255); the dark-gated diagnosis never sees the most common cover case


archledger:
Found during #196's hardware validation on the ASUS 3277:0059 (built-in laptop camera), three runs, identical result each time.

## Measured

An opaque cover (card) over the camera with the emitter active does not produce a dark frame. It produces total saturation: the burst's chosen frame reads mean 255.0 with the camera classifying 10/10 frames, because the cover reflects the 850nm emitter straight back into the sensor at zero distance. The #196 diagnosis is gated on `mean < IR_DARK_HINT_MAX` (35), so no cause is ever named for the cover case the flat-frame arm was written for.

For contrast, the two related signatures already measured on this camera:

- E-shutter engaged (#186): firmware-substituted constant 144.0, stddev exactly 0.00, `privacy` reads 1, and capture refuses before diagnosis.
- Genuinely dark room with the emitter off (#196 phase C): mean 0.7-1.8 with real texture.

So on this hardware, "flat and saturated" is the cover signature, "flat at 144" is the shutter signature, and "dark with texture" is a dark scene. The current code names only causes below the dark gate.

## Shape

Extend the diagnosis entry to a second gate: a near-constant frame at ANY mean is not a scene. A flat frame at or near saturation with lit flags present reads as "something flat is covering the lens, reflecting the emitter back"; the existing `FlatFrame` wording rules apply (name the measurement, claim only what was measured on which camera). The `IRLUME_IR_EMITTER=off` output policy stays first.

Wants a stddev reading from a saturated covered frame before coding: 255-clipped frames are plausibly stddev 0.00 exactly, but that is an assumption until measured (the #196 harness logs only the mean today).

Relates to #185 (this is its seventh signature), #186, #192.

archledger:
Confirmed on a second camera model: the NexiGo 3443:c803 covered under its active emitter reads mean 252.8, three runs out of three (ASUS 3277:0059 read 255.0, also three of three). Both sit far above the dark gate, so the flat-at-any-mean extension applies to both measured emitters. The NexiGo's near-saturation at 252.8 rather than a full 255 means the flat check must not assume exact clipping.

PR #236, opened 2026-08-03
Gate the brightest lit frame that is not clipped, not the brightest outright

### LINKED ISSUE #221: Sensor saturation compresses the centre/edge ratio toward the spoof floor, and the first capture is the most likely to clip
Found while researching #174 from data already on disk (`~/irlume-caldata`), not from a new capture.

## The measurement

`depth_real_dark.jsonl` and `depth_real_lit.jsonl` each hold 15 captures with both an IR brightness and `ir_depth` (the centre/edge ratio under its old name). In both sessions the FIRST capture is a lone outlier at the top of the sensor range, and its ratio collapses toward the floor:

| session | first capture | every later capture |
|---|---|---|
| depth_real_dark | brightness 238.4, ratio **1.061** | 55-90, ratio 1.187-1.275 |
| depth_real_lit | brightness 233.6, ratio **1.117** | 129-151, ratio 1.285-1.424 |

`MIN_CENTER_EDGE_RATIO` is 1.03. The first capture clears it by 0.03; the rest of the same session clears it by 0.16 to 0.39.

(Worth stating because it nearly became a false finding: pooled over all 15 samples, brightness and ratio look strongly anti-correlated, r = -0.87 and -0.85. That is entirely this one outlier. Excluding it, r = -0.11 and -0.28 on n=14, which is no evidence of a trend.)

## Why saturation would do this

A clipped centre cannot read brighter than a clipped rim, so saturation compresses the ratio toward 1 in exactly the way an ambient pedestal does: `(C+A)/(E+A) < C/E`. irlume already reasons about the ambient end and guards it (`IR_AMBIENT_FLOOD` = 170, verdict kept fail-closed, only the REASON reworded so the user is told what to fix). The saturation end has no equivalent guard, even though `ir_probe::saturated_fraction` already exists and is printed in the ambient-subtraction debug line (`crates/irlume-camera/src/lib.rs:1589`).

Two design details make the exposure end worth checking:

- The gate reads the BRIGHTEST frame of the burst (`brightest_lit`), which is also the frame most likely to be clipped.
- `AE_WARMUP` is 6 frames. This data suggests six may not be enough for the first capture after a stream starts, which is precisely the capture an authentication performs.

## What is NOT claimed

That any login has failed this way. These are dev-tool captures, the sample is two sessions of one subject on one module, and no production authentication has been observed clipping. The point is that the cue has a documented starvation mode at one end and an unguarded, instrumented, plausible one at the other.

## Cheapest next step

The saturated fraction of the face region is already computable. Recording it alongside the cues (as #220 just did for `face_frac`) would show whether real authentications ever run clipped, before anyone changes a threshold or a warm-up constant. If they do, the symmetric fix to the flood guard is to say the cue is starved rather than read a compressed ratio as flatness.

Full working notes: `~/irlume-research/2026-08-02-liveness-normalization/FINDINGS.md`.

archledger:
Closes #221 (the gating half; the floor retune it unblocks is tracked there and on #235).

## What changes

`capture_with_stats` gated the brightest camera-flagged-lit frame of the burst. The brightest frame of a warming burst is exactly the one most likely to clip, and the 2026-08-02 hardware sitting showed what that costs: the first capture after a daemon restart at close range gated a 78.8%-clipped frame, the centre/edge ratio collapsed to 1.11, the IR detector found no facial structure, and a real user was denied as Spoof. The same position one capture later read 12.7% clipped and passed at 1.45.

Selection now runs through `ir_metadata::best_gate_frame`:

- among camera-flagged-lit frames, gate the brightest whose clipped fraction is at most `CLIPPED_FRAC_MAX` (5%, the same cutoff the ambient-subtract debug line has always called a blown exposure; measured captures at 0.9 to 13% read ratios 1.41 to 1.54, 20.5% read 1.39, 70%+ read 1.11 to 1.19);
- when every lit frame clips harder than that, gate the least clipped one (brightest on a tie), so a cover or extreme proximity still gates the best frame that exists;
- with no illumination metadata, or a source format that cannot name its ceiling (`clipping_white_level` answers `None`), the long-standing brightest scan is unchanged. An unclassified strobing burst's cleanest frames are the emitter-off ones, so clip-aware selection there would trade a clipped face for no face.

A missing per-frame clipping entry counts as fully clipped, never clean: failure to observe must not authorize.

## What this deliberately does not change

- The 1.03 centre/edge floor. The flat-vinyl measurements on #235 (print reads 1.12 to 1.17) show the floor should rise, but only after this change stops clipped genuine captures (1.11) from overlapping the print range. Retune is follow-up work with #235 as the attack reference.
- The cover diagnosis (#197 band): a covered sensor clips every frame, so least-clipped is still saturated and the diagnosis still fires on `best_mean`.
- `ambient_mean` and the flood guard: computed as before.
- Ambient subtraction stays opt-in and untouched.

## Verification

- 7 unit tests cover each selection branch: skip-clipped-brightest, never-trade-face-for-emitter-off, clean-burst parity with `brightest_lit` (including first-on-tie), no-clip-data parity, no-flags parity, threshold boundary (exactly 5% is clean), and the clipping-tie brightness tie-break.
- 5 hand-applied mutants (ignore clipping, drop the lit restriction, `<=` to `<`, drop the tie-break, run clip-selection without flags) each fail their named test; applied and reversed by a self-restoring script, tree byte-identical after.
- Full workspace suite and the no-TPM namespace run (`unshare -Umr`, tmpfs `/dev`) green.
- Hardware: [pending, will be edited in before merge]

## Not tested

- The call-site wiring (`clipped_fracs` computed and passed) has no unit test: nothing in the suite can construct a live capture, the same disclosure as #220's `face_frac`. The hardware check is the `clipped N%` field now printed in the `[ir] illumination` debug line, plus the login line's `ir_clipped` dropping under the forced-clipping repro.
- Grey16/NV12/YUYV sources take the unchanged fallback by construction; no such module is on hand to measure.
- Bright ambient. All measurements behind the threshold were dark-room.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -1163,9 +1163,10 @@ impl Engine {
                 .map(|f| format!("{:.1}%", f * 100.0))
                 .unwrap_or_else(|| "n/a".into()));
         // Opt-in third-party PAD cue: score whenever an IR face is present (the
-        // `ir` frame is the brightest strobe phase, i.e. the LIT frame, which is
-        // the regime the cue was measured in), so the dark path can consult the
-        // result too. DENY-ONLY: it can downgrade Live to Spoof and nothing else.
+        // `ir` frame is a LIT strobe phase, since #221 the brightest one that
+        // is not clipped, which is closer to the regime the cue was measured in
+        // than a blown exposure), so the dark path can consult the result too.
+        // DENY-ONLY: it can downgrade Live to Spoof and nothing else.
         let thirdparty_fake = match (self.tp_pad.as_mut(), ir_top.as_ref()) {
             (Some((pad, _, _)), Some(f)) => match pad.p_fake(&ir_view, &f.bbox) {
                 Ok(p) => Some(p),
--- a/crates/irlume-camera/src/ir_metadata.rs
+++ b/crates/irlume-camera/src/ir_metadata.rs
@@ -320,6 +320,64 @@ pub(crate) fn brightest_lit(means: &[f64], flags: &[Option<Illumination>]) -> Op
     best.map(|(i, _)| i)
 }
 
+/// A gate frame may clip at most this fraction of its pixels before selection
+/// prefers a cleaner frame. 5% is the cutoff the ambient-subtract debug line
+/// has always used for "blown exposure", and it sits under the smallest
+/// clipping measured to move the centre/edge ratio (#221: captures at 0.9-13%
+/// read 1.41-1.54, 20.5% read 1.39, 70%+ read 1.11-1.19 against a 1.03 floor).
+pub(crate) const CLIPPED_FRAC_MAX: f64 = 0.05;
+
+/// Pick the burst frame the liveness gate and matcher read.
+///
+/// The brightest frame of a warming burst is precisely the most likely to
+/// clip: the first capture after a daemon restart at close range gated a frame
+/// with 78.8% of its pixels at the ceiling, the IR detector found no facial
+/// structure in it, and a real user was denied as a spoof; the same position
+/// one capture later read 12.7% (#221). So among the frames the camera itself
+/// flagged lit, take the brightest whose clipped fraction is at most
+/// [`CLIPPED_FRAC_MAX`], and when every lit frame clips harder than that, the
+/// least clipped one (brightest on a tie): the gate reads the best frame that
+/// exists rather than the worst.
+///
+/// `clipped[i]` is the fraction of frame `i`'s pixels at the sensor ceiling,
+/// `None` when the source format cannot say where its ceiling is
+/// ([`super::clipping_white_level`]). A missing per-frame entry is treated as
+/// fully clipped, never as clean: failure to observe must not authorize.
+///
+/// Both fallbacks keep the long-standing brightest scan unchanged: without
+/// camera illumination flags a strobing burst's cleanest frames are the
+/// emitter-OFF ones, so clip-aware selection there would trade a clipped face
+/// for no face at all.
+pub(crate) fn best_gate_frame(
+    means: &[f64],
+    flags: &[Option<Illumination>],
+    clipped: Option<&[f64]>,
+) -> Option<usize> {
+    let lit = |i: usize| matches!(flags.get(i), Some(Some(Illumination::Lit)));
+    let any_lit = (0..means.len()).any(lit);
+    let Some(clipped) = clipped.filter(|_| any_lit) else {
+        return brightest_lit(means, flags);
+    };
+    debug_assert_eq!(clipped.len(), means.len());
+    // Strictly-greater keeps the FIRST frame on a mean tie, matching
+    // `brightest_lit`'s long-standing scan.
+    let mut clean: Option<(usize, f64)> = None;
+    let mut least: Option<(usize, f64, f64)> = None; // (index, clipped, mean)
+    for (i, &m) in means.iter().enumerate() {
+        if !lit(i) {
+            continue;
+        }
+        let c = clipped.get(i).copied().unwrap_or(1.0);
+        if c <= CLIPPED_FRAC_MAX && clean.is_none_or(|(_, best)| m > best) {
+            clean = Some((i, m));
+        }
+        if least.is_none_or(|(_, bc, bm)| c < bc || (c == bc && m > bm)) {
+            least = Some((i, c, m));
+        }
+    }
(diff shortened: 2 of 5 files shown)


archledger:
## The merge-blocking finding is refuted by measurement, and the measurement inverts it

The review's High finding says clipping may be rejecting the print attack by accident, and that preferring a cleaner frame removes that protection. That is a fair hypothesis from the code alone. It is wrong on this hardware, and the way it is wrong matters more than the PR did.

Setup: ASUS FHD IR pin, GREY8 640x400, dark room, `IRLUME_LOG=debug` plus `IRLUME_DEBUG_IR=1`, the vinyl print from #235 held about 15cm from the lens with the subject's real face hidden behind it, confirmed by the person holding it. Each phase starts with `systemctl restart irlumed` so the first capture is the warm-up case, then `irlume identify` repeatedly without moving.

**Old selector (main, `521f3f0`), first capture after restart:**

```
[ir] illumination: chose frame 1 (mean 157.0), lit 157.0 / ambient 0.0
  ir_bright=170 ir_center_edge_ratio=1.19 face_frac=0.571 ir_clipped=24.8%
thirdparty-pad('flir'): p_fake 0.749 is between the measured genuine ceiling and the deny threshold 0.90; abstaining
```

The print passed liveness. It failed only at recognition, because at 15cm the printed face was too large to match the template.

The five captures that followed, same print, same position, clipping 0.1 to 1.6%:

```
thirdparty-pad('flir'): p_fake 1.000 >= 0.90; downgrading Live to Spoof
```

**New selector (this PR), same restart, same print, same distance,** six captures: the gate frame clipped 3.6% then 0.0 to 0.8%, `p_fake` read 1.000 every time, and all six were denied. Where the old selector gated frame 1 at mean 157.0, this one gated frame 7 at mean 114.5 from the same burst.

The mechanism is the reverse of the finding: clipping does not reject the print, it blinds the model that rejects the print. `p_fake` collapses from 1.000 to 0.749 when 24.8% of the frame is at the ceiling, and 0.749 sits in the abstain band, so the cue says nothing and the native gate is left alone with a cue the print already defeats.

The centre/edge cue moves the same direction. On clean frames the print reads 0.97 to 1.20, and twice tonight it was rejected outright by the 1.03 floor:

```
liveness(cross-spectrum): Spoof (IR too flat (center/edge 0.97); looks 2D, not a 3D face); ir_clipped=0.9%
liveness(cross-spectrum): Spoof (IR too flat (center/edge 1.03); looks 2D, not a 3D face); ir_clipped=0.4%
```

On clipped frames it reads 1.17 to 1.19, comfortably above the floor. Clipping inflates the print's ratio away from rejection, so both the third-party model and the native cue are more likely to catch the attack on the frame this PR selects.

That also answers the review's request for a threshold justified against the attack: a retune is now measuring against print ratios of 0.97 to 1.20 rather than the 1.12 to 1.17 recorded on #235, because those earlier numbers came from frames chosen by the old selector.

The genuine-user half of the same sitting, patched daemon, close range after restart: chosen frames clipped 0.2 to 2.7%, ratios 1.50 to 1.64, all Live. The old selector in that same position gated a 78.8%-clipped frame at ratio 1.11 and denied a real user as a spoof.

**What this PR does not do:** the all-clipped fallback still gates a clipped frame when no lit frame is clean, so the bypass window is narrowed, not closed. That is filed as #237 with the numbers, since a fail-closed rule is a new denial path and needs its own false-reject validation rather than being folded in here.

## Review items applied

The Low finding is fixed in `90742a7`. `IrCaptureStats`, `accumulate`'s calibration rationale, `ARCHITECTURE.md` and `THREAT_MODEL.md` all said `lit_mean` was the burst peak; each now says which frame is selected and why. The calibration comment keeps the property it actually depends on, that the figure comes from a lit phase rather than whichever phase the caller received.

The same commit fixes a CI failure the review could not see, since it could not compile: the public doc on `capture_with_stats` linked the private `CLIPPED_FRAC_MAX`, and `-D rustdoc::private-intra-doc-links` rejects that. The doc lane is not part of `cargo test`, which is why it reached CI.

## Corrections to the review's premises

- "The repository documents that this floor is already defeated by a glossy IR print in 69 of 70 trials." That result is a different medium under different selection, and tonight's vinyl was rejected by the floor alone on 2 of 13 clean-frame captures. The floor is weak; it is not uniformly defeated.
- The claimed test and mutant runs could not be audited from the tree, correctly noted. The mutation script is a scratch file rather than a committed harness, which is why. Each mutant and the test it fails is named in the PR body, so they can be reapplied by hand.

PR #238, opened 2026-08-03
Refuse a blown IR frame instead of judging it

### LINKED ISSUE #237: Heavy clipping pushes the PAD cue into its abstain band, and a flat print is then declared Live
Measured on the Zenbook 2026-08-02 (ASUS FHD IR pin, GREY8 640x400, dark room, daemon built from main `521f3f0`, `IRLUME_LOG=debug` and `IRLUME_DEBUG_IR=1`). The vinyl print from #235 was held about 15cm from the lens with the subject's real face hidden behind it, confirmed by the person holding it, after `systemctl restart irlumed` so the first capture caught auto-exposure before it settled.

The first capture:

```
[ir] illumination: 10/10 frames classified by the camera; chose frame 1 (mean 157.0), lit 157.0 / ambient 0.0
  ir_bright=170 ir_center_edge_ratio=1.19 glint=196.00 ambient=0 face_frac=0.571 ir_clipped=24.8%
thirdparty-pad('flir'): p_fake 0.749 is between the measured genuine ceiling and the deny threshold 0.90; abstaining
```

Five more captures of the same print at the same distance, once exposure settled to 0.1 to 1.6% clipped:

```
thirdparty-pad('flir'): p_fake 1.000 >= 0.90; downgrading Live to Spoof
```

**The finding: at 24.8% clipping the PAD model's confidence falls from 1.000 to 0.749, which lands in the abstain band below the 0.90 deny threshold, so the cue stays silent and the print is declared Live.** Every layer that remains is one this print already clears: IR reflectance (vinyl reflects 850nm at full contrast), co-location and frontality (a print held square satisfies both), and the 1.03 centre/edge floor (the clipped frame reads 1.19).

Liveness is the whole of what failed here. Recognition then refused the print for an unrelated reason, that at 15cm the printed face was too large to match the enrolled template, so this particular attempt ended in a denial. A print sized and positioned to match would have authenticated: nothing in the liveness path objected.

An earlier attempt in the same sitting is consistent and worse, though I cannot rule out the subject's real face being in frame for it, so I am recording it as unconfirmed: a 42.7%-clipped capture at ratio 1.17 produced no PAD line at all, meaning `p_fake` fell below the measured genuine ceiling, and the result was a full match (`score 0.663`).

**Why clipping does this, and which direction it runs.** Both anti-spoof signals get weaker as clipping rises, not stronger:

| condition | ir_clipped | centre/edge | flir p_fake | verdict |
|---|---|---|---|---|
| print, first capture after restart | 24.8% | 1.19 | 0.749 (abstains) | **Live** |
| print, settled | 0.1 to 1.6% | 0.97 to 1.20 | 1.000 | Spoof |

On clean frames the print was twice rejected by the native floor alone (`IR too flat (center/edge 0.97)` and `(1.03)`). On clipped frames it never was. Saturation removes the texture both the model and the ratio depend on, and what is left resembles a lit 3D face more than the print does.

**Reachability.** Any condition that blows out an IR frame reproduces it: proximity, a daemon restart, or bright ambient IR. #221 records the same warm-up producing 78.8% clipping on a genuine user. The attacker does not need to induce it deliberately; the camera's own auto-exposure produces it on the first capture, and the attacker controls distance.

**What PR #236 changes.** With clip-aware frame selection the same attack at the same distance was denied 6 of 6, because the gate frame clipped 3.6% and below and `p_fake` read 1.000 each time. That narrows the window but does not close it: the all-clipped fallback still gates a clipped frame when no lit frame in the burst is clean, which is reachable with a cover or at extreme proximity.

**Proposed fix, for discussion rather than as a settled design.** A gate frame above some clipping level is not evidence of anything, so `Live` should not be reachable from it. Options:

1. Refuse a heavily clipped gate frame with a retry-shaped verdict (`Uncertain`, "the IR frame is blown, move back or retry"), never `Live`. Fail-closed and honest about the cause.
2. Treat a PAD abstention as a denial when the gate frame is clipped, on the ground that the abstain band was measured on unclipped frames and says nothing about this regime.

Both cost false rejections in exactly the regime #221 exists to fix, so either needs its own false-reject measurement before shipping. The threshold should come from measurement: `p_fake` was 1.000 across every capture at or below 1.6% clipping and 0.749 at 24.8%, so the intermediate range is unmeasured and worth sampling before a number is picked.

Related: #235 (the same print clears the native gate), #221 (clipping and the warm-up that causes it), #101 (the corpus that would let the PAD threshold be retuned rather than guessed).

Not tested: bright ambient, other modules, other print media, and the clipping range between 2% and 24%. One subject, one room, one camera.

archledger:
Closes #237.

## Why

A face region at the sensor ceiling measures nothing, and irlume was judging it anyway. Measured against the flat vinyl print from #235 on the ASUS module, the third-party PAD cue decays as clipping rises:

| face region clipped | flir p_fake |
|---|---|
| 0.0 to 2.7% | 1.000 to 0.999 |
| 5.3% | 0.998 |
| 8.8% | 0.963 |
| 24.8% | 0.749, abstains |
| 42.7% | below 0.13, silent |

Past roughly 13% the cue falls under its own 0.90 deny threshold and abstains, so the print is left facing only cues it already clears, and it was declared Live. The centre/edge ratio moves the same way: clipped it reads 1.17 to 1.19, clean it reads 0.97 to 1.20 and twice rejected the print outright against the 1.03 floor. Saturation weakens both sides of the evidence at once, which is why this refuses the frame rather than interpreting it.

The abstain band it lands in is documented as "the band neither class occupied" during qualification, and that qualification never covered clipped frames. An abstention there is not a considered "no objection".

## What changes

`Signals::ir_saturated_frac`, recorded since #224, is now gated at `IR_SATURATED_FRAC_MAX` (10%), before any cue that reads the pixels. Over the limit the gate returns `Uncertain` with the exposure named, never `Live`.

`Uncertain`, not `Spoof`, because the capture failed to measure a face rather than showing a fake one; that also puts it in `presence_retryable`, which is what keeps the cost off real logins. A format that cannot name its ceiling reports `None` and is judged exactly as before, since refusing on a number nobody produced would deny every non-GREY8 module.

10% sits between two measured populations: with clip-aware selection (#221) every genuine gate frame measured at or below 6.3%, and the print needed about 13% before the PAD cue went quiet. Both margins are under 4 points, from one camera in one room, and the constant says so.

## Verification

Container: 2 unit tests (refusal across 0.11 to 1.0; unchanged judgement at `None`, 0.0, 0.063 and exactly the limit, on both the live and flat sides), 5 hand-applied mutants each failing their named test (gate removed, unmeasurable-denies, boundary excluded, Spoof instead of Uncertain, limit loosened to 30% past the PAD cliff), reversed by a self-restoring script with the tree byte-identical after. Full workspace suite green.

Hardware, ASUS module, dark room, one subject, glasses on throughout:

- **Spoof side.** The vinyl print at close range, which reached `Live` on main through a 24.8%-clipped frame, is now refused by the exposure gate before any cue reads it.
- **False rejects, single capture.** 12 `irlume identify` runs: 9 authenticated (normal seating and arm's length, gate frames at or under 1.3%), 3 refused while leaning in close, at 16%, 25% and 50.5% of the face clipped. Those are the captures where the whole burst clipped, so #221's selection had nothing clean to choose.
- **False rejects, real login.** 3 `pam_authenticate` attempts on the sudo stack from that same close position, after a daemon restart each time: all 3 succeeded in 4.0 to 7.8 seconds, gate frames at 6.7 to 8.7%.

The difference between the last two is the point of the `Uncertain` classification. `irlume identify` takes one capture and reports what it got; a login gets a 15 second grace window that retries presence failures, and auto-exposure settles across captures (78.8, 12.7, 5.3, 4.3% over four consecutive captures at one position, measured on #221).

## Not tested

- **I never caught a login whose first capture was over the limit.** The three PAM attempts all drew a gate frame under 10%, so the retry absorbing this specific refusal is an inference, not a measurement: it rests on the refusal mapping to `OutcomeKind::Uncertain`, `presence_retryable` including that kind, and the grace window being measured retrying a different `Uncertain` (an off-angle frame at 21:02, retried and authenticated 4 seconds later). The condition is intermittent, and I could not force it.
- Bright ambient, other modules, other subjects, and the clipping range between 10% and 16% on the genuine side.
- Whether 10% is right for a camera whose face region clips differently from its frame average. On this module the face clipped 50.5% while the whole frame clipped 20.1%, since the face is nearest the emitter; selection can only see the frame-wide number, which is why some bursts have no clean frame to offer.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -1152,7 +1152,7 @@ impl Engine {
         // Log the cue values on PASS too; a near-miss on a genuine user is
         // invisible in the outcome line but obvious here.
         irlume_common::dlog!(
-            "liveness(cross-spectrum): {verdict:?} ({reason}); ir_bright={:.0} ir_center_edge_ratio={:.2} glint={:.2} ambient={:.0} yaw_asym={:.2} pitch={:.2} face_frac={:.3} ir_clipped={} (face_frac #174, clipped #221; both gate nothing)",
+            "liveness(cross-spectrum): {verdict:?} ({reason}); ir_bright={:.0} ir_center_edge_ratio={:.2} glint={:.2} ambient={:.0} yaw_asym={:.2} pitch={:.2} face_frac={:.3} ir_clipped={} (face_frac #174, recorded only; clipped #237, refused past the limit)",
             signals.ir_face_brightness, signals.ir_center_edge_ratio, signals.ir_eye_glint,
             signals.ir_ambient, signals.head_yaw_asym, signals.head_pitch_frac,
             signals.face_frac,
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -1137,8 +1137,13 @@ impl Engine {
             ir_ambient: ir_stats.ambient_mean,
             // From the IR frame, because the IR cues are measured there.
             face_frac: face_frac_of(ir_top.as_ref().map(|f| &f.bbox), ir.width),
+            // Measured on the RAW gate frame. `ir.data` is the subtracted image
+            // when ambient subtraction is on, and subtraction drops every
+            // ceiling sample below the ceiling, so a 25%-clipped face would
+            // report 0% and the exposure gate would pass a frame carrying
+            // nothing (#238 review).
             ir_saturated_frac: saturated_frac_of(
-                &ir.data,
+                ir_stats.saturation_frame.as_deref().unwrap_or(&ir.data),
                 ir.width,
                 ir.height,
                 ir_top.as_ref().map(|f| &f.bbox),
@@ -1877,6 +1882,21 @@ impl Engine {
             }
         }
         let a = self.assess()?;
+
+        // An unreadable frame is reported as unreadable BEFORE anything derived
+        // from it is consulted. The eye cue below is computed from the same IR
+        // pixels, so a blown frame that hides the corneal glints would deny with
+        // OtherDeny, which is NOT presence-retryable: the grace window would
+        // stop instead of letting exposure settle, turning a retryable quality
+        // refusal into a terminal one for anybody with require_eyes_open on
+        // (#238 review). Uncertain is the only verdict this promotes; a Spoof
+        // still reaches its own branch below with its own reason.
+        if a.verdict == Verdict::Uncertain {
+            return Ok(Outcome::deny(
+                liveness_deny_kind(a.verdict, &a.reason),
+                format!("liveness {:?}: {}", a.verdict, a.reason),
+            ));
+        }
 
         // Opt-in hard gate: never unlock unless both eyes read open.
         if enr.require_eyes_open && !a.eyes_open {
@@ -3930,6 +3950,35 @@ mod tests {
         assert_eq!(
             saturated_frac_of(&grey, 4, 4, Some(&bbox), Some(255)),
             Some(1.0)
+        );
+    }
+
+    /// Ambient subtraction hides the ceiling it subtracts from, so the exposure
+    /// gate must read the raw gate frame that `IrCaptureStats::saturation_frame`
+    /// preserves. Measuring the returned pixels instead reports a blown face as
+    /// pristine, which is the fail-open the #238 review found: 255 minus an
+    /// ambient 1 is 254, and 254 is not at the ceiling.
+    #[test]
+    fn subtraction_hides_clipping_so_the_gate_reads_the_raw_frame() {
+        let bbox = [0.0f32, 0.0, 4.0, 4.0];
+        // A face region a quarter of which reached the sensor ceiling.
+        let raw: Vec<u8> = [
+            255, 255, 255, 255, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10,
+        ]
+        .into_iter()
+        .collect();
+        let ambient = vec![1u8; 16];
+        let returned = irlume_camera::ir_probe::subtract(&raw, &ambient);
+
+        assert_eq!(
+            saturated_frac_of(&returned, 4, 4, Some(&bbox), Some(255)),
+            Some(0.0),
+            "control: the returned image no longer shows the clipping"
+        );
+        assert_eq!(
+            saturated_frac_of(&raw, 4, 4, Some(&bbox), Some(255)),
+            Some(0.25),
+            "the raw gate frame is where the clipping is still measurable"
         );
     }
 
--- a/crates/irlume-camera/src/lib.rs
+++ b/crates/irlume-camera/src/lib.rs
@@ -84,6 +84,7 @@ pub(crate) mod testenv {
 
 use irlume_common::Error;
 use v4l::buffer::Type;
+use v4l::format::Quantization;
 use v4l::io::traits::CaptureStream;
 use v4l::video::Capture;
 use v4l::{Device, Format, FourCC};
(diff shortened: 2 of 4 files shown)


archledger:
## Review round applied: the High finding was real, and reachable on this machine

**The IR-only path was unguarded.** `evaluate_ir_only` is the evaluator that authenticates when RGB finds nothing, it releases credentials, and it kept returning `Live` for exactly the frames `evaluate` had begun refusing. My claim that the frame is refused "before any cue that reads the pixels" was false for that caller.

It was not theoretical here. A `pam_authenticate` run on this laptop 20 minutes before the review landed went through it:

```
face auth 'wisbfime': granted=true live=true score=0.721 (match: BEN (ir-fallback, dim light; rgb 0.00<0.60))
```

Fixed in `4e7714f`, but not by copying the block into the second evaluator. The check now lives in one `exposure_refusal` function that both call, because a guard duplicated across two entry points is the same defect waiting for a third. The tests assert both paths, and a mutant that removes the call from `evaluate_ir_only` alone fails them; the previous single-path test could not have caught this, which is the point.

**The harness gap was real too.** `padcapture` built its `Signals` with `..Default::default()`, so `ir_saturated_frac` was always `None` and every qualification run judged its corpus against a gate configuration authentication never uses. It now captures through `capture_ir_with_stats` and passes the measured fraction, and the JSON carries both `ir_saturated_frac` and `ir_exposure_ok` so a refused frame is distinguishable from a judged one.

The follow-up is fixed in the same commit: the cross-spectrum log line still advertised that clipping gates nothing.

**Verification after the fixes:** 6 mutants killed (gate removed, IR-only path unguarded, unmeasurable-denies, boundary excluded, Spoof instead of Uncertain, limit loosened to 30% past the PAD cliff), each failing its named test, applied and reversed by a self-restoring script with the tree byte-identical after. Full workspace suite, clippy and the doc lane green.

**On the one claim you could not verify:** the p_fake decay curve is mine, measured tonight on one ASUS module against the #235 vinyl print, and it is the whole basis of the 10% constant. The raw journal lines are on #237. It deserves the scepticism; a second camera and a second subject would be the way to settle it, and neither exists yet.

archledger:
## Round 2 applied: both findings were real fail-opens

**Ambient subtraction erased the evidence.** Correct, and it is the worse of the two because the gate would have looked like it was working. `ir.data` is the subtracted image when `IRLUME_IR_AMBIENT_SUBTRACT=1`, and saturating subtraction moves every ceiling sample below the ceiling, so a 25%-clipped face measures 0% and the refusal never fires. `IrCaptureStats` now carries the raw gate frame, and both the auth path and `padcapture` measure that instead.

Two deviations from the proposed patch, both deliberate. The field is populated **only on the branch where subtraction actually replaces the payload**, so the shipped default (subtraction off) clones nothing and `None` means "the returned frame IS the raw one, measure it". And `IrCaptureStats` loses `Copy`, since a struct that silently copies a frame of pixels per use is not something a caller should get by accident.

The regression test asserts both halves of your decisive observation: the subtracted image reports `Some(0.0)` and the raw frame reports `Some(0.25)`.

**The eyes-open ordering was real too**, and it contradicted the central claim of this PR. `require_eyes_open` denies with `OtherDeny`, `presence_retryable` excludes it, and a blown frame is exactly the condition that hides the corneal glints, so users with that option enabled would have got a terminal denial where the grace window was supposed to retry. An `Uncertain` verdict now returns before any cue derived from the unreadable frame is consulted. `Spoof` still falls through to its own branch with its own reason, so nothing else moves.

Both fixed in `a34a34e`. Workspace suite, clippy and the doc lane green.

**On the qualification-band claim you flagged as unestablished:** you are right that the diff does not support it. What the repository states is that `MEASURED_GENUINE_CEILING = 0.13` is the highest `p_fake` a genuine face reached during qualification and that scores between it and the deny threshold are "the band neither class occupied". That the band was never *exercised with clipped frames* is my inference from the qualification notes, not something the code establishes, and the wording in `Signals` should not have implied otherwise. The load-bearing part is measured: `p_fake` decays with clipping on this camera, and the numbers are on #237.

archledger:
## Final-code validation (`ef4252a`)

Re-validated after the round-3 quantization change, because the earlier hardware numbers were taken on a revision three commits back.

**Genuine user**, ASUS module, dark room, glasses on:

| position | gate frame clipped | outcome |
|---|---|---|
| normal seating, 2 runs | 0.8%, 0.9% | Live, authenticated |
| leaning close, 2 runs | 57.5%, 24.0% | refused, "IR frame blown out" |
| arm's length, 2 runs | 0.9%, 1.1% | Live, authenticated |

**The #235 vinyl print**, held close, 3 runs: gate frames 0.9 to 1.4% clipped, `p_fake` 0.999, 1.000, 1.000, denied every time.

Worth stating because it is not obvious: the print produces *cleaner* frames than a real face at the same distance, since vinyl returns far less 850nm than skin. So the print gets a readable frame and the cue that catches it works, while the blown regime that used to silence that cue is now refused before anything reads it. The two failure modes do not overlap the way I expected when I wrote the gate.

The close-range refusals remain the cost, unchanged from the earlier round and still classified `Uncertain` so the grace window can retry.

**Also re-run on the final code:** 6 mutants killed, tree byte-identical after; workspace suite, clippy and the doc lane green; the no-TPM namespace run shows the same two model-verification failures as `main` (harness artifacts under an empty `/dev`, green in CI), so no regression.

**On process:** this PR took three review rounds, and the project's standing rule is one. That is my error, not the reviewer's; each round returned real findings and I kept going rather than fixing the round's findings and deciding on the measurements. All three rounds' merge-blocking findings are fixed and proven, and the merge decision rests on the hardware data above.

PR #239, opened 2026-08-03
Raise the centre/edge floor to where the measurements put it

### LINKED ISSUE #235: A flat vinyl print passes every native cross-spectrum cue; only the third-party PAD model denies it

archledger:
Closes #235.

## Why

`MIN_CENTER_EDGE_RATIO` was 1.03, and a flat vinyl print of the enrolled user's face reads 1.02 to 1.21 on the ASUS module. `padcapture` presented it 24 times tonight and the gate called **22 of them Live**, with the harness printing its own warning:

```
[padcapture] ⚠ 12/12 attack presentations were ACCEPTED; investigate before trusting this build.
```

This is the default configuration. The third-party PAD cue that reliably denies this print is opt-in: `docs/SETUP.md:424` documents `sudo irlume models enable flir` as a manual step, and the weights live in the state directory rather than in any package. An install that never ran that command has this gate and nothing else standing between a printed face and a Live verdict.

#237 was the same failure amplified by clipping. This is the unamplified version, and it does not need a blown frame or a warm-up window.

## The corpus

84 presentations through the real `LivenessGate` via `padcapture`, ASUS FHD IR pin, GREY8 640x400, dark room, one subject. Saved as `docs/pad-results/2026-08-02-center-edge-corpus.jsonl` with a summary beside it.

| presentation | n | ratio range | called Live |
|---|---|---:|---:|
| live, normal distance, glasses on | 12 | 1.40 to 1.47 | 12 |
| live, normal distance, glasses off | 12 | 1.40 to 1.43 | 12 |
| live, arm's length | 12 | 1.36 to 1.41 | 12 |
| live, off-angle (~15 degrees) | 12 | 1.43 to 1.49 | 12 |
| live, close (~20cm) | 12 | 1.26 to 1.37 | 10 |
| vinyl print, normal distance | 12 | 1.02 to 1.16 | 10 |
| vinyl print, close | 12 | 1.12 to 1.21 | 12 |

The lowest genuine capture that passed liveness read **1.31**; the highest attack read **1.21**. Every floor from 1.22 to 1.30 separates the two populations completely on this corpus. 1.25 sits near the middle, 0.04 above the attack ceiling and 0.06 under the genuine floor.

The two close-range live captures that did not pass were refused by the #237 exposure gate, not by this cue.

## What changes

- `MIN_CENTER_EDGE_RATIO`: 1.03 to 1.25, with the corpus and its limits written at the constant.
- `live_signals()`, the test fixture, carried ratio 1.20. This corpus says 1.20 is a **print**, so every test built on that fixture was asserting live-face behaviour about spoof-shaped input. It now uses 1.42, measured.
- A test that pins the floor between the two measured extremes (1.21 must fail, 1.31 must pass), so a future change cannot quietly move it back across either population.

## Verification

- Workspace suite, clippy, fmt, doc lane green.
- Hardware, this build installed as the daemon: the print was denied 4 of 4 by **this cue alone**, reading "IR too flat (center/edge 1.00 to 1.05)", with no third-party model consulted. Genuine authentications at normal and arm's-length distance succeeded; the only genuine refusals were the #237 exposure gate at close range (43% and 21% clipped) and the frontality gate on a turned head, both pre-existing and both retryable. **No genuine capture was refused by the ratio floor.**

## Not tested, and it matters here

- **One subject, one camera, one room, one attack medium.** A face whose geometry reads lower than this subject's, a module with a different emitter baseline, or bright ambient IR could be false-rejected where 1.03 passed. The fallback is a password prompt rather than a lockout, but it is a real cost and it is unmeasured outside this room.
- **A glossy print may still clear 1.25.** The 69-of-70 result on 2026-06-30 used a glossy medium; this corpus used vinyl. This change raises the bar from "essentially no barrier" to "stops the medium that was measured". It does not make the cue proof of a live face, and the constant says so.
- Screens and tablets were not re-measured against the new floor.
- #101 remains the corpus that would let this be tuned against more than one person.


archledger:
Closing: the change is refuted by a measurement this repository already contained, and by a fresh one confirming it.

`docs/pad-results/2026-06-30-ir-liveness-selftest.md` recorded the same vinyl banner on the same Zenbook at 1.02 to 1.58 across 70 frames presented at varied angle and distance, and stated the conclusion plainly: because the banner's ratio rises above the entire genuine range, no floor value accepts the live user and rejects the banner, and 'a naive 1.03 to 1.30 tightening would have been false confidence.'

My corpus missed it because every attack presentation was held square to the camera at two fixed distances. Re-running with the banner deliberately tilted and angled, 20 presentations:

| presentation | n | ratio | accepted Live at the 1.25 floor |
|---|---|---|---|
| banner held square | 24 | 1.02 to 1.21 | 0 |
| banner, varied angle | 20 | up to 1.51 | 4 (1.32, 1.38, 1.39, 1.51) |
| genuine, passed liveness | 58 | 1.31 to 1.49 | n/a |

A floor that blocked the angled banner would need to exceed 1.51, which rejects 58 of 58 genuine captures. The populations overlap; threshold tuning cannot separate them, exactly as the 2026-06-30 result concluded.

What survives from the work: the corpus itself is checked in under `docs/pad-results/2026-08-02-center-edge-corpus.jsonl` and now includes the varied-angle species, so the overlap is recorded with numbers rather than restated. The stale 1.03 references this round found in `suncal` output, `PAD_SELFTEST.md` and ADR 0001 are real and will be fixed separately, since they were wrong about the shipped value even before this PR proposed changing it.

#235 stays open, and the honest reading is that the built-in gate cannot solve it: the fix direction is the opt-in PAD model becoming a default, or a cue that is not a brightness ratio.

PR #245, opened 2026-08-03
Let systemd own the socket, and stop sweeping enrollments every boot

### LINKED ISSUE #244: The first login after a boot cannot reach the daemon, so a fingerprint login leaves the keyring locked
Measured on a ThinkPad X13 Yoga Gen 4, Ubuntu/GNOME, irlume 0.8.0, fingerprint login through `gdm-fingerprint`.

**Symptom.** After a reboot, unlocking with the fingerprint reader works, but the login keyring stays locked: the first app that touches the wallet (a browser, anything using Secret Service) raises the keyring password prompt. Logging in with a password does not do this.

**The timing is the whole bug.**

```
02:21:52  boot
02:22:02  irlumed: loading models (det=..., model=...)
02:22:05  gdm-password][3272]: pam_irlume(gdm-password:auth): conversation failed
02:22:05  gnome-keyring-daemon started, GNOME_KEYRING_CONTROL=/run/user/1000/keyring
02:22:13  irlumed: listening on /run/irlume.sock          <-- 8s after the login
02:22:39  irlumed: UnsealKeyring: OK for 'user' (fingerprint-authenticated), password unsealed
02:22:39  irlumed: connection error: Broken pipe (os error 32)
```

The greeter authenticated the fingerprint at 02:22:05. The daemon did not accept connections until 02:22:13. `pam_irlume.so keyring` on the `gdm-fingerprint` stack asks the daemon to release the TPM-sealed password so `pam_gnome_keyring` can open the wallet; with no socket to talk to, that request cannot be made. The module is `optional` and returns IGNORE on any failure, by design, so **the login proceeds and nothing is logged**. The keyring simply stays locked.

The wiring is correct and the seal is fine. Later requests succeed: `UnsealKeyring: OK ... password unsealed` at 02:22:39, 02:22:50 and 02:23:03. They are just too late to matter, because `pam_gnome_keyring` ran during the login.

**Why the daemon is late.** `irlumed.service` is `Type=simple` with `After=multi-user.target` and there is no `.socket` unit. systemd treats the service as started the instant the process execs, so nothing waits for it, and irlumed binds `/run/irlume.sock` only after loading its ONNX models, about 11 seconds on this machine. The display manager is up and taking authentications during that window.

Every cold boot therefore has a window where face and fingerprint keyring release cannot work, and the first login is almost always inside it.

**The `Broken pipe` at 02:22:39** is the same shape one layer along: the daemon unsealed the password and the client was gone before it could be delivered. Worth understanding as part of a fix rather than separately.

**Fix directions, none of them obviously free:**

1. **Socket activation.** A `.socket` unit makes systemd create `/run/irlume.sock` at boot, so `connect()` succeeds immediately and the request waits in the queue until irlumed accepts it. Standard answer, and it costs a login that waits for model loading rather than one that silently loses its keyring.
2. **Listen before loading models.** Bind the socket first, then load models, and have handlers wait on a readiness gate. Same effect without a second unit, but it changes daemon startup ordering and needs care so early requests block rather than get a wrong answer.
3. **Order irlumed before the display manager** with `Type=notify`. Correct, and it delays every boot to the greeter by the model-load time, which is the worst trade for users who never use face auth.
4. **Retry in `pam_irlume`** for a bounded few seconds when the socket is absent. Smallest change, but it puts the delay in every PAM transaction that finds no daemon, including machines where irlumed is legitimately not running.

My preference is 1 or 2, because they make the socket's existence independent of model loading, which is the actual mismatch. That is a design decision rather than a bug fix, so I have not written it.

**Not established:** whether a logout-and-back-in without a reboot shows the same thing. It should not, since the daemon is already listening by then, and the report that prompted this mentioned both. Worth confirming separately.

Related: the `conversation failed` line from `gdm-password` at 02:22:05 is the `ondemand` probe on the password stack and looks unrelated to the keyring, but it is in the same second and I have not chased it.

archledger:
Closes #244 and #249.

## The bug

`irlumed` bound its own socket, and greeters are ordered after `basic.target` while the service was ordered after `multi-user.target`. On a ThinkPad X13 the greeter authenticated a fingerprint **eight seconds before the socket existed**:

```
02:21:52  boot
02:22:02  irlumed: loading models
02:22:05  gdm-fingerprint authenticates; gnome-keyring-daemon starts
02:22:13  irlumed: listening on /run/irlume.sock      <-- too late
02:22:39  irlumed: UnsealKeyring: OK, password unsealed   <-- nobody left to receive it
```

`pam_irlume.so keyring` had nothing to connect to. That line is `optional` and returns IGNORE on any failure by design, so the login succeeded, **nothing was logged**, and the user met a keyring prompt later from whatever first touched the wallet. Every cold boot had this window and the first login was usually inside it.

## What this changes, and why the three parts are inseparable

**1. systemd owns the socket.** `irlumed.socket` binds `/run/irlume.sock` at `sockets.target`, so a PAM `connect()` lands in the backlog instead of failing. The daemon prefers the inherited descriptor (sd_listen_fds protocol: `LISTEN_PID` must name us, `LISTEN_FDS` must be 1, fd 3) and still binds its own when run directly, so development and any distro without the unit are unaffected.

`Sockets=irlumed.socket` is named **explicitly** rather than relying on the implicit same-name association. That association did not deliver the descriptor here: the daemon fell through to self-binding, which unlinks systemd's socket and binds over it, leaving both systemd and irlumed holding the path. It is silent, and `fuser -v /run/irlume.sock` is how it was caught.

**2. The startup sweep runs once, not every boot.** Asking whether a user needs an IR retag costs a TPM unseal, because the answer lives inside the encrypted enrollment (`EncEnvelope` is `{version, enc}` and nothing else). It ran per user, per boot. The TPM serializes, so that work collided with the login it was delaying. A marker records the embedding space the sweep completed for. It only ever skips work: absent, stale or unreadable means sweep, and nothing security-relevant reads it.

**3. The policy-digest cache is warmed deliberately.** Removing the sweep removed what had been warming the pcrlock branch-digest memo **by accident**, and 8.2s moved onto the first login. This is the "first-login benefit" claim a previous review flagged as unproven and I withdrew; it is now proven, in the opposite direction, by measurement.

`warm_pcrlock_policy_cache()` is explicitly a workaround and says so at the definition. The TCG spec defines these digests as hash recurrences and systemd computes the same values in userspace with zero TPM commands (#248); doing that deletes this function and removes the 8.2s rather than relocating it.

## Measured, ThinkPad X13 (STMicro discrete TPM), cold keyring release

| build | result |
|---|---|
| shipped 0.8.0 | **never happened** after a reboot; keyring prompt |
| naive early bind (first attempt) | 14.26s |
| serve-early, no socket unit | 18.97s (TPM contention with startup) |
| **this branch** | **2.72s, 2.73s** |

`connect()` succeeds in **0.000s with the daemon fully stopped**, and the connection starts it.

## Confirmed on hardware: two reboots and two logouts

Every boot: `using the socket systemd bound` → `serving` → `UnsealKeyring: OK`. Keyring reads `Locked: false`, and the user confirms no prompt when opening a browser. Boot 0 timeline:

```
10:03:15.616  Started fprintd.service
10:03:30.574  irlumed: serving on /run/irlume.sock
10:03:36.995  irlumed: UnsealKeyring: OK for 'fimerlwi', password unsealed
10:03:37.018  gdm-fingerprint: session opened          <-- 23ms later
```

Second login on the same boot: unseal at 10:04:13.105, session at 10:04:13.122, 17ms apart. Retag work performed that boot: **zero**.

## Packaging

The socket unit ships in all four lanes (Fedora spec `%install`/`%files`/`%systemd_post`/`%systemd_preun`, Arch PKGBUILD, Debian nfpm) and as `systemd.sockets.irlumed` in the NixOS module.

## Verification

Workspace suite, clippy, fmt and doc lane green. Tests added: the retag marker's contract (recorded space matches, a different space sweeps again, garbage falls back to sweeping), and a request needing the engine is refused rather than queued while it loads. A previous round's insertion silently stole an existing test's `#[test]` attribute; that is repaired and both tests are confirmed to run.

## Not tested

- **A synthetic test that should not be repeated:** stopping the daemon and connecting measures ~19s, because that forces an on-demand start where the login's own connection triggers the warm-up it then waits for. That is not the boot path; the service still starts from `multi-user.target` and warms before any greeter exists.
- Machines with no pcrlock policy, where `warm_pcrlock_policy_cache` returns immediately.
- Whether socket activation behaves the same on a distro whose greeter is ordered differently; the ordering claim is verified on Fedora/KDE and Ubuntu/GNOME only.

### DIFF
--- a/crates/irlume-core/src/storage.rs
+++ b/crates/irlume-core/src/storage.rs
@@ -493,6 +493,39 @@ pub fn delete(user: &str) -> irlume_common::Result<bool> {
     Ok(existed)
 }
 
+/// Has the one-time IR retag already run for this embedding space?
+///
+/// The retag itself is cheap; ASKING is not. The answer lives inside the
+/// encrypted enrollment, so the daemon used to unseal every user's TPM-sealed
+/// template key at startup just to find nothing to do. On a discrete TPM that
+/// is seconds per user, it happens on every boot, and the TPM serializes, so it
+/// collided with the login it was delaying: a keyring unseal measured 2.70s on a
+/// quiet daemon and 18.97s against that startup (#249).
+///
+/// The marker records the embedding space the sweep completed for. It only ever
+/// SKIPS WORK: a missing, stale or unreadable marker runs the sweep, and no
+/// security decision reads it. Its absence costs a slow startup, never a wrong
+/// answer.
+pub fn retag_done_for(space: &str) -> bool {
+    fs::read_to_string(retag_marker_path())
+        .map(|s| s.trim() == space)
+        .unwrap_or(false)
+}
+
+/// Record that the sweep finished for `space`. Best-effort: failing to write it
+/// costs a repeated sweep next boot, which is the pre-existing behaviour.
+pub fn mark_retag_done(space: &str) {
+    let path = retag_marker_path();
+    if let Some(dir) = path.parent() {
+        let _ = fs::create_dir_all(dir);
+    }
+    let _ = fs::write(&path, format!("{space}\n"));
+}
+
+fn retag_marker_path() -> PathBuf {
+    state_dir().join(".ir-retag-space")
+}
+
 /// Every OS user with an enrollment on this host (the `<user>.json` stems in the
 /// state dir), sorted. For 1:N identify and status reporting. Returns an empty
 /// list if the state dir doesn't exist yet.
@@ -515,6 +548,50 @@ pub fn list_users() -> Vec<String> {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    /// The retag marker skips work and answers a question about work only.
+    ///
+    /// It exists because ASKING whether a user needs a retag costs a TPM unseal,
+    /// and doing that per user at startup collided with the login it delayed
+    /// (#249). Its whole contract: it matches only the space it recorded, an
+    /// absent or unreadable marker means "sweep", and a different embedding
+    /// space means "sweep again". Nothing security-relevant may ever read it,
+    /// which is why it lives beside the enrollments rather than inside one.
+    #[test]
+    fn the_retag_marker_only_matches_the_space_it_recorded() {
+        let dir = std::env::temp_dir().join(format!("irlume-retag-{}", std::process::id()));
+        let _ = fs::remove_dir_all(&dir);
+        fs::create_dir_all(&dir).unwrap();
+        std::env::set_var("IRLUME_STATE_DIR", &dir);
+
+        // Nothing recorded yet: the sweep must run.
+        assert!(
+            !retag_done_for("raw"),
+            "an absent marker must not skip the sweep"
+        );
+
+        mark_retag_done("raw");
+        assert!(
+            retag_done_for("raw"),
+            "the recorded space must be recognised"
+        );
+
+        // An adapter change moves the space, so the sweep is owed again.
+        assert!(
+            !retag_done_for("adapter:deadbeef"),
+            "a different embedding space must run the sweep again"
+        );
+
+        // Garbage on disk is not a match, so it fails towards doing the work.
+        fs::write(dir.join(".ir-retag-space"), b"\x00not a space").unwrap();
+        assert!(
+            !retag_done_for("raw"),
+            "an unreadable or unexpected marker must fall back to sweeping"
+        );
+
+        std::env::remove_var("IRLUME_STATE_DIR");
+        let _ = fs::remove_dir_all(&dir);
+    }
 
     #[test]
     fn legacy_format_migrates_to_one_profile() {
--- a/crates/irlume-core/src/tpm.rs
+++ b/crates/irlume-core/src/tpm.rs
@@ -1168,6 +1168,36 @@ pub fn seal_with_pcrlock(secret: &[u8], nv_index: u32) -> Result<SealedEnvelope>
 /// machine has none: `pcrlock.json` absent (the common case), unparseable,
 /// rejected by the empty-policy guard, or carrying no NV index. [`seal`] uses
 /// this to decide whether the Tier 2 rung exists on this machine.
+/// Derive the pcrlock branch digests now, so the first unseal does not.
+///
+/// The derivation costs one trial TPM session per predicted branch, measured at
+/// 8.2s on a ThinkPad X13, and its result depends only on the prediction file
+/// (#246). It is memoized per process, so SOMETHING has to pay for it first.
+/// That used to be the daemon's startup enrollment sweep, by accident: removing
+/// the sweep (#249) moved the cost onto the first login, which measured 10.88s
+/// against 2.71s warm.
+///
+/// Calling this from daemon startup keeps the warm path without unsealing
+/// anyone's enrollment to get it. Best-effort by design: every failure here
+/// means the next real unseal derives them itself, exactly as before.
+///
+/// This is a workaround for computing the digests on the TPM at all. The TCG
+/// spec defines them as plain hash recurrences and systemd computes the same
+/// values in userspace with zero TPM commands (#248); doing that removes this
+/// function's reason to exist.
+pub fn warm_pcrlock_policy_cache() {
+    let Some(_) = pcrlock_provisioned() else {
+        return; // Not sealed under pcrlock; nothing to derive.
+    };
+    let Ok(plock) = read_pcrlock_json() else {
+        return;
+    };
+    let Ok(mut ctx) = open_context() else {
+        return;
+    };
+    let _ = build_super_pcr(&mut ctx, &plock);
+}
+
 pub fn pcrlock_provisioned() -> Option<u32> {
     let plock = read_pcrlock_json().ok()?;
     u32::try_from(plock.nv_index?).ok()
--- a/crates/irlume-daemon/src/main.rs
+++ b/crates/irlume-daemon/src/main.rs
@@ -314,6 +314,11 @@ fn main() {
             // delaying, taking a keyring unseal from 2.70s to 18.97s on a
             // discrete TPM (#249). The marker only skips work; a missing or
             // stale one just runs the sweep as before.
+            // Derive the pcrlock branch digests before any login needs them.
+            // The sweep below used to warm this by accident; skipping the sweep
+            // moved 8.2s onto the first keyring release (#249, #246).
+            irlume_core::tpm::warm_pcrlock_policy_cache();
+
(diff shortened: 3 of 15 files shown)


archledger:
Holding this rather than merging. The review finding is real and I am not landing a concurrency change to the credential-release path on the strength of one reading at 03:00.

**The finding.** Binding early creates a world-connectable listener that nothing drains for the length of startup, 2s here and 11s on the ThinkPad. Any local process can fill the kernel backlog in that window; the greeter's request then cannot enter the queue, stalls on the connect timeout, and `pam_irlume` returns IGNORE with the keyring still locked. That is the failure this PR exists to close, made reproducible on demand. The old placement did not have that shape, because the listener began draining the moment it existed.

Severity, for the record: it does not lock anyone out (the password path is unaffected) and the un-attacked case works, measured 2.11s to 0.05s. But a fix whose benefit disappears under a trivial local action is not a fix, and this is authentication software.

**The better design, which I would rather build than patch this one into shape.** Start accepting immediately after the bind, and answer from the connection thread while the engine is still loading:

- `UnsealKeyring` needs `irlume_core::keyring` and the peer credentials only. It never touches the engine; I checked every line of its arm. It can be served correctly from the moment the socket exists, which fixes the keyring case FASTER than this PR does, with no waiting for models at all.
- Requests that do need the engine get a clear "still starting" response instead of sitting in a queue, so a greeter's face attempt falls through to the password immediately, as it does today.
- Draining from the start removes the backlog primitive entirely, so the per-uid bounds that already exist post-start are the only limit that matters.

That is strictly better than what is here: it fixes the reported bug sooner, it does not introduce the undrained window, and it does not need a startup connection buffer. It also touches request dispatch and credential release, which is exactly the code to write rested and to test properly rather than at the end of a long session.

Leaving the PR open with the measurements and the regression test, since both stand on their own: the A/B (2.11s to 0.05s to first connect) and a test that fails on a mutant which delays the bind. #244 stays open with the design recorded.

archledger:
Rebased onto `main` (so it carries #247's memo) and measured on the ThinkPad, where the original report came from. This makes the case for the redesign concrete rather than theoretical.

**A keyring request issued the instant the daemon starts is now answered, and takes 14.26s:**

```
connect() succeeded at 0.05s after restart
keyring request issued immediately: answered in 14.26s, response=PasswordUnsealed
```

It queues behind the whole startup (socket ready 08:06:29, serving 08:06:40) and then pays the 2.7s unseal. So this branch does fix the reported bug: after a reboot the keyring unlocks instead of prompting. It also makes that login wait 14 seconds, which is worse than the 9 the user originally complained about.

**That is the argument for the design on #244, now with a number.** `UnsealKeyring` needs `irlume_core::keyring` and the peer credentials; it never touches the engine. Answered on the connection thread while models load, it would cost ~2.7s instead of 14.26s, and draining from the start also removes the undrained-listener problem this PR was held for. Both objections have the same fix.

Still not merging this. It trades a locked keyring for a slow login and leaves the backlog window open; it is installed on the reporting machine as a diagnostic only, with the packaged binary backed up.

**What the diagnostic is worth:** it confirms end to end that the startup race is the entire cause of the reboot symptom, which no amount of code reading establishes.

archledger:
Rebuilt to the design I described, measured on both machines, and the result splits: it is right on one and NOT sufficient on the other. Still unmerged.

**The design now in the branch.** Engine construction moved to its own thread, so `main` reaches the accept loop immediately: the socket is drained from the first moment (closing the undrained-listener objection), and until the engine is published `serve` answers from an engine-free path. `unseal_keyring` is free-standing, so the claim that it needs no engine is checkable rather than asserted; every authorization in it is a property of the request.

**Zenbook, restart to answer:**

```
connect()                            0.06s
keyring request issued immediately   0.40s   PasswordUnsealed
request needing the engine           0.00s   "still starting; retry, or use your password"
```

0.40s against 14.26s for the previous approach, and a face attempt now falls through to the password instantly instead of waiting out startup.

**ThinkPad, same test: 18.97s.** Worse than the 14.26s it replaces.

The reason is #249: that machine's startup is TPM-bound, because it unseals every enrollment's template key to check for legacy IR scans. The TPM serializes, so answering the login's request *while* startup hammers it costs more than making it queue behind it. A quiet daemon on the same machine answers in 2.70s.

So serving without the engine was necessary and not sufficient: on TPM-slow hardware the bottleneck was never the engine. This branch should land together with #249, not before it, or it trades a 14.26s login for a 19s one on exactly the machine that reported the bug.

The tests and the Zenbook numbers stand on their own, and there is a new one covering the refusal path, which also caught that my insertion had silently stolen an existing test's `#[test]` attribute (defect pattern 35, twice in one session).

archledger:
Both merge-blocking findings applied in `49693cf`, and the first one is the more instructive.

**The socket was installed everywhere and enabled nowhere.** `[Install]` only describes the symlink `systemctl enable` would create; Fedora's preset decides, and Arch, the universal `.deb` and the PPA each enable units explicitly. The PPA lane did not install the unit at all. And `Sockets=irlumed.socket` makes the socket accompany a service start, which happens at the service's existing late boot point, so it does not put the socket into `sockets.target`. Every lane would have shipped with the exact window this PR closes.

Worth saying plainly: **I verified the fix on a machine where I had run `systemctl enable --now irlumed.socket` by hand**, then measured 2.72s and called it confirmed. The measurement was real and the packaging was not.

**The repository already had a guard for this, and it fired.** `scripts/check-packaging-parity.sh` exists because the 0.6.x rpm installed a timer it left out of `%files` and the 0.7.0 audit found the PPA missing the reconcile timer. It failed this branch in the `systemd units` lane and named `irlumed.socket` against `packaging/ppa/debian/rules`. I polled CI, read the summary rather than the count, and moved on to the review. That is my error, not the guard's.

Verified after the fact that it still fires: dropping the socket from the PPA lane again produces `MISS irlumed.socket packaging/ppa/debian/rules` and a non-zero exit, then restored.

**The retag marker could certify an incomplete sweep.** A per-user `load` or `save` failure still let the marker be written, which retires the migration permanently for that user. It is now written only when every user was reached, and the daemon logs that the sweep will run again when it was not.

**On the other CI failure**, `stream_record::tests::a_failed_retirement_spends_no_attempt` ("free again: Busy"): this branch does not touch `irlume-camera` at all, main's CI is green, and locally it is 0 failures in 25 isolated runs and 0 in 6 whole-crate parallel runs. Treating it as a pre-existing CI-only flake rather than assuming; if it recurs on the re-run it wants its own issue with the flake rate measured, per the pattern where a local suite is not CI's suite.

PR #247, opened 2026-08-03
Derive the pcrlock branch digests once, not on every unseal

### LINKED ISSUE #246: A keyring unseal costs 11 seconds on one machine and 0.36 on another, and the login waits for it
Measured 2026-08-03 on two machines, prompted by a user noticing that logging back in after a logout stalls for about 9 seconds while a login after a reboot does not.

**The measurement.** Three consecutive `UnsealKeyring` round-trips over `/run/irlume.sock`, timed by the client, on an otherwise idle system:

| machine | TPM | seal policy | bound PCRs | unseal round-trip |
|---|---|---|---|---|
| ThinkPad X13 Yoga Gen 4 | `STM0176` (STMicro, discrete) | pcrlock NV, Tier 2 | `[0, 1, 2, 3, 5, 7, 13, 15]` | **10.88s, 10.89s, 10.88s** |
| Zenbook S14 (UX5406S) | `MSFT0101` (firmware TPM) | pcrlock NV, Tier 2 | `[2, 3, 13, 15]` | 0.40s, 0.36s, 0.36s |

Same mechanism and same tier, a factor of 30 apart, and reproducible to within 0.01s across attempts.

**Why it is user-visible.** `pam_irlume.so keyring` runs inside the PAM auth stack, so the login blocks on it. The journal timeline for a fingerprint login on the ThinkPad:

```
03:00:44.714  fprintd started
              (11.5s, nothing logged)
03:00:56.251  irlumed: UnsealKeyring: OK for 'fimerlwi', password unsealed
03:00:56.274  session opened
```

The user reports roughly 9 seconds of dead time between authenticating and reaching the desktop, which matches.

**The reported asymmetry is the same bug seen from the other side.** After a reboot they see no delay, because the daemon is not listening yet (#244) and the request fails instantly; the keyring is then left locked. After a logout the daemon is reachable, so the unseal runs, the keyring unlocks, and the login pays 11 seconds. Fixing #244 without this one would give that machine a working keyring and an 11-second login.

**What is not yet established.** Whether the cost is the discrete TPM being slow at policy sessions, the four extra bound PCRs, or something in irlume's unseal path such as a Tier 2 attempt that falls back. The consistency to 0.01s across three runs argues for a deterministic sequence rather than variable hardware latency, and is worth understanding before assuming it is simply slow silicon. Timing the individual TPM commands (policy session start, NV read, PCR extend/read, unseal) would separate those.

**Directions, once the cause is known:**

- If the work is genuinely unavoidable, it should not sit in the auth stack synchronously. Releasing the keyring password is not a decision the login has to wait on: the module could set the token when it arrives and let the session-phase keyring unlock happen slightly later, or the daemon could hold a short-lived post-authentication cache so the second surface in a session does not pay it again.
- If the cost comes from binding eight PCRs, that is a threat-model trade worth stating explicitly rather than inheriting from whatever the arming machine happened to measure.

Related: #244 (the socket is not bound during early boot, so this path never runs after a reboot).

archledger:
Closes #246.

## The bug

`build_super_pcr` derives each predicted PCR branch's policy digest with a trial TPM session, one per branch. The result depends only on `/var/lib/systemd/pcrlock.json`: same prediction, same digests, every time. It was recomputed on every unseal.

`pam_irlume.so keyring` runs inside the PAM auth stack, so that cost is login latency. On a ThinkPad X13 the journal showed 11.5 seconds between `fprintd` starting and the session opening, with the unseal completing 23ms before it. The user's report was "about 9 seconds signing into the DE".

## Why the two machines differ so much

Branch counts read from each machine's own prediction file:

| machine | TPM | PCRs | multi-value branches | per branch |
|---|---|---|---|---|
| ThinkPad X13 Yoga Gen 4 | STMicro, discrete | 8 | **17** | ~0.64s |
| Zenbook S14 | firmware (MSFT0101) | 4 | 4 | ~0.09s |

Four times the branches on a TPM roughly seven times slower per operation accounts for the 30x gap, and it explains the determinism: a fixed number of TPM round-trips rather than variable latency.

## The fix

Memoize the derivation, keyed on the prediction's own content, so `systemd-pcrlock make-policy` invalidates it by changing the key rather than by anyone remembering to clear it.

In memory only. A persisted cache would need its own argument about what a tampered entry can do; my reading is that `policy_authorize_nv` binds the final digest to the NV index, so injected branches would make the unseal FAIL rather than succeed wrongly, but that is reasoning I have not verified and it is not needed to remove the repeat cost.

## Measured

Same-session A/B on the ThinkPad, swapping only the binary, three unseals each:

```
PRE-FIX (packaged 0.8.0):  10.88s  10.89s  10.88s
WITH MEMO:                  2.71s   2.70s   2.72s
```

**8.17 seconds off every unseal on that machine.** Zenbook, same method: 0.36s to 0.20s. On both, every run after the first is identical to the first, because the daemon's own startup work warms the memo, so the first login of a boot benefits too.

## What is left, and it is not this change

2.7 seconds remain on the ThinkPad and are still deterministic, so something else in the unseal path costs a fixed number of TPM round-trips. The candidates are the SRK primary (`with_srk`, and `CreatePrimary` is expensive on slow TPMs) and the live policy replay, which is one `PolicyPCR` plus a `PolicyPCR`/`PolicyOR` pair for each of the seven multi-value PCRs. Worth its own issue and its own measurement rather than a guess folded in here.

## Interaction with #244

These have to ship together. Serving `UnsealKeyring` early fixes the reboot case where the socket does not exist yet, but on this hardware that would have left the login blocked for 11 seconds instead of prompting later. #244 alone trades a locked keyring for a slow one.

## Not tested

- Whether a `systemd-pcrlock make-policy` mid-run correctly invalidates the memo. The key is derived from the prediction content, so it should, but I did not re-run `make-policy` on either machine to watch it happen.
- Machines with a single-value prediction on every PCR, where the loop does no work and the memo is free.

### DIFF
--- a/crates/irlume-core/src/tpm.rs
+++ b/crates/irlume-core/src/tpm.rs
@@ -927,6 +927,7 @@ fn digest_list(digests: &[Digest]) -> Result<DigestList> {
 }
 
 /// A single-value PCR group plus the ordered multi-value PCRs of a super policy.
+#[derive(Clone)]
 struct SuperPcr {
     /// Single-value PCRs, ascending (replayed live as one PolicyPCR at unseal).
     singles: Vec<u32>,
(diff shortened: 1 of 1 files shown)


archledger:
Review finding applied in `2e543b0`, and it was the right call.

**The `u64` key is gone.** It was a `DefaultHasher` digest, and while a collision is absurdly unlikely, the failure it admits is serving branch digests derived from a different prediction under credential release. A prediction is a handful of PCRs with a few hex strings each, so the whole thing is now kept and compared exactly. There is no performance argument for the hash at that size.

The test pins what the key has to separate: listing order must NOT invalidate a memo, while a changed value, an added branch, the same values moved to another PCR, and a reordering of branches within one PCR must all miss it, the last because branch order feeds the derivation.

**On the first-login claim you flagged as unsupported:** you are right that the PR asserted more than it showed. What I measured is that every run including the first is identical (2.71/2.70/2.72 on the ThinkPad after a restart, 0.20 three times on the Zenbook), which is consistent with the memo being warm before the first request but does not prove what warmed it. I have not traced which startup path performs the first pcrlock unseal. The claim is not load-bearing for this change, so I have stopped short of asserting it; if #244 ends up depending on a warm memo at first login, that dependency needs its own measurement.

**Third machine, since it was suggested:** archhost has a 3-branch prediction on a firmware TPM, which the model says should never have been slow. I could not measure it, because its unseal fails outright with `cannot read /var/lib/systemd/pcrlock.json: Permission denied` even though the daemon is uid 0 with `CAP_DAC_OVERRIDE`, the path is `0644` under `0755` directories, and reading it through the daemon's own mount namespace works. That machine still runs 0.7.0, so it may already be fixed. Recorded on #246; it is a separate bug and not evidence about this one.

So the measurement behind this PR is still the ThinkPad same-session A/B: **10.88s to 2.71s**, with the Zenbook as a second, smaller confirmation. The remaining 2.7s is #248.

PR #252, opened 2026-08-03
Compute the pcrlock policy digests in software, not on the TPM

### LINKED ISSUE #248: A keyring unseal still costs 2.7 deterministic seconds after the branch-digest memo
Follow-up to #246, which took the ThinkPad X13 from 10.88s to 2.72s per `UnsealKeyring` by memoizing the pcrlock branch-digest derivation. The remainder is worth its own measurement.

**What is left.** Same machine (STMicro discrete TPM), with the memo in place, four consecutive unseals: 2.72s, 2.72s, 2.71s, 2.72s. Still deterministic to 0.01s, so it is a fixed number of TPM round-trips rather than variable latency, and the memo is warm from the first request (the daemon's own startup work primes it).

For scale, the same operation on a firmware TPM with a four-branch prediction is 0.20s.

**Where it plausibly goes**, in `unseal_pcrlock`:

1. `with_srk`, which creates the storage root key primary. `TPM2_CreatePrimary` is one of the most expensive commands on a discrete TPM and it runs on every unseal. A persistent handle, or caching the primary for the daemon's lifetime the way #246 caches the digests, would remove it if this is the bulk.
2. The live policy replay: one `PolicyPCR` for the single-value group, then a `PolicyPCR` and a `PolicyOR` for each multi-value PCR. This machine has seven multi-value PCRs, so that is fifteen policy commands before `PolicyAuthorizeNV` and `Unseal`.
3. `Load` of the sealed object, and the NV index read.

Splitting those is a matter of timing each command rather than reasoning about them; the arithmetic in #246 was right about the cause precisely because the branch count was counted rather than assumed, and the same discipline applies here.

**Why it matters.** `pam_irlume.so keyring` runs inside the PAM auth stack, so this is login latency on every fingerprint or face login that releases the keyring. 2.7s is a great deal better than 11s and still noticeable.

**Worth deciding alongside:** whether the release has to be synchronous with authentication at all. The password is needed by `pam_gnome_keyring`, which runs later in the same transaction, so there may be room to start the unseal earlier or in parallel rather than shortening it.

Related: #246 (the memo), #244 (the socket is not bound during early boot, so this path does not run at all after a reboot).

archledger:
Closes #248.

## What was happening

The pcrlock branch digests were derived by running **one trial TPM session per predicted branch**. On a ThinkPad X13 with 17 branches on a discrete STMicro TPM that is 8.2s, paid on every unseal until #247 memoized it, and then paid by the first login of each boot once #245 removed the enrollment sweep that had been warming the memo by accident.

Two workarounds stacked on one avoidable cost.

## It was never necessary

TCG TPM 2.0 Library Part 3 defines these as hash recurrences:

- §23.7 eq. 20, PolicyPCR: `policyDigest_new := H(policyDigest_old || TPM_CC_PolicyPCR || pcrs || pcrDigest)`
- §23.6 NOTE 2, PolicyOR: the running digest **resets to zero**, then `H(0…0 || TPM_CC_PolicyOR || digests)`
- §23.19: `TPM2_PolicyGetDigest` *allows* the TPM to precompute a policy. It does not require it.

systemd computes the same values the same way with zero TPM commands (`tpm2_calculate_policy_pcr` / `_or` in `src/shared/tpm2-util.c`). The Rust `tss-esapi` crate has no offline calculator, so this implements the extends directly.

## The oracle

The TPM has been computing these all along, so the replacement is correct only if it agrees with the chip. `software_digests_match_a_trial_session` compares software against a real trial session over three shapes chosen so a mistake cannot hide:

- a single-PCR `PolicyPCR`;
- a **multi-PCR selection**, where a `TPML_PCR_SELECTION` marshalling error would show (count, `TPMI_ALG_HASH`, `sizeofSelect`, and the bitmap where PCR n is bit n%8 of byte n/8);
- a `PolicyPCR` → `PolicyOR` → `PolicyPCR` chain, where mishandling the OR's reset-to-zero would show.

**Passes on all three TPMs available here**, which matters because a marshalling bug could easily appear on one implementation and not another:

| machine | TPM | result |
|---|---|---|
| Zenbook S14 | Intel PTT firmware (`MSFT0101`) | ok |
| ThinkPad X13 Yoga Gen 4 | STMicro discrete (`STM0176`) | ok |
| archhost | firmware (`MSFT0101`), Arch | ok |

## Measured

| | before | after |
|---|---|---|
| ThinkPad, first unseal after a cold start | 10.88s | **2.74s** |
| ThinkPad, steady state | 2.71s | 2.71s |
| Zenbook, cold start | ~0.36s | 0.26s, then 0.20s |

There is no longer a cold-versus-warm difference, because there is nothing to warm. The 8.2s did not move; it stopped existing.

## Net effect is deletion

Removed: the `SUPER_PCR_MEMO` and its key type, `pcrlock_key`, the memo's identity test, `warm_pcrlock_policy_cache()` and its call from daemon startup. `build_super_pcr` no longer takes a TPM context and is pure.

`trial_replay` survives **only** as the oracle and is `#[cfg(test)]`, so putting a trial session back on a hot path has to be deliberate rather than accidental.

This also removes the coupling that caused trouble in #245: deleting the startup sweep silently moved 8.2s onto the first login, because the sweep had been warming the memo. With the derivation pure, nothing warms anything.

## Not tested

- A prediction with more than 8 alternatives on one PCR. TCG Part 3 §23.6 NOTE 3 says a TPM need only support 8 tagged digests in a `PolicyOR`, and systemd enforces that with `E2BIG`; irlume does not check it, and no prediction here exceeds 4.
- Hash algorithms other than SHA-256. The marshalling hardcodes `TPM_ALG_SHA256` and a 3-byte select, matching every prediction observed (`pcrBank: sha256`).
- Whether the digests remain correct across a `systemd-pcrlock make-policy` that changes the bank; that is the same untested axis as above.

### DIFF
--- a/crates/irlume-core/src/tpm.rs
+++ b/crates/irlume-core/src/tpm.rs
@@ -1516,15 +1516,21 @@ mod tests {
     /// command code shows up as a mismatch rather than as a later unseal
     /// failure nobody can attribute.
     #[test]
-    #[ignore = "requires a real TPM; run as root: cargo test -p irlume-core --lib tpm:: -- --ignored"]
+    #[ignore = "real TPM, run as root and BY NAME so it cannot be lost among the other ignored \
+                TPM tests: sudo <test-binary> software_digests_match --ignored --nocapture"]
     fn software_digests_match_a_trial_session() {
-        let mut ctx = match open_context() {
-            Ok(c) => c,
-            Err(e) => {
-                eprintln!("SKIP: no TPM context ({e})");
-                return;
-            }
-        };
+        // FAILS rather than skips when the TPM is out of reach. This is the only
+        // thing standing behind replacing a TPM computation with arithmetic, and
+        // a test that reports success without observing the hardware it is named
+        // for is worse than no test: it is already `#[ignore]`, so anyone running
+        // it has asked for hardware.
+        let mut ctx = open_context().unwrap_or_else(|e| {
+            panic!(
+                "this test exists to compare software digests against a real TPM, \
+                 and no TPM context could be opened ({e}). Run it as root on a \
+                 machine with a TPM; do not treat this as a skip."
+            )
+        });
 
         // A single-value PolicyPCR, the shape the singles group takes.
         let one = pcr_composite_digest(&[[0x11; 32]]).unwrap();
--- a/crates/irlume-core/src/tpm.rs
+++ b/crates/irlume-core/src/tpm.rs
@@ -896,9 +896,96 @@ enum PolicyStep {
     Or(Vec<Digest>),
 }
 
+/// `TPM_CC_PolicyPCR`, TCG TPM 2.0 Library Part 2 "Structures", command codes.
+const TPM_CC_POLICY_PCR: u32 = 0x0000_017F;
+/// `TPM_CC_PolicyOR`.
+const TPM_CC_POLICY_OR: u32 = 0x0000_0171;
+/// `TPM_ALG_SHA256`.
+const TPM_ALG_SHA256: u16 = 0x000B;
+
+/// A `TPML_PCR_SELECTION` in the TPM's own wire encoding.
+///
+/// The policy digest is taken over the marshalled structure, so this has to
+/// match the TPM byte for byte: `UINT32 count`, then per selection a
+/// `TPMI_ALG_HASH` (UINT16), a `UINT8 sizeofSelect`, and that many bitmap
+/// bytes, PCR n living in bit n%8 of byte n/8. Part 2 §10.9.7.
+fn marshal_pcr_selection(pcrs: &[u32]) -> Vec<u8> {
+    // Three bytes covers PCRs 0-23, which is every PCR irlume will select; the
+    // TPM reports sizeofSelect=3 for SHA-256 on every part we have measured.
+    let mut bitmap = [0u8; 3];
+    for &p in pcrs {
+        if let Some(byte) = bitmap.get_mut((p / 8) as usize) {
+            *byte |= 1 << (p % 8);
+        }
+    }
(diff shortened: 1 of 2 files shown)


archledger:
Review finding applied, and it was the right catch on the one test that matters here.

**The oracle could pass without a TPM.** It returned early with a SKIP when `open_context()` failed. That test is the only thing standing behind replacing a TPM computation with arithmetic, so a version of it that reports success without observing the hardware it is named for is worse than not having it. It now panics with the underlying error instead.

**Verified both directions**, because a guard nobody has seen fail is a guard nobody has tested:

```
IRLUME_TCTI=device:/dev/does-not-exist  ->  FAILED  ("do not treat this as a skip")
real TPM                                ->  ok
```

**Re-run on all three machines with the hardened test:**

| machine | TPM | result |
|---|---|---|
| Zenbook S14 | Intel PTT firmware | ok, 0.12s |
| ThinkPad X13 Yoga Gen 4 | STMicro discrete | ok, 0.51s; FAILED with an unusable TCTI |
| archhost | firmware, Arch | ok, 0.25s |

The differing durations are themselves evidence the TPM is being exercised rather than skipped: a skip returns in ~0.00s, which is exactly what the unreachable-TCTI run now shows.

**The ignore message also documented a command that runs every ignored TPM test**, including the seal/unseal round trips that touch real hardware state. It now names the isolated invocation by test name.

On the three things you flagged as unverifiable: a PCR selection wider than three bytes, hash banks other than SHA-256, and predictions with more than eight alternatives per PCR are all still untested, and all three are listed in the PR's "not tested" section rather than claimed. The eight-alternative limit is TCG Part 3 §23.6 NOTE 3; irlume does not reject it early, and as you say it fails closed during live replay rather than producing a wrong policy.

Issue #250, opened 2026-08-03
We seal the login password; every comparable project seals a derived secret instead


archledger:
Research finding, and the most consequential one from a survey of how other projects solve passwordless keyring unlock. This is a design question, not a bug, and it wants a decision rather than a patch.

**What irlume does.** `/var/lib/irlume/keyring/<user>.json` holds the user's actual Unix login password, TPM-sealed. `pam_irlume` releases it so `pam_gnome_keyring`/`pam_kwallet` can open the wallet with the password they already expect.

**What everyone else does.** Three designs exist, and we are on the one with the largest blast radius:

| design | what is sealed | if it leaks | who |
|---|---|---|---|
| stash the login password | the Unix password | **root escalation** | irlume, Windows "convenience PIN" |
| derive a stable token | `hex(HMAC(master_key, username))` | keyring only | pinpam |
| rekey to a random secret | a random key | keyring only | ChromeOS, Android, systemd-homed |

**pinpam** (github.com/RazeLighter777/pinpam, GPL-3.0, Rust, TPM2, in the AUR and nixpkgs) is the closest existing project to irlume, and its module header describes stamping `PAM_AUTHTOK` with "a stable, TPM-bound per-user wallet token". It also uses the TPM's PinFail NV index for per-user brute-force counting, marked read-only at creation so even root cannot reset it, which avoids the global DA lockout that `systemd-cryptenroll(1)` warns about.

**Canonical shipped our design and is moving to pinpam's.** authd#1190 ("pam: Set PAM_AUTHTOK on successful authentication") merged 2026-07-07, closing an issue titled "GNOME Keyring Unable to be Automatically Unlocked on Login". In its review, Marco Trevisan argued for a broker-minted opaque token over the login password, and Adam Dombeck proposed generating a secret on first authentication and sealing it with the TPM via systemd credentials.

**systemd upstream describes the derived design too.** Poettering, systemd#26306: "we could unlock/derive some secondary key from the fido2 token, and propagate that into the PAM stack. This could then be picked up by things like gnome keyring, believing it was the actual user's password, even though it isn't."

**Microsoft names our mode and disrecommends it.** Windows Hello proper never stores the password; the mode that does is called "convenience PIN", and Microsoft's own docs say a user with one "is using a password stuffer, which doesn't have any of the security qualities of Windows Hello for Business". The Azure baseline recommends disabling it.

[...]

PR #253, opened 2026-08-03
Seal the KDE wallet key instead of the login password

archledger:
Seals the 56-byte key `ksecretd` opens the KDE wallet with, instead of the user's Unix login password. Closes the KDE half of #250.

## Why this shape

KDE's secret store never receives the password. `pam_kwallet5` runs it through PBKDF2-HMAC-SHA512 (50000 rounds, 56 bytes, salt at `~/.local/share/kwalletd/kdewallet.salt`) and writes the result to a pipe; `ksecretd`'s `waitForHash()` reads exactly those bytes and opens the wallet with them.

So the wallet is already keyed to a derived value. Sealing that value instead of the password costs nothing: no re-key, no migration of the wallet, and a typed password still opens it through the normal `pam_kwallet5` path because the KDF input has not changed. What changes is that a leaked envelope yields a wallet key, useless anywhere else, rather than a Unix password that is not.

gnome-keyring has no equivalent. `pam_gnome_keyring` passes the password string and `gkd_login_unlock()` builds the credential from that string, so there is no derived intermediate to seal in its place. This PR is KDE only and leaves the GNOME path exactly as it was.

## Measured, not assumed

`scripts/kwallet-handoff-check.sh` runs the whole chain against real software: a key **irlume derives**, handed over by the helper **irlume ships**, opening a wallet a real **ksecretd** guards. On Fedora 44 KDE:

```
PASS: the derived key opened the wallet
PASS: a key from the wrong password was refused
stress: 5/5 opened, daemons 0 -> 0, caller fds 4 -> 4
```

The negative control matters as much as the positive one; without it the check could be passing for some other reason. The derivation is separately pinned against a vector computed outside this code, so the three wire constants have something independent to check against.

## Delivery

`pamOpen` looked like it would avoid all process plumbing, since it takes the raw hash over D-Bus. It belongs to the wrong process. In a live Plasma session `org.kde.kwalletd6` is owned by `/usr/bin/kwalletd6`, the compatibility shim, while the store is `ksecretd`; the shim translates to the Secret Service, which has no unlock-with-a-raw-hash operation, and `ksecretd` does not expose `pamOpen` at all. The startup pipe is the only route.

`irlume-kwallet-init` does that handoff. It is a separate program because `pam_irlume` is loaded into sshd, login and every greeter and has no fork, exec or privilege dropping today. It is launched from the PAM process rather than from `irlumed` so `ksecretd` lands in the session's cgroup, where `pam_kwallet5` puts it, instead of dying with an irlume restart.

No PAM stack change is required. Setting `PAM_KWALLET5_LOGIN` is what makes Plasma's `plasma-kwallet-pam.service` deliver the session environment, and it is also the interlock with `pam_kwallet5`: both its authenticate and open_session phases return early with "we were already executed" when that variable is present.

## Detection

The daemon decides by inspection. A wallet key is chosen only for a home with a wallet salt and no gnome-keyring `login.keyring`, because a machine running both would otherwise be left with a GNOME keyring nothing can unlock. Anything ambiguous stays on the login password.

## Bugs this pass introduced and then fixed

Recorded because each was caught by a control rather than by reading:

- The forked wallet daemon inherited the helper's stdout, so any pipe or `$(...)` capture blocked forever waiting for a writer that outlives the helper by design. Found by hanging exactly that way.
- Both rewrite paths in `reseal_password` build a fresh envelope via `tpm::seal`, which defaults the kind, so a wallet-key envelope could be silently downgraded to a login-password one. Both now stamp it, verified by breaking each in turn.
- That test could not initially see the tier-climb path at all: sealing normally lands on the machine's best tier, so the reseal returned `Unchanged` and the rewrite never ran. It now seeds at the weakest tier on purpose; before that, breaking the stamp went unnoticed.
- `reseal_password` briefly required a home directory for every envelope, which would have failed resealing for any account NSS cannot resolve a home for.
- The packaging parity check knew nothing about helper programs, the same blind spot that let the socket unit ship unenabled. It now checks all five lanes, verified by deleting the Arch line and watching it fail.

## Not covered

- Not exercised on a real login yet. The handoff check drives the same code paths a login does, but it builds its own session rather than going through a greeter.
- KDE only. The GNOME side of #250, which needs a wallet re-key and a recovery wrapping, is untouched and still open.
- The handoff check needs a KDE wallet daemon, so it runs on the Fedora KDE machine and is skipped elsewhere.
- Existing armed users are not migrated. Their envelopes keep working as login passwords; re-running `irlume keyring arm` on a KDE-only machine moves them over.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -992,6 +992,14 @@ name = "irlume-fingerprint"
 version = "0.8.0"
 
 [[package]]
+name = "irlume-kwallet-init"
+version = "0.8.0"
+dependencies = [
+ "irlume-common",
+ "libc",
+]
+
+[[package]]
 name = "irlume-liveness"
 version = "0.8.0"
 dependencies = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -558,6 +558,7 @@ dependencies = [
  "block-buffer 0.12.1",
  "const-oid 0.10.2",
  "crypto-common 0.2.2",
+ "ctutils",
 ]
 
 [[package]]
@@ -809,6 +810,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
 
 [[package]]
+name = "hmac"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6303bc9732ae41b04cb554b844a762b4115a61bfaa81e3e83050991eeb56863f"
+dependencies = [
+ "digest 0.11.3",
+]
+
+[[package]]
 name = "home"
 version = "0.5.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -952,6 +962,7 @@ dependencies = [
  "base64 0.23.0",
  "irlume-common",
  "irlume-vision",
+ "pbkdf2",
  "rand 0.10.2",
  "rsa",
  "serde",
@@ -1467,6 +1478,16 @@ dependencies = [
  "base64ct",
  "rand_core 0.6.4",
  "subtle",
+]
+
+[[package]]
+name = "pbkdf2"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "112d82ceb8c5bf524d9af484d4e4970c9fd5a0cc15ba14ad93dccd28873b0629"
+dependencies = [
+ "digest 0.11.3",
+ "hmac",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
     "crates/irlume-vision",
     "crates/irlume-liveness",
     "crates/irlume-core",
+    "crates/irlume-kwallet-init",
     "crates/irlume-auth",
     "crates/irlume-fingerprint",
     "crates/irlume-daemon",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ sha2       = "0.11"     # PolicyAuthorize digest math (signed-PCR path)
 rsa        = { version = "0.9", features = ["sha2"] }  # parse systemd PCR-signing pubkey
 aes-gcm    = "0.11"     # AES-256-GCM for template-at-rest encryption
 argon2     = "0.5"      # Argon2id KDF for the recovery passphrase
+pbkdf2     = { version = "0.13", default-features = false, features = ["hmac"] }  # PBKDF2-HMAC-SHA512: the KDE wallet key ksecretd expects
 rand       = "0.10"      # CSPRNG for template keys, nonces, recovery salts
 libc       = "0.2"      # mlock/madvise (memlock), NSS getpwnam_r, SO_PEERCRED
 rustfft    = "6"        # 2D FFT for the RGB-only moiré/screen-replay PAD cue
--- a/crates/irlume-cli/src/commands.rs
+++ b/crates/irlume-cli/src/commands.rs
@@ -1328,6 +1328,7 @@ pub fn reseal(args: &[String]) -> ExitCode {
         return ExitCode::from(2);
     };
     let req = Request::SealPassword {
+        kind: None, // let the daemon judge from what the user has
         user,
         password: irlume_common::SecretBytes::new(pw.into_bytes()),
     };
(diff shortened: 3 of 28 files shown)


archledger:
## Review round applied

Four findings, all verified against the code before acting, all fixed in `708eeac`.

**Critical, local privilege escalation.** The helper bound the socket as root inside `/run/user/<uid>`, which the target user owns, then `chown()`ed it by pathname. `chown()` follows a final symlink while `unlink()` does not, so the user could rename the socket between `bind()` and `chown()` and leave a link to any root-owned file in its place. Same shape as KDE's CVE-2018-10380.

Fixed by removing the primitive rather than racing it: the parent drops to the target user before touching the runtime directory at all, so the socket is created with the right ownership and there is no second pathname resolution. The `chown` is gone, and the child's privilege drop with it. Verified: the socket comes out owned by the target user and the daemon runs as them.

**False success.** The parent treated a completed pipe write as proof the daemon had started. It never learned whether `execve` succeeded, and the caller exports `PAM_KWALLET5_LOGIN` on the strength of that exit status, which makes `pam_kwallet5` stand down. So a daemon that failed to start left the wallet locked and removed the fallback that would have opened it. A close-on-exec status pipe now separates exec (EOF) from pre-exec failure (one byte). Verified both directions: a non-executable target exits 1, a real one exits 0. Codified as a control in the check script.

**Nix.** The compiled-in helper path is an FHS path that does not exist on NixOS, and nothing set `IRLUME_KWALLET_INIT`, so a KDE-only NixOS user with a wallet-key envelope would have got no wallet and no fallback. The build now substitutes the store path and the install is unconditional. The parity check that passed anyway was only checking that the name appeared somewhere; it now checks the destination, verified by deleting the substitution and watching it fail.

**Race.** The secret and its kind were two separate reads of a file a concurrent re-arm can replace between them, which would tag one envelope's bytes with another's kind. The second read also collapsed "could not read the kind" into `LoginPassword` via `.ok().unwrap_or_default()`, letting an unreadable file authorise the wrong delivery. One load now yields both.

Also from the round: the stress check's descriptor assertion was measuring `/proc/self/fd`, the script's own shell, which the handoff never touches. It now counts the wallet daemon's descriptors and any sockets left behind.

## After the fixes

```
PASS: the derived key opened the wallet
PASS: a key from the wrong password was refused
PASS: a daemon that cannot exec is reported as a failure
stress: 10/10 opened, daemons 0 -> 0, peak daemon fds 9, sockets left 1
```

## CodeQL

The run reported 5 new critical alerts, all `rust/hard-coded-cryptographic-value`, and all inside `#[cfg(test)]` modules: fixed salts and passwords in the derivation tests. Test fixtures have to be deterministic to assert anything. Dismissed individually as "used in tests" with the file and line in each comment, rather than filtering the rule out of the config, so it stays active for non-test code.

## Still not covered

Not exercised through a real greeter login. The check drives the same code the login path drives, but it builds its own session rather than going through SDDM.

archledger:
Tested with two real greeter logins on the Zenbook (Fedora 44 KDE, SELinux enforcing), boots at 15:25 and 15:44 today, both authenticated by face. The chain held end to end both times. The 15:44 boot:

```
15:44:50 irlumed[1684]: UnsealPassword: OK for 'wisbfime' (score 0.7344), KDE wallet key unsealed
15:44:51 ksecretd[2409]: ksecretd5: Got pam-login param
15:44:51 ksecretd[2409]: ksecretd5: Waiting for hash on 4
15:44:51 ksecretd[2409]: ksecretd5: waitingForEnvironment on: 3
15:44:51 plasmalogin-helper[2402]: pam_kwallet5(plasmalogin:session): pam_kwallet5: we were already executed
15:44:54 pam_kwallet_init[2891]: socat[2891] W address is opened in read-write mode but only supports read-only
15:44:54 ksecretd[2409]: ksecretd5: client connected
15:44:54 ksecretd[2409]: Wallet opened by PAM
```

`journalctl -b -t irlume-kwallet-init` is empty (the helper logs only on error), `PAM_KWALLET5_LOGIN=/run/user/1000/kwallet5.socket` is in the session environment, and `ausearch -m avc` shows no denials for the helper on either boot. The wallet opened from the 56 sealed bytes with no re-key and no typed password.

A wallet prompt still appeared shortly after login on both boots. That is not this PR: the pam-spawned ksecretd registers org.kde.ksecretd and org.freedesktop.secrets, loses its bus connection within about 2 seconds of the greeter session closing, and xdg-desktop-portal's Secret probe then D-Bus-activates a bare ksecretd that takes both names with the wallet locked. The journal for the Aug 2 boot shows the same two-instance signature with stock pam_kwallet5 doing the spawning, so the race predates this branch. kwallet 6.28's KDBusService::Replace fix (KDE bug 509680) covers activation before the pam instance, not this ordering; tracking it separately.

Merging per the test plan: CI 15/15 on 708eeac, one review round applied, chain validated on hardware twice.

PR #260, opened 2026-08-04
pam: bound the read of the kwallet helper's stdout

### LINKED ISSUE #257: irlume-kwallet-init: the PAM module waits on the helper without a bound
`hand_key_to_wallet_daemon` in `crates/irlume-pam/src/lib.rs` calls
`child.wait_with_output()` with no ceiling. This runs in the PAM session
path, where the login blocks on it, so a helper that wedges (a stuck
`ksecretd` exec, a full pipe, a stopped process) hangs the login rather
than falling back to the password.

Found during the review of #256, where the same shape was fixed on the
GNOME path: that one now polls `try_wait` against a deadline and kills the
child when it expires. The KDE path is not the same two-line change,
because `wait_with_output` is also how the socket path is read back from
the helper's stdout, so bounding it needs a bounded READ, not just a
bounded wait.

Not urgent: no wedge has been observed, and every irlume PAM line is
optional or sufficient with a password fallback, so the failure mode is a
stalled login rather than a lockout.

archledger:
Closes #257.

`hand_key_to_wallet_daemon` called `wait_with_output()` with no ceiling, in the PAM session phase where the login blocks on it. A wedged helper (a stuck `ksecretd` exec, a stopped process) would hang the login instead of falling back to the password.

The GNOME path got a bounded wait in #256. This path needs a bounded read, not just a bounded wait: `wait_with_output` is also how the socket path comes back from the helper's stdout, and that read only returns when every holder of the pipe's write end has closed it. One holder is outside our control: the helper points the exec'd wallet daemon's stdio at /dev/null, but when that open fails the daemon inherits the pipe and outlives the login, so even a wait-then-read-to-EOF could block on a helper that exited cleanly.

## What changed

- `read_stdout_bounded`: the helper's stdout is set non-blocking and polled alongside `try_wait` against the existing 15s `HELPER_BUDGET`. On expiry the child is killed and reaped. Output is capped at 4096 bytes (the whole legitimate output is one socket path); past the cap the helper is killed rather than read further.
- After the child exits and the pipe is drained, the read returns with what is in hand instead of waiting for an EOF a leaked write end may never deliver.
- `wait_bounded` (the GNOME path) now delegates to a shared `reap_by`; its semantics are unchanged.

## Testing

- 5 new unit tests: normal output, nonzero exit, a wedged helper killed at the deadline, a leaked write end not holding the login after exit, and an output-spewing helper killed at the cap. All pass; full `cargo test -p irlume-pam` green.
- Each guard was proven to fire: three hand-applied mutants (deadline check removed, leaked-fd early break removed, output cap removed) each make exactly their targeted test fail, and the tree was restored byte-identical afterwards.

## Not tested

- A real KDE face login on hardware with this build. The wedge itself has never been observed in the wild (#257 says as much); the change is fail-safe by construction since every return of `false` lands on the password fallback.
- The /dev/null-redirect-failure leak path with the real `ksecretd` (simulated with a shell grandchild holding the pipe).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1021,6 +1021,7 @@ name = "irlume-pam"
 version = "0.8.1"
 dependencies = [
  "irlume-common",
+ "libc",
  "pamsm",
  "serde_json",
  "zeroize",
--- a/crates/irlume-pam/Cargo.toml
+++ b/crates/irlume-pam/Cargo.toml
@@ -19,3 +19,4 @@ irlume-common.workspace = true
 serde_json.workspace = true
 zeroize.workspace = true
 pamsm = { version = "0.5", features = ["libpam"] }
+libc.workspace = true   # O_NONBLOCK on the kwallet helper's stdout, so the read can carry a deadline
--- a/crates/irlume-pam/src/lib.rs
+++ b/crates/irlume-pam/src/lib.rs
@@ -525,30 +525,126 @@ fn hand_token_to_keyring_daemon(user: &str, token: &irlume_common::SecretBytes) 
     wait_bounded(&mut child, HELPER_BUDGET)
 }
 
-/// Ceiling on the unlock helper: its own socket deadlines are 10s, so this is
-/// that plus room to start and exit.
+/// Ceiling on the keyring helpers. The GNOME unlock helper's own socket
+/// deadlines are 10s, so this is that plus room to start and exit; the KDE
+/// helper forks and execs the wallet daemon and normally exits in
+/// milliseconds, so the same ceiling is generous there.
 const HELPER_BUDGET: Duration = Duration::from_secs(15);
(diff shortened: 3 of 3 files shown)


archledger:
Codex round complete: one merge-blocking finding, applied in 95bde56. The kill-then-wait cleanup was itself unbounded: SIGKILL does not end an uninterruptible kernel sleep, so a helper wedged in one would have hung the login right through the deadline. All four kill sites (including the pre-existing GNOME-path pair) now go through kill_bounded, which polls try_wait for at most 200ms after the kill and otherwise leaves the child for init. The review's added-test audit found all five tests have reachable failing observations; its remaining caveat, that no test places a child in an uninterruptible sleep, stands, since that needs privileged fault injection.

PR #261, opened 2026-08-04
cli: --fingerprint-only requires a surface that reaches pam_fprintd

### LINKED ISSUE #234: The --fingerprint-only gate accepts any installed PAM service, not one a surface reaches
Raised by the Codex round on #232, about code merged in #231.

## Mechanism

`pam_fprintd_wired` returns true when any enumerated service file carries an active `auth` rule naming `pam_fprintd.so`. On Arch and `Other` that boolean is what permits `irlume fingerprint enable --fingerprint-only`, which records `method=fingerprint` and stands face authentication down.

A service file being loadable does not establish that anything on the machine invokes that service. PAM service files define a configuration for the name an application asks for; installing one does not make it reachable from login, the lock screen, the console, or sudo.

So a package shipping, say, `/usr/lib/pam.d/example-fingerprint` with `auth sufficient pam_fprintd.so`, which no greeter or lock screen calls, satisfies the gate. Face goes down and every real prompt stays password-only.

## This predates #231

`live_pam_stacks` before #231:

```rust
entries
    .flatten()
    .filter(|e| !e.file_name().to_string_lossy().contains('.'))
    .filter_map(|e| std::fs::read_to_string(e.path()).ok())
```

Every dotless file in `/etc/pam.d`, with no requirement that a tracked surface reach it. So the gate has always accepted an unreferenced service; a stray `/etc/pam.d/anything` was enough.

What #231 changed is the size of that set. Vendor files are shipped by packages rather than chosen by an admin, so the odds of an unreferenced carrier being present are meaningfully higher than they were for `/etc/pam.d` alone. The measured Arch case is not an example of the problem (`kde-fingerprint` is a tracked surface and kscreenlocker does invoke it), but it shows the directory is full of package-owned services.

## The tension to resolve

The gate is broad on purpose. From its own doc comment: `pam_fprintd_wired` says "an active line exists somewhere", which is the right GATE, while `fprintd_coverage` is the right REPORT. The reasoning was that face must not stand down while nothing drives a prompt.

Tightening it to "a surface in `FP_SURFACES` reaches `pam_fprintd`" is the fail-closed direction and would close this. The cost is a false refusal: someone whose real login path is a custom service not in `FP_SURFACES` would be told nothing is wired when their finger genuinely works. Whether that trade is right is the decision this issue exists to make.

## Acceptance

- An unreferenced service carrying `pam_fprintd.so`, in either directory, does not by itself permit `--fingerprint-only`.
- A surface irlume tracks that reaches the module through an `include` still does.
- The negative control is a test: with the tracked surface removed and only the unreferenced carrier left, the gate refuses.

archledger:
Closes #234.

`pam_fprintd_wired` establishes that an active line exists somewhere, and a package can ship a service file nothing invokes, so the gate let `--fingerprint-only` stand face down while every real prompt stayed password-only. Recording `method=fingerprint` now additionally requires that some `FP_SURFACES` stack present on the machine reaches an active `pam_fprintd.so` line, through the same include-resolving walk the coverage table already uses.

## What changed

- New `fprintd_reaches_tracked_surface` (any ✓ row in the coverage walk) and `fingerprint_only_permitted`, the decision extracted as a testable value.
- One check at the choke point before `set_method`, so the authselect, pam-auth-update, and already-wired paths are all covered by it. The refusal names what was found and prints the coverage table.
- Coexist mode keeps the broad existence check: face stays active there, and a refusal would cost a custom-stack user their working finger for no safety gain.
- `pam_fprintd_wired` keeps its existence semantics for displays and the did-the-tool-write-anything verify; its doc now says so.

## Testing

- Six new tests covering the acceptance criteria: unreferenced carrier (machine and vendor dirs) refused, include-reached surface permitted, the negative control (remove the tracked surface, carrier alone flips the gate to refusal), the decision keyed on both inputs, and the measured Arch layout pinned verbatim.
- Three mutants (flag conjunct dropped, decision inverted, reach regressed to existence) each killed by exactly the matching tests; tree restored byte-identical.
- Real layouts on all three machines: the ThinkPad (Ubuntu, live fingerprint login) reads `gdm-fingerprint ✓`, so the tightened gate still permits the one machine whose finger genuinely works. archhost's Arch layout carries `pam_fprintd` only in vendor `kde-fingerprint` in the `-auth` tolerant form, which the walk recognizes (now a committed fixture). The Zenbook's Fedora layout resolves through authselect's `system-auth` symlink and include chain.

## Not tested

- A live `enable --fingerprint-only` run end to end (needs root plus an enrolled reader on a machine whose method I'm willing to flip; the decision it consults is the extracted function under test).
- Layouts outside these three distro families.

### DIFF
--- a/crates/irlume-cli/src/fingerprint.rs
+++ b/crates/irlume-cli/src/fingerprint.rs
@@ -387,6 +387,26 @@ fn enable(user: &str, args: &[String]) -> ExitCode {
         );
         return ExitCode::FAILURE;
     }
+    // Disabling face demands more than "a line exists": with --fingerprint-only
+    // the line must sit in a stack a tracked surface reaches (#234), or face
+    // stands down while every real prompt is password-only. One check here
+    // covers the authselect, pam-auth-update and already-wired paths alike.
+    // Coexist mode keeps face active, so it stays on the broad check: a
+    // custom-stack setup loses nothing there, and a dead finger line costs
+    // only a wrong "either factor" message rather than a lost biometric.
+    if !fingerprint_only_permitted(&PamSearchPath::live(), fingerprint_only) {
+        eprintln!(
+            "[fingerprint] an active pam_fprintd.so line exists, but no login surface irlume tracks reaches it"
+        );
+        eprintln!(
+            "[fingerprint] --fingerprint-only refused: it would disable face while no tracked prompt has a fingerprint path"
+        );
+        eprintln!(
+            "[fingerprint] method unchanged: wire pam_fprintd into a stack your login actually uses (see the table below), then re-run"
+        );
+        report_fprintd_coverage(&PamSearchPath::live());
+        return ExitCode::FAILURE;
+    }
     let method = if coexist {
         Method::Both
     } else {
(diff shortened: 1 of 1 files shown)


archledger:
Codex round complete: one merge-blocking finding, applied in 057d835. The walk read text, not control flow: 'auth requisite pam_deny.so' above a fingerprint line counted as reached although PAM returns before the module runs. Verified by measurement first: both of the review's regression cases failed against the submitted code, and pass with the tri-state walk. The modeled exit set is exactly requisite+pam_deny; sufficient+pam_permit is stateful and requisite+pam_nologin is conditional (it opens Arch's real vendor kde-fingerprint stack, which must keep its permit, and does; pinned in the fixtures). Unmodeled rules keep the permissive pre-#261 reading. Zenbook live-layout coverage verdicts are unchanged under the new walk.

Issue #174, opened 2026-07-29
Liveness gates are absolute thresholds on quantities that move with face distance, and face distance is measured then discarded


archledger:
Raised from the #167 measurements, which showed that ordinary user variation moves infrared brightness by roughly a factor of two on this hardware.

## irlume already measures distance and does not use it

`position_sample()` (`crates/irlume-auth/src/lib.rs:2611`) computes face width as a fraction of frame width, how far the face sits off centre, and mean face luma:

```rust
const MIN_FRAC: f32 = 0.12;   // face width / frame width
const MAX_FRAC: f32 = 0.55;
const CENTER_TOL: f32 = 0.18;
```

All three go into `PositionReport`, which exists to produce guidance strings during guided enrolment. Nothing downstream consumes them. `face_frac` spans a 4.6x range across its accepted band, and that band is wide by design, because it has to accept how people actually sit.

## Why that band is not harmless

Several liveness cues are absolute thresholds on quantities that plausibly depend on where the face sits:

```
IR_FACE_MIN_BRIGHTNESS   35.0   crates/irlume-liveness/src/lib.rs:146
MIN_CENTER_EDGE_RATIO     1.03  crates/irlume-liveness/src/lib.rs:157
GLINT_MIN               180.0   crates/irlume-liveness/src/lib.rs:181
RGB_FACE_MIN_BRIGHTNESS  60.0   crates/irlume-liveness/src/lib.rs:87
```

`IR_FACE_MIN_BRIGHTNESS` is the same shape of gate #167 removed: a fixed number on infrared brightness. Its rationale records the calibration source, "a lit live face ran ~83 mean overall on the Shinetech module".

The centre/edge ratio is the one most exposed to face size, because the geometry it measures changes with it. A face occupying 12% of the frame and one occupying 55% do not put the same physical part of the head at the "edge" of the sampled region, so the falloff being measured is not the same measurement at both distances. Its own documentation already calls it weak: a glossy infrared print cleared it in 69 of 70 trials (`docs/pad-results/2026-06-30-ir-liveness-selftest.md`).

## What was actually measured

Under #167, the same subject sitting still, varying only seating distance and glasses, over 2160 frames on the ASUS module. Whole-frame infrared means with the illuminator on:

```
close, no glasses      60.7 -  79.2
close, glasses        102.7 - 133.5
arm's length          114.5 - 124.8
arm's length, glasses 111.8 - 146.3
```

[...]

PR #262, opened 2026-08-04
doctor: report the negotiated streams against the Windows Hello minimums

### LINKED ISSUE #223: Nothing checks the IR stream against the Windows Hello minimums it was designed around
From a Microsoft Learn pass over the Hello camera requirements. This is a gap in what irlume TELLS the user, not a defect in what it does.

## The published minimums

Microsoft's HLK criteria for an IR camera module used for face authentication require the Hello-capable stream to be "MJPG/L8/NV12" with a resolution "greater than or equal to 320 x 320 pixels" ([camera driver bring-up guide](https://learn.microsoft.com/windows-hardware/drivers/stream/windows-hello-camera-driver-bring-up-guide)), and the UVC implementation guide states the profile requirement outright: "Windows Hello has a minimum requirement of 480x480@7.5fps for the RGB stream and 340x340@15fps for the IR stream" ([UVC camera implementation guide](https://learn.microsoft.com/windows-hardware/drivers/stream/uvc-camera-implementation-guide)). The same criteria require startup under 500 ms and verified timestamp precision.

## What irlume does today

It negotiates a format and streams. Nothing compares the negotiated IR resolution, frame rate or pixel format against those numbers, so a module that comes up at, say, 320x240@10fps is treated exactly like one at 640x480@30fps. The consequences land in places already under investigation: a smaller face means fewer pixels behind the centre/edge ratio (#174, #221), and a lower frame rate means fewer usable frames per burst for the strobe pairing.

The camera tier irlume reports ("secure", "convenience") answers a different question: whether an IR pin exists, not whether its stream meets the specification the whole cue set was calibrated against.

## Suggested shape

A `doctor` line, not a gate. Refusing to authenticate on an under-spec stream would be a regression for anyone whose camera works fine; saying so is what the user can act on:

```
IR stream        640x480@30fps L8  (Windows Hello minimum: 340x340@15fps)
```

and a warning only when below it. The values are already known at open time. Worth pairing with the same reading for RGB (480x480@7.5fps minimum).

This also gives the hardware-report issue template something concrete to collect, which is useful for the modules irlume has never run on.

archledger:
Closes #223.

Nothing compared what the cameras negotiate against the floors Microsoft publishes for Hello modules ("Windows Hello has a minimum requirement of 480x480@7.5fps for the RGB stream and 340x340@15fps for the IR stream", UVC camera implementation guide on learn.microsoft.com), so a module at 320x240@10fps read exactly like one at 640x480@30fps. The consequences land where #174/#221 are already looking: fewer face pixels behind the centre/edge ratio, fewer usable burst frames.

## What changed

- `StreamSpec` and `negotiated_stream(device, role)` in irlume-camera: doctor opens each selected node through the same `RgbCamera::open` / `IrCamera::open` negotiation capture uses, no streaming, no buffers, so the report cannot drift from what capture does. The rate comes from the driver's VIDIOC_G_PARM interval; irlume never sets one.
- A doctor line per present spectrum, warning only below the floor, per the issue: a gate would break cameras that work fine. `Some(false)` = below; `None` (dims meet, rate unreported) reads as cannot-say, never as a pass. An unnegotiable node (mid-capture daemon, pin mismatch) reads Unknown, never silence.
- New machine-report check ids `ir-stream-hello-minimum` / `rgb-stream-hello-minimum`.
- The backlight-compensation write moved from `RgbCamera::open` to `session()`: doctor must not mutate camera state, and every capture path goes through `session()`, so capture behaviour is unchanged (the write is idempotent and precedes the auto-exposure warm-up either way).

## Testing

- `hello_minimum_verdicts_cover_all_three_outcomes` pins the decision: at-floor passes, one short dimension fails regardless of rate, sub-floor rate fails, unreported rate with passing dims is indeterminate, unreported rate with failing dims is still a definite fail.
- Measured on all three machines with this build: Zenbook ASUS module IR 640x400@15fps GREY ✓ (exactly at the rate floor) and RGB 640x480@30fps YUYV ✓; archhost NexiGo N930W IR 640x360@30fps GREY ✓ (20 rows above the floor) and RGB 640x480@30fps ✓; ThinkPad RGB 640x480@30fps ✓ with no IR line printed, since it has no IR node.
- Full workspace clippy clean; irlume-camera, irlume-cli, irlume-pam suites green.

## Not tested

- A genuinely under-spec module (none on hand); the Warn branch is exercised only by the unit test.
- MJPEG-only or non-UVC nodes; `negotiated_stream` reuses the existing negotiation, which already refuses those with its existing message.

### DIFF
--- a/crates/irlume-camera/src/lib.rs
+++ b/crates/irlume-camera/src/lib.rs
@@ -1315,30 +1315,104 @@ impl StreamSpec {
 }
 
 /// The driver's reported capture rate for `dev`, from VIDIOC_G_PARM. The
-/// interval is time-per-frame, so the rate is its inverse; a zero on either
-/// side of the fraction means the driver reports nothing usable.
+/// interval is time-per-frame, so the rate is its inverse.
 fn driver_fps(dev: &Device) -> Option<f64> {
-    let p = Capture::params(dev).ok()?;
-    let (num, den) = (p.interval.numerator, p.interval.denominator);
-    if num == 0 || den == 0 {
-        return None;
-    }
-    Some(f64::from(den) / f64::from(num))
+    fps_from_params(Capture::params(dev).ok()?)
 }
 
-/// What the capture path would negotiate on `device`, without streaming: the
-/// same verify + privacy + format walk as [`RgbCamera::open`] /
-/// [`IrCamera::open`], no buffers allocated, capture LED off. Reusing the real
-/// open is the point: a parallel reimplementation here could drift from what
-/// capture actually does, and this feeds a report people act on (#223).
-pub fn negotiated_stream(device: &str, role: Role) -> irlume_common::Result<StreamSpec> {
-    match role {
-        Role::Rgb => Ok(RgbCamera::open(device)?.spec()),
-        Role::Ir => Ok(IrCamera::open(device)?.spec()),
-        _ => Err(Error::Hardware(format!(
-            "{device}: no stream to negotiate for this role"
-        ))),
+/// [`driver_fps`]'s conversion, split out so the capability rule is testable
+/// without hardware. V4L2 specifies that `timeperframe` is meaningful only
+/// when the driver sets `V4L2_CAP_TIMEPERFRAME`; without the flag the fields
+/// are whatever the driver left in the struct, and reading a plausible-looking
+/// `1/30` there would let an unestablished rate publish as a pass. A zero on
+/// either side of the fraction is equally unusable.
+fn fps_from_params(p: v4l::video::capture::Parameters) -> Option<f64> {
+    if !p
+        .capabilities
+        .contains(v4l::parameters::Capabilities::TIME_PER_FRAME)
+    {
+        return None;
     }
+    let (num, den) = (p.interval.numerator, p.interval.denominator);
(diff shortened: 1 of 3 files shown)


archledger:
Codex round complete: all four findings applied in a3c9346. (1) driver_fps now requires V4L2_CAP_TIMEPERFRAME before trusting timeperframe, with a hardware-free test on the extracted fps_from_params. (2) Both check ids are always emitted, Info when no node is selected, and are added to the MACHINE-API registry; a regression test drives the extracted stream_minimum_checks with nonexistent paths and asserts both ids appear exactly once. (3) The probe no longer calls the capture constructors: negotiated_stream shares the candidate walks (negotiate_rgb_format, negotiate_ir_format_via) but applies them through VIDIOC_TRY_FMT via the crate's low-level ioctl, so doctor is genuinely read-only; measured by parking /dev/video0 at 320x240, running doctor, and reading 320x240 back while the line reported the 640x480 capture would negotiate. (4) Rates print with two decimals. Re-measured on the Zenbook and archhost with the TRY_FMT build: same verdicts, same dimensions, rates now shown as 15.00/30.00.

Issue #192, opened 2026-08-01
BRIO (046d:085e): camera-tune completes no rounds; IR dark with the face-auth control already at the applied value


archledger:
Observed 2026-07-31 on archhost (Arch, kernel USB layout: BRIO at /dev/video0-3, RGB video0 MJPG/NV12/YUYV, IR video2 GREY) while extending the #188 hardware validation to a third camera model.

## What happens

- irlumed logs \`"Logitech BRIO": IR is dark (mean 16-24) with no active emitter\` and \`irlume camera-tune --rounds 1\` reports \`capture-mode probe: no round completed on this camera pair\`.
- The daemon issues zero emitter \`SET_CUR\`s during the whole capture (confirmed with IRLUME_LOG_EMITTER_WRITES=1).
- Reproduced identically on main at 4db2d53 in a separate worktree: same probe failure, same zero writes, same dark-IR log line. Not introduced by #191.

## Why no emitter write happens

- The BRIO publishes Microsoft camera control on unit 12, selectors 0x06 and 0x09 (no IR Torch 0x0a). Derived from ir-setup's dry run, not assumed.
- Its GET_DEF for selector 0x06 is 010202000000000000, and ir-setup refuses with "selector 0x06 is already set to the value setup would apply": the power-on default already carries the mode irlume would request, so write_if_different has nothing to write and the #188 stream record is inert on this camera by design.
- There is no device-reported payload that differs from the wanted one, so the camera cannot be staged into a must-write state without inventing bytes (the #159 class; not done).

## What this issue is actually about

The mode bits are set, a GREY stream is up, and the IR image stays at mean 16-24. Candidate causes, none verified (the #185 list applies):

- the BRIO gates its emitter on something beyond the D-mode bits on selector 0x06 (frame rate, resolution, or the streaming interface named in the payload; its payload names interface 2 where the ASUS/NexiGo name 3);
- the emitter fires only in short strobes the probe's sampling misses;
- a physical cover or privacy shade on the unit (not checked; nobody was at the machine);
- the room is genuinely dark and the lift threshold logic never ran because no write ever happened.

## What was tested

xu_set get/def against unit 12 works (park wrote its own GET_DEF and read it back). ir-setup's consent gate and refusal path work. The #188 harness correctly fails the run on this camera since c9d4237; before that fix it would have reported the sections as green or skipped.

Related: #185 (a dark IR frame has several causes, distinguish them), #168-#170 (Windows camera contract).

archledger:
[...]

PR #263, opened 2026-08-04
camera-tune: a concurrent arm that only errors is a Sequential verdict

archledger:
Fixes the probe half of #192.

`measure_contention` required both arms to complete a round, so a camera whose concurrent captures always error reported "no round completed" and threw away the strongest evidence the probe can produce. Measured today on the BRIO (046d:085e) on archhost: sequential rounds measure fine (IR lit 58-62 by the camera's own strobe), while concurrent rounds fail with EINVAL on the RGB open while the IR sibling streams — the camera cannot stream both nodes at once, which is precisely the answer camera-tune exists to store.

## What changed

- `PairSample.failed` counts errored rounds.
- Only a dead sequential arm fails the probe now, with a message naming the failure counts. A concurrent arm with zero completed rounds returns a report.
- `concurrent_impossible()` decides Sequential explicitly (not via what division does with empty samples) and makes the verdict conclusive in any light: an errored open is not a brightness reading, so the dark-room caveat does not apply.
- The daemon renders the case as "cannot stream RGB and IR at once (all N concurrent attempts errored)" instead of "retained 0% of brightness", which read as dimming when nothing streamed.

## Testing

- New unit test pins the measured BRIO shape (dark room, zero concurrent rounds, all failed) to Sequential + conclusive, and distinguishes it from the nothing-attempted shape.
- Hardware: ran the fixed probe against the real BRIO pair on archhost. Before: "capture-mode probe: no round completed". After: sequential 3/3, concurrent 1 completed + 2 errored, verdict Sequential, conclusive — the mixed case lands correctly through the retention arithmetic when a round does sneak through.
- Existing contention tests updated for the new field; all 247 camera tests green.

## Not tested

- The daemon's message rendering on hardware (the daemon on archhost is inactive; the report fields it formats are the ones the hardware run produced).
- The remaining #192 question — the "IR is dark ... no active emitter" line on a camera whose emitter demonstrably strobes — is a separate defect in the dark-frame diagnosis (no inter-frame evidence) and gets its own issue.

### DIFF
--- a/crates/irlume-camera/examples/capture_bench.rs
+++ b/crates/irlume-camera/examples/capture_bench.rs
@@ -113,6 +113,18 @@ fn main() {
     // camera-tune` would make from the same measurement, run through the shared
     // implementation so the tool and the example cannot drift apart.
     match irlume_camera::measure_contention(&rgb_dev2, &ir_dev, rounds.min(6)) {
+        Ok(report) if report.concurrent_impossible() => {
+            // Same rule as the daemon: an arm that never streamed has no
+            // retention to print, and "keeps 0%" would read as dimming.
+            println!(
+                "\nRecommendation: {} capture for this camera\n  \
+             it cannot stream RGB and IR at once (all {} concurrent attempts \
+             errored; {} sequential round(s) completed)",
+                report.recommended_mode().as_str(),
+                report.concurrent.failed,
+                report.sequential.rounds,
+            );
+        }
         Ok(report) => {
             println!(
                 "\nRecommendation: {} capture for this camera\n  \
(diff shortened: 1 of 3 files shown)


archledger:
Codex round complete: all three findings applied in 24b462f + efe339b. (1) A panicking capture worker now aborts the probe instead of counting as a failed round, so a software defect can no longer persist as a camera-capability verdict; the captures are injected through measure_contention_impl so the case has a hardware-free test. (2) An all-error concurrent arm now requires a trailing one-at-a-time control to succeed before the verdict stands, with tests for both the passing and the camera-died shapes; the daemon and capture_bench print observed counts. (3) capture_bench handles concurrent_impossible instead of rendering 0% retention. Re-ran on the real BRIO after the changes: sequential 3/3, concurrent 0 completed with 3 errored, trailing control passed, verdict Sequential and conclusive — the full new path exercised on the hardware that motivated it.

PR #265, opened 2026-08-04
auth: evaluate the completed consent take, not just the in-loop cadence

archledger:
Part of #101 (the boundary defect found in today's measured session; the issue's larger residual, consent-without-intent, stays open).

consent_watch checks the accumulating pose series every CHECK_EVERY=6 poses, so a take whose length is not a multiple of six ends with unevaluated trailing poses. A gesture completing in exactly those frames is refused while the whole-take evidence passes every gate the log line prints. Measured on hardware today: two 20-pose windows read pitch_range 0.077-0.085 against the 0.075 floor after their last in-loop check ran at pose 18, and one cost a real trial its release (6 of 8 continuous-nod trials; the refused trial 8's post-match window is exactly this shape).

## What changed

One evaluation of the complete pose/EAR series after the stream ends, short-circuited when an in-loop check already accepted. Zero added capture cost. A journal line marks when the final evaluation is the one that decided, so a replay can prove this path fired rather than inferring it from a released trial.

## Trade stated plainly

The final check also evaluates still takes, so any take whose full-series evidence clears the gates now fires where the cadence previously masked it. That is the detector judging the take instead of a scheduling artifact, and today's still windows sat at pitch_range 0.012-0.029 against the 0.075 floor, nowhere near it. The gate's discrimination properties are unchanged; only the evaluation coverage is.

## Testing

- The decision itself (detect_nod / detect_deliberate_closure over a series) is unchanged and keeps its existing tests; irlume-auth suite green, clippy clean.
- The new path cannot be unit-driven without injecting the camera stream into consent_watch; validation is a live replay on the Zenbook: nod trials against the dev daemon, checking the journal for the completed-take marker and that still trials stay at zero accepts. Running next, results will be posted on the PR.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -271,6 +271,33 @@ fn consent_gesture_enabled() -> bool {
     !irlume_common::config::read_kv("settings.conf", "polkit_gesture").is_some_and(|v| falsy(&v))
 }
 
+/// The consent verdict once the watch's stream has ended: what the in-loop
+/// cadence already found, or a gesture visible only on the COMPLETE take.
+///
+/// The in-loop check runs every `CHECK_EVERY` poses, so a take whose length is
+/// not a multiple of it ends with unevaluated trailing poses, and a gesture
+/// completing in exactly those frames was refused with whole-take evidence
+/// that passes every gate (measured 2026-08-04, #101: two 20-pose windows at
+/// pitch_range 0.077-0.085 against the 0.075 floor, last in-loop check at
+/// pose 18; one cost a real trial its release). A pure function rather than
+/// inline in `consent_watch`, because this boolean directly satisfies the
+/// consent gate for credential release and a test must be able to fail if the
+/// completed-take evaluation is removed.
+fn completed_consent_take_hit(
+    hit_in_loop: bool,
+    allow_nod: bool,
+    poses: &[irlume_liveness::PoseSample],
+    ears: &[irlume_liveness::EarSample],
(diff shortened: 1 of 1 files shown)


archledger:
Codex round complete: the finding is applied in 3c57af6. The post-stream decision is extracted as completed_consent_take_hit, a pure function, with three regression tests: the measured boundary shape (18 flat poses plus the nod completing in the trailing two, with the premise asserted — the 18-pose prefix must not read as a nod), that the final evaluation widens take coverage but never the accepted gesture set, and that a still series stays quiet through the new path. A hand-applied mutant removing the completed-take evaluation is killed by exactly the trailing-frames test (verified by a self-reversing run; tree restored from the pre-mutant commit). The remaining seam the tests cannot see is consent_watch calling the helper, one grep-visible line, same shape as the #261 choke point. Live replay on hardware still to come, after the current capture session frees the camera.

archledger:
Live replay done on the Zenbook with the branch daemon (byte-verified install): 8 continuous-nod trials released 7 (morning session on the old daemon: 6 of 8), 5 still trials released 0. The completed-take accept path did not trigger naturally this time; every accepted take ended at a pose count that is a multiple of CHECK_EVERY, so all releases were in-loop, and the one refused nod trial's windows read pitch_range 0.053-0.068, genuinely under the floor rather than a boundary case. The path's SAFETY side did run on hardware: every refused 40- and 80-pose window (neither a multiple of 6) had its completed-take evaluation execute, 17 windows, zero false accepts. The accept side's proof is the unit test pinning the morning session's measured boundary shape and the mutant it kills.

Issue #173, opened 2026-07-29
Closure calibration is one snapshot per user, but glasses change the eye it measured


archledger:
Found while measuring something else: infrared frame classification for #167, with the same subject sitting still, wearing glasses and not wearing them.

## What is there now

`irlume calibrate-closure` records `ear_open` and `ear_closed` from the user's own eyes and stores them against their username (`Request::SetClosureCalibration { user, ear_open, ear_closed }`, `crates/irlume-cli/src/main.rs:727`). `ClosureCalibration::closed_threshold()` then places the gate a deep fraction of the way from that user's closed extreme toward their open one, rather than at a fixed value. That is the right design and it is already per-user, so this is not a request to add calibration.

The gap is that it is **one** snapshot per user. `calibrate-closure` refuses to overwrite an existing calibration without a terminal, so the first capture is the one that sticks.

The code already half-says this. From `main.rs:731`, printed immediately after storing:

> What was just stored describes the eye shape under the light in the room right now.

Glasses are a second thing it describes, and nothing records which state the user was in.

## Why glasses specifically

A lens sits between the camera and the eye. It refracts the eyelid boundary the landmarker keys on, and on an infrared stream it adds a specular return from the 850nm illuminator across part of the eye region. A user who calibrates bare-eyed and then authenticates wearing glasses is being measured against extremes taken from a different face, and the failure is asymmetric: a threshold derived from a wider open-EAR makes a genuine closure fail to register, so the consent gesture silently does not fire and the credential is not released.

## Measured, though not on EAR

While testing #167 the same subject sat still, close to the camera, and the only thing that changed between two runs was putting glasses on. Whole-frame infrared means:

```
glasses off   emitter on  68.0- 79.2    emitter off  15.7- 22.8
glasses on    emitter on 104.1-122.4    emitter off  35.6- 56.0
```

That is not an EAR measurement and should not be read as one. It establishes the weaker but relevant claim that glasses materially change what this camera returns for the same person in the same position, enough to move every emitter-off frame across a fixed threshold. Whether the EAR distribution shifts by a comparable amount has not been measured here, and should be before anything is tuned.

## Suggested shape

Store more than one calibration per user and pick between them, rather than detecting glasses as a category:

- let a user hold several named calibrations and select at capture time by which one the live EAR distribution matches
- or capture both states during `calibrate-closure`, prompting for a second pass, and keep whichever the current frames sit closer to
[...]

PR #267, opened 2026-08-04
cli: calibrate-closure --measure-only, EAR readings without storing

archledger:
Part of #173 (the measurement mode; the multiple-calibrations design stays open).

The #173 question needed EAR distributions from several states (glasses on and off, eyes open and closed) compared against one stored calibration, and the only path that measured EAR also overwrote that calibration. `--measure-only` captures N `CaptureEarMedian` rounds through the daemon, prints each reading and the median under an operator-supplied `--pose` label, and stores nothing. The median goes through `median_ear`, the same even-sample-averaging median the calibration path stores, and `measure_pose_label` refuses a flag-shaped or missing label rather than titling research data '--rounds'.

## Testing

- Produced the measurement posted on #173: four states, five rounds each, Zenbook ASUS FHD IR module, 2026-08-04.
- One instrument fact from that session changes how the tool should be used: the operator has to look at the camera, not the terminal. Reading output mid-capture pointed the gaze down and read open eyes at EAR 0.07-0.16; the same eyes looking at the camera read 0.246-0.254. A self-paced run in the operator's own terminal avoids this; a remotely driven one invites it.
- `measure_pose_label` is pinned by a test over its four argument shapes (absent flag, proper label, `--pose --rounds`, trailing `--pose`). The median arithmetic keeps the existing `median_ear` test.

## Not tested

- Runs against an unreachable daemon: each round prints its own error and the summary counts what succeeded, but every real run had a live daemon behind it.

### DIFF
--- a/crates/irlume-cli/src/commands.rs
+++ b/crates/irlume-cli/src/commands.rs
@@ -1569,6 +1569,10 @@ ENROLLMENT & AUTH
   calibrate-closure [--rounds N] [--force]   teach the eye-closure gesture for app
                         prompts; captures N rounds (default 3) and stores the median
                         (sudo; the head nod is the default and needs no calibration)
+  calibrate-closure --measure-only [--rounds N] [--pose LABEL]
+                        print labelled EAR readings and their median WITHOUT
+                        replacing the stored calibration (sudo; for comparing
+                        states such as glasses on/off before re-calibrating)
 
 KEYRING / TPM
   keyring <arm|status|forget>     TPM-sealed secret so a login opens your wallet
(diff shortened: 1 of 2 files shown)


archledger:
Codex round complete plus the CI failure, all applied in 2ec8d62 and fc971f9. The two failed checks were one clippy deny (useless str conversion), fixed first. From the round: the summary median now goes through the existing median_ear (even samples average their middles; the inline upper-middle pick disagreed with the repo's own tested definition once a failed round made the count even  today's posted #173 data is unaffected, all its runs completed 5 of 5), the pose label is parsed by an extracted measure_pose_label that refuses swallowed or missing values (tested over all four shapes: absent flag, proper label, --pose --rounds, trailing --pose), both help surfaces document the mode, and the timing message now matches the loop.

PR #268, opened 2026-08-04
ir_dark: a lit frame anywhere in the burst refutes the dead-emitter advice

### LINKED ISSUE #264: A strobing emitter reads as 'IR is dark with no active emitter', and the advice points at firmware
Found while closing the probe half of #192 on the BRIO (046d:085e), which strobes its emitter from power-on without any irlume write.

## What happens

During a sequential capture on archhost today, irlumed printed:

```
[ir] "Logitech BRIO": IR is dark (mean 34) with no active emitter; run `sudo irlume ir-setup` ...
```

while the raw burst on the same camera, seconds apart, measured a hard strobe: lit frames at 54-128 interleaved with ambient frames at 0.6-1.7 (`ir_strobe_probe`, 36 frames, dark room). The emitter is firing; the whole-burst view just averages the phases, and a chosen frame can land near the interleave mean.

## Why the diagnosis misses it

`ir_dark::DarkEvidence` carries the CHOSEN frame's mean and stddev, plus the camera's illumination metadata counts (which the BRIO does not provide: 0/30 frames classified). It has no inter-frame evidence, so a burst whose brightest frame measured 128 and whose chosen frame measured 34 is indistinguishable from a burst that was uniformly dark. The rendered advice sends the user to `ir-setup`, i.e. at camera firmware, for an emitter that works — the exact #185 class this module exists to prevent.

irlume applied no emitter control on this camera (its GET_DEF already carries the wanted mode, #192), so `emitter_active` is false, and the mean landed under `DARK_MEAN_MAX` — both arms of the "no active emitter" diagnosis satisfied by a working camera.

## Suggested shape

Add the burst's brightest-frame mean (or `lit_mean` from `IrCaptureStats`, which already exists for exactly this) to `DarkEvidence`. A dark chosen frame beside a bright burst maximum is a frame-selection or strobe-phase question, not a dead emitter, and the message should say so instead of recommending a firmware write.

Also worth checking while in there: why frame selection (brightest lit under the clipping limit, #221) landed on a 34-mean frame when the burst plausibly carried 54+ frames — the BRIO's strobe warms up over ~4 frames and has two regimes (alternating 0.6/128 for ~10 frames, then 1.7/54 steady), so a short early burst may genuinely contain nothing lit.

Measurements: this session's probe output on archhost (BRIO at /dev/video4 RGB, /dev/video6 IR, native IR 340x340 GREY).

archledger:
Closes #264, narrowed by its review round to the arm the evidence supports.

The dark-frame diagnosis had no inter-frame evidence. The first version keyed a new arm on the burst's overall maximum; the round showed that was wrong twice. With no metadata the #221 clip demotion never engages, so selection always takes the maximum and a dark choice beside a brighter unclassified frame cannot reach the diagnosis band (the invariant now has its own test in ir_metadata). With metadata present, a bright frame the camera flagged DARK is what ambient light looks like, and concluding the emitter works from it could suppress a genuine emitter failure.

## What changed

- `DarkEvidence.lit_max_mean`: the brightest frame the camera's own metadata flagged LIT, 0.0 when it classified none.
- A new arm, `LitFrameDemoted`: when a metadata-lit frame is bright while the chosen frame is dark, metadata and optics agree the emitter fired and selection demoted the lit frame, which the #221 clip-aware pick does when it is blown. It outranks `LitButDark`, whose failed-LED and covered-lens possibilities that agreement refutes, and stays behind the shutter and `IRLUME_IR_EMITTER=off` arms. The message names both numbers and points at nothing.
- The reviewer's conflict scenario (means 90/34/20, flags Dark/Lit/Lit) is pinned to `LitButDark`: the lit maximum there is the dim 34, the arm stays quiet, and the honest uncertainty stands.

## What this means for the original BRIO report

The BRIO classifies nothing (0 of 30 frames in every measured run), so this arm cannot fire there, and the selector invariant says its mixed-burst shape never existed: its 'IR is dark (mean 34) with no active emitter' lines came from bursts dark end to end, the pre-strobe warm-up window. That residual needs burst timing, not diagnosis evidence; noted on #264.

## Testing

- Tests pin the demoted shape, the conflict scenario, the shutter and off-policy precedence, the no-metadata quiet case, and the selector invariant (no metadata selects the maximum, clipping input or not). Full camera suite green (252), clippy clean.
- Hardware: four capture rounds against the real BRIO produced healthy lit frames (78-85) and correct silence from every diagnosis arm.

## Not tested

- The demoted arm end to end on hardware: it needs a camera that classifies frames AND blows its lit maximum in the diagnosis band; neither the NexiGo nor the ASUS entered that state today.

### DIFF
--- a/crates/irlume-camera/src/ir_dark.rs
+++ b/crates/irlume-camera/src/ir_dark.rs
@@ -58,6 +58,14 @@ pub struct DarkEvidence {
     pub frame_mean: f64,
     /// Pixel standard deviation of the chosen frame.
     pub frame_stddev: f64,
+    /// Mean of the BRIGHTEST frame in the burst. Inter-frame evidence the
+    /// diagnosis lacked (#264): a strobing emitter interleaves lit and
+    /// ambient frames, so a dark CHOSEN frame beside a bright burst maximum
+    /// means the emitter fired and selection (or clipping demotion, #221)
+    /// landed elsewhere; the BRIO measured lit 54-128 against ambient 0.6-1.7
+    /// with no irlume write at all, while the old evidence read its ambient
+    /// phase as a dead emitter and recommended a firmware write.
+    pub burst_max_mean: f64,
 }
 
 /// The named cause of a dark IR burst, most specific evidence first.
@@ -101,6 +109,12 @@ pub enum IrDarkCause {
     /// destroyed a reporter's camera (#159). `ir-setup` does the same job
     /// from the camera's own descriptor and its own `GET_DEF` values.
     NoEmitterDriven,
+    /// The chosen frame is dark but the burst holds a bright frame: light
+    /// reached the sensor this capture, so the emitter works (a strobing
+    /// module interleaves lit and ambient phases) and the question is frame
+    /// SELECTION, never the firmware. Advising `ir-setup` here is the #185
+    /// wrong-turn this module exists to end (#264).
+    LitBurstDarkChoice { burst_max_mean: f64 },
     /// The control is active, the camera supplies no metadata, and the frame
     /// carries real signal: the evidence cannot separate an emitter the
     /// camera ignored from a subject out of range or a genuinely dark scene.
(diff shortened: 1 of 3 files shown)


archledger:
Codex round complete, both halves applied in 657ff7f: the selector invariant refutation was verified against best_gate_frame (the clip demotion is gated on any_lit, so the no-metadata mixed state really is unreachable; now pinned as a test in ir_metadata), and the arm was narrowed to the metadata-confirmed maximum with the review's conflict scenario pinned to LitButDark. The PR body is rewritten to the final semantics, including what this means for the original BRIO report: its dark lines were pre-strobe warm-up bursts, which no diagnosis evidence can fix, noted on #264.

PR #269, opened 2026-08-04
vision: settle onnxruntime resolvability before ort can park on it

### LINKED ISSUE #266: A missing onnxruntime dylib hangs model-loading commands forever with no output
Found 2026-08-04 while running `padcapture` under sudo: the command produced nothing and sat indefinitely. strace shows the loader walking the standard paths for `libonnxruntime.so`, six ENOENTs, then parking in `futex(..., FUTEX_WAIT_BITSET_PRIVATE, ...)` forever. No error, no message, no exit.

Trigger: any environment where `ORT_DYLIB_PATH` is unset and the runtime is not on the loader path. sudo is the easy way to get there (it scrubs the user's environment), but a plain shell without the variable does the same. The packaged daemon carries its own configuration, which is why normal operation never sees this; the CLI's model-loading commands (`padcapture`, likely `identify`/`enroll` on a source install) reach `Detector::load_from_file` and block before their first print.

Expected: the immediate, actionable error the situation deserves, e.g. "libonnxruntime.so not found (looked in the loader path; set ORT_DYLIB_PATH); irlume installs it at /usr/share/irlume/onnxruntime/lib/". `doctor` already knows the answer (`ort-dylib-path` check); the failing command should say it at the point of failure.

Worth checking whether the futex wait is ort's dylib-load path or something in irlume's own lazy init, and whether a pre-flight existence check before handing control to ort is the cheaper fix.

archledger:
Closes #266.

ort's load-dynamic failure mode for an unresolvable library is not an error: the loader walks its paths and the process parks forever in a futex with no output (straced on 2026-08-04; sudo scrubbing ORT_DYLIB_PATH is the easy trigger, and the CLI's model commands hit it where the daemon's unit file does not).

## What changed

`ensure_ort_resolvable` runs at the single session-construction choke point and settles resolvability three ways, in order:

1. An explicit `ORT_DYLIB_PATH` is verified to exist; a wrong value gets the actionable message, not the park.
2. With none set, the packaged copy at /usr/share/irlume/onnxruntime/lib is used when present, which is what lets a bare CLI invocation work at all.
3. Failing both, a `dlopen` probe asks the real loader, so a distro-managed copy anywhere the loader looks (hwcaps subdirectories included) still passes without irlume maintaining a path list.

Only when all three fail does it error, naming both fixes. The verdict is cached in a `OnceLock`, so the probe costs one stat or one dlopen per process.

## Testing

- The measured before/after on the machine that found it: the bare `sudo ... padcapture` invocation went from a silent hang (over 120 seconds before it was killed, zero output) to a working capture via the packaged-copy arm, and a deliberately wrong `ORT_DYLIB_PATH` errors in 64 milliseconds with the fix in the message.
- irlume-vision suite green, clippy clean.
- No unit tests for the resolver itself: it reads process-global state (an environment variable it may also set, plus a `OnceLock` verdict), so in-process tests would race the crate's other tests and could pin only one arm per process. The behavioural A/B above covers the two arms a test could not.

## Not tested

- The dlopen arm on a machine whose onnxruntime lives only on the loader path (none of the three machines here has one); its failure mode is the old behaviour, reached only when the first two arms already found nothing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1033,6 +1033,7 @@ version = "0.8.1"
 dependencies = [
  "irlume-camera",
  "irlume-common",
+ "libc",
  "ort",
  "rustfft",
  "thiserror 2.0.19",
--- a/crates/irlume-vision/Cargo.toml
+++ b/crates/irlume-vision/Cargo.toml
@@ -18,7 +18,7 @@ coreml = ["ort/coreml"]
 tensorrt = ["ort/tensorrt"]
 
 [dependencies]
-libc.workspace = true   # dlopen probe: is libonnxruntime resolvable, BEFORE ort parks on a missing one (#266)
+libc.workspace = true   # dlopen/dlsym probe: prove a library is loadable ort BEFORE ort can park on it (#266)
 irlume-common.workspace = true
 irlume-camera.workspace = true
 thiserror.workspace = true
--- a/crates/irlume-vision/Cargo.toml
+++ b/crates/irlume-vision/Cargo.toml
@@ -18,6 +18,7 @@ coreml = ["ort/coreml"]
 tensorrt = ["ort/tensorrt"]
 
 [dependencies]
+libc.workspace = true   # dlopen probe: is libonnxruntime resolvable, BEFORE ort parks on a missing one (#266)
 irlume-common.workspace = true
 irlume-camera.workspace = true
 thiserror.workspace = true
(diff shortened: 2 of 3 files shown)


archledger:
Codex round complete: all four findings applied in 0f1fcd6, with the second one's replacement corrected by measurement. Applied as proposed: no environment mutation (finding 1), the Debian/Ubuntu /opt candidate (finding 3), the extracted configured_ort with the four tests plus the empty-variable-is-unset semantic (finding 4). Finding 2's mechanism was right but its replacement was not: delegating to ort::init_from assumes ort returns load errors, and straced with ORT_DYLIB_PATH=/etc/hostname it parks in the same futex as the lazy path (the file opens, dlopen rejects it, FUTEX_WAIT forever). So the probe stays ours, upgraded to answer the finding: dlopen plus a dlsym check for OrtGetApiBase, reported in the loader's own words, and only an accepted name reaches init_from, which retains the handle and applies the api-24 floor. Measured end to end: bare sudo works via the packaged copy, a missing path errors with the fix in the message, and the not-a-library case that parked both earlier designs errors in 61 milliseconds ('file too short'). The version-floor failure class remains untested, no pre-1.24 library being on hand.

PR #270, opened 2026-08-04
auth: landmark_replay, the offline twin of landmark_dump, plus the #25 relief measurement

archledger:
Measurement and tooling for #25. **No gate code**; the results doc states what would have to be measured before any threshold is written.

## The tool

`landmark_replay` reads stored PGM frames and writes the same `frameNN.landmarks.csv` and `index.txt` that the live `landmark_dump` produces, with identical 3x3 patch sampling. The relief work needs one instrument applied to corpora captured in different sessions; re-capturing in order to change an analysis is how a measurement ends up describing the session rather than the cue.

Instrument check before any analysis used it: replaying a 2026-08-02 burst reproduced that session's CSVs **byte-identically, 18 of 18**, with the same detection count. Re-verified after rustfmt touched the file.

## What it measured

Full numbers in `docs/pad-results/2026-08-04-landmark-relief.md`. 462 mesh frames, two sessions, five print presentations.

The 2026-08-02 attempt on #25 recorded a blocker: the two classes never overlapped in exposure (face 105-120, print 55-85), so every candidate ratio was confounded with brightness. A lit room on 2026-08-04 pulled the face down to 40-78, into the print's band, which resolves it.

**That control kills the previous attempt's two best candidates.** `brow/socket_deep` had scored +0.626 with an extrapolation suggesting the face stays 0.179 above the print; the real dim face reads 0.478-0.952, **below** the print's 1.192-1.507. The extrapolation had the sign wrong, and a gate built on it would have rejected genuine faces in ordinary room light. Same for `nose/socket`. Both use an eye-socket region whose response collapses with exposure.

**Two survive**, across both sessions and every attacker presentation:

| ratio | face | print | separation |
|---|---|---|---|
| cheek/chin | 2.580-3.532 | 1.295-1.610 | 0.971 |
| forehead/chin | 3.910-6.266 | 1.286-1.785 | 2.126 |

The mechanism is in the raw regions: a face's chin returns 0.302 of its cheek (angled away from a coaxial emitter, shadowed by the jaw), a print's returns 0.655 (coplanar). Tilting the print, which drives the global centre/edge ratio to 1.51 against a genuine 1.44-1.54, moves this by 0.2 and never approaches the face; hand-curving, which raised the global ratio from 1.04-1.12 to 1.15-1.20 the same morning, leaves it at 1.295-1.432.

## Why no gate follows yet

- A print with a **darkened chin** would defeat it, and the corpus cannot test that: the banner is an evenly lit studio portrait. The strobe suggests a defence (a geometric shadow exists only in the lit frame, a printed one in both), but ambient was near zero in both captures, so the differential demonstrates only that the quantity is computable.
- One subject, and facial relief is the quantity being measured.
- Head pitch is never varied deliberately.
- Neither envelope #25 requires (dim IR-emitter-only, direct-sun saturation) is in this corpus.

## Testing

- Tooling validated as above; `cargo fmt --all --check` clean, clippy clean on irlume-auth.
- The doc's every number comes from the committed corpora, recomputed from the CSVs the tool wrote.
- No production code path changes, so no behavioural test applies; the example is developer tooling like `landmark_dump` and `burst_dump` beside it.


archledger:
Codex round complete: all six findings applied in 3313ce4, including two that were failures in my own writing.

**Reproducibility (finding 1).** The report's numbers could not be recomputed by anyone, and the PR body above claimed 'every number comes from the committed corpora' when no corpus was committed at all. That claim was false and I have corrected it here. The raw frames are infrared images of the operator's face and stay out of the repository like every other PAD capture, but that is no reason for the prose to be unauditable: a derived corpus now ships beside the report, one record per detected frame carrying region brightness means and no image, built by `scripts/landmark-relief-corpus.py`. `scripts/analyze-landmark-relief.py` regenerates every published figure from it and `--check` asserts the report still matches. It is proven to fail rather than decorate: an edited separation figure exits 1, and a corpus with one condition removed exits 1 on the population assertion before computing anything. It also caught a real defect on first run, a Unicode minus in the report where the tool emits ASCII.

**Silent partial corpora (finding 2).** Readdir errors were flattened away and an unreadable frame yielded exit 0 with a short corpus, which is the exact bias the tool's own comments claim it prevents. Every frame is now parsed before any output exists, a truncated frame aborts, and a directory already holding replay output is refused instead of mixed with stale CSVs. Verified: truncated frame exits 1 with zero files written; stale directory exits 1; good run exits 0. (My first check of this reported 'exit 0' for all three because I read the exit of a `tail` in a pipeline rather than the program.)

**Frame identity (finding 3).** Frames now keep their capture number. This mattered more than it looked: the lit/ambient differential in the report pairs frames by number arithmetic, so a re-labelled frame would have paired with the wrong neighbour. Verified on a gapped set: 00, 02, 04 keep those numbers instead of becoming 00, 01, 02. The corpora in this PR are contiguous and two-digit, so no published number was affected.

**Causal overclaim (finding 4).** Correct, and the contradiction was internal: the page said 'the ratio is a geometry measurement' while its own next section describes an attack that reproduces the signal with no geometry. It now presents the geometric reading as a hypothesis consistent with the medians, and states what matching exposure does not rule out: skin, facial hair, source-image shading, ink and coating, head pitch.

**Follow-ups**, both applied rather than deferred: the PGM parser checks the raster separator is whitespace and rejects trailing bytes, and the six print presentations are no longer described as two.

Replay parity re-verified after every change, on a third burst not used during development: 18/18 byte-identical.

PR #271, opened 2026-08-04
relief: what the banner shows about printed tone at 850nm, and what it does not

archledger:
Follow-up to #270 on #25. The darkened-chin attack that report called untested needs a second print, and the only instrument available is the original vinyl banner. This measures what the existing corpus can say about it, and after the review round, is explicit that the answer is not "the attack fails".

## Measured

Each region as camera brightness over the forehead brightness in the same frame (medians):

| region | real face | vinyl print |
|---|---|---|
| brow (eyebrow) | 0.728 | 0.961 |
| socket_deep | 0.424 | 0.721 |
| chin | 0.192 | 0.625 |

A real eyebrow reads 0.728 of forehead because hair absorbs 850nm; the printed eyebrow, strongly dark to the eye, reads 0.961. Across the sampled regions the face spans 0.192-1.000 and this print 0.625-1.137.

Defeating cheek/chin needs a printed chin at 0.363 of forehead, below every region sampled here.

## What that does not establish

An earlier revision of this PR concluded the attack was "not reachable on this instrument". That was wrong and is withdrawn. The six values are landmark regions of one portrait, differing simultaneously in source tone, surface orientation, emitter incidence, coating response and viewing angle, so the lowest of them is not the darkest tone the ink can print. Camera brightness over a printed forehead is also not a reflectance measurement, since reflectance is defined against a reference under stated geometry and none is in frame, so no absorption multiple follows from these numbers either.

The report now says what was measured, records that the earlier revision overclaimed, and keeps the requirement as a real test: a deliberately darkened or infrared-absorbing chin, printed or physically occluded, captured and refused. The occluder variant needs no new instrument.

## Testing

Every table in the report is now rendered from the committed corpus and matched verbatim, after the round found the guard incomplete: protecting the ink table had left the prose arithmetic as loose substrings, so editing 0.625 there passed because the table still carried it, and the 1.7 multiplier was never checked because it was never claimed.

Verified by mutating all 30 numeric table rows individually, each caught, plus the four prose figures the round named. My earlier tamper proof tested a single case and generalised; this one does not.

### DIFF
--- a/docs/pad-results/2026-08-04-landmark-relief.md
+++ b/docs/pad-results/2026-08-04-landmark-relief.md
@@ -123,16 +123,15 @@ by brightness alone on this hardware. This is an extrapolation from faces
 measured to 120 and is offered as a bound, not as evidence about saturated
 frames; saturation clips regions unevenly and is measured separately (#221).
 
-## The printed-shadow attack, measured on this instrument
+## The printed-shadow attack remains untested
 
 The cue reads brightness at the chin and cannot tell an emitter shadow from a
-printed one, so an attacker who darkens the chin region of the print should be
-able to raise the ratio without changing any geometry. That assumes a visibly
-dark print tone also reads dark at 850nm, which is a claim about ink rather
-than about geometry, and the corpus can test it: the banner carries printed
-tones whose real counterparts absorb infrared.
+printed one, so darkening the chin region of the source image should raise the
+ratio without changing the print's geometry.
 
-Each region as a fraction of its own class's forehead (medians):
+What the banner does show is how little contrast its current portrait retains
+at 850nm across the sampled regions. Each value is camera brightness divided by
+the forehead brightness in the same frame (medians):
 
 | region | real face | vinyl print |
 |---|---|---|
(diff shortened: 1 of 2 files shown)


archledger:
Codex round complete: both findings applied in b2a68f6, and both were corrections to claims rather than to misbehaving code.

**The ink-limit claim is withdrawn.** The round is right that six landmark regions of one portrait cannot establish what the ink can print: they differ at once in source tone, orientation, emitter incidence, coating response and viewing angle, so the lowest of them is not a gamut floor, and camera brightness over a printed forehead is not reflectance, so the 1.7x absorption figure did not follow either. The section now reports what was measured, states that an earlier revision claimed the attack was unreachable and that this does not establish it, and the requirement returns to an actual darkened or infrared-absorbing chin by any means.

This is the second time in two rounds that I turned an observation into a mechanism claim: the previous one called the chin ratio 'a geometry measurement'. The pattern is the same and the fix is the same, which is to state the observation and let the mechanism stay a hypothesis.

**The checker had the same hole twice.** Guarding the ink table left the arithmetic as loose substrings, so the round's reproduction is exactly right: editing the prose 0.625 to 0.425 passed because the table still carried 0.625, and 1.7 was never checked because it was never claimed. Every table in the report is now rendered from the corpus and matched verbatim.

The deeper problem was my verification, not the check. I tested one mutation, saw it caught, and generalised. Rerunning that way now: **all 30 numeric table rows mutated individually, every one caught**, plus the four prose figures named in the round. That sweep found three survivors before this fix, all from the same cause, a figure duplicated between a table and the prose around it.

PR #272, opened 2026-08-04
relief: the occluder attack, ambient fragility, and the strobe differential refuted

archledger:
The relief report required this test before any gate could be written. No second print exists, so it was run with household material over the existing banner: same camera, subject, and afternoon.

## Material decides the direction, and colour does not predict it

| chin covering | chin brightness (median) | vs bare print |
|---|---|---|
| none | 67.9 | — |
| black polyester | 143.4 | 2.11x brighter |
| black cotton sock | 25.2 | 0.37x |

Polyester **reflects** 850nm and made the chin brighter than the forehead (2.06, against the bare print's 0.692), which no face produces. Cotton absorbs strongly: nine of nineteen frames read the chin at exactly zero.

## It defeats the shipped centre/edge cue

| presentation | centre/edge | accepted |
|---|---|---|
| bare print | 1.08-1.15 | 6/6 |
| print + cotton | 1.31-1.53 | 6/6 |
| genuine face, same day | 1.44-1.54 | — |

A sock moves the print's global ratio into the genuine band. No verdict changes today, because the bare print already passes, but it forecloses the raise-the-floor direction harder than the tilt result did: a floor anywhere inside the genuine band is now reachable with a household item.

## Against the relief cue, the implementation decides

Face bands from the relief corpus: cheek/chin 2.580-3.532, forehead/chin 3.910-6.266. Of 19 occluded frames:

| cue | floor gate | band gate |
|---|---|---|
| cheek/chin | accepts 11/19 | accepts 1/19 |
| forehead/chin | accepts 9/19 | accepts 0/19 |

The nine chin-at-zero frames explain the gap: a fully absorbed chin gives an unbounded ratio, which a floor reads as maximally face-like. **A floor implementation is defeated by a sock.** Two rules for any future gate: band not floor, and a chin at zero is not measurable and must refuse rather than score.

## Why the band column is not a survival claim

Occluded frames span cheek/chin 1.31 to 3.69. That range sweeps **through** the genuine band, and one frame landed inside it with a hand-held sock and no aim. Coverage is a continuous parameter the attacker controls, so an attacker who knows the band exists can tune absorption to sit in it; this corpus traversed the whole path by accident in nineteen frames. The cue's premise, that a flat print cannot reproduce a real chin's response however presented, does not survive the addition of an absorber.

## Testing

Both corpora are committed beside the report, per-frame region means (59 frames) and the gate's own verdicts and cues (18 presentations). Every figure in the document was recomputed from them before publishing, including the ones in the prose.

## Not tested

A tuned occluder aiming at the band, other fabrics, a printed carbon-black chin, and the strobe differential, which remains the one idea that might separate an absorbing patch from real relief: an occluder's darkness is in the material and a shadow's is in the emitter geometry. Ambient was near zero in every capture here, so it is still uninvestigated.


archledger:
Codex round complete: all three findings applied in 660219b, plus the suggested checker.

**The genuine row.** Correct on both counts: neither new corpus held a genuine frame, so 'every number is recomputable from them' was false, and the committed 2026-08-02 genuine population spans 1.26-1.49, not the 1.44-1.54 I quoted. The same-day baseline is now committed in the gate corpus so both ranges are auditable, and the conclusion is stronger once stated properly: the cotton attack reached 1.5345, above every one of those 60 genuine samples, so no floor accepting the committed genuine population rejects it.

**Material properties.** This is my third instance today of reporting a camera reading as a physical property, after 'the ratio is a geometry measurement' and the ink-gamut claim. 'Black polyester reflects 850nm' is not something this corpus can establish: no reference surface was in frame, and angle, distance and coating all contribute. The text now says what was observed, camera brightness at one geometry, which costs the argument nothing since the attack conclusion rests on the readings rather than on why the materials produced them.

**The sweep.** Withdrawn. Coverage, placement and thickness were never recorded, so attributing the 1.31-3.69 spread to coverage was an inference about an uncontrolled variable, and the reviewer is right that pose or cloth intrusion explains the early frames' left/right cheek disagreement equally well. The section now states what the data show, one in-band sample of nineteen, and names the controlled coverage sweep as the missing experiment.

**Checker added** as suggested, since the relief checker cannot consume these corpora: it refuses zero denominators, which are the central observation here. Tables are rendered from the data and matched verbatim; all 9 numeric table rows were mutated individually and every one is caught.

One further change from work done after the round: I tried to run the strobe differential and could not, which the report now records rather than leaving as an open suggestion. In every capture so far the subject receives almost no ambient infrared, the emitter-off frame reading the cheek at 8-16 and the chin at 1-3 counts, so any ratio from it is quantization noise. Testing it needs an infrared-rich source falling on the subject.

PR #274, opened 2026-08-04
setup: offer the anti-spoof model as a recommended step

archledger:
Acts on the measurements from #235 and #25, and updates ADR-0001 to match.

## Why

The algorithmic gate contributes nothing against the demonstrated print. It returned Live for all 24 presentations in #235, and again for all six presentations of an enhanced attack measured 2026-08-04: a black cotton patch over the print's chin carries the centre/edge ratio from 1.06 into 1.32-1.44, inside the committed genuine population (1.26-1.49). The same enhancement defeats a floor-style implementation of the #25 relief candidate.

`flir` refused every one of those presentations, 0.988-1.000 bare and 0.998-0.999 occluded, against its 0.9 threshold, and has not failed to deny this attack across three sessions.

A default posture in which a printed photograph of an enrolled user passes is not defensible when a cue that stops it is one command away and most users will never run that command.

## What changed

`irlume setup` gains a step, after enrollment because that is the face a print impersonates. It shows the license and provenance, states that a printed photo passes without it, and defaults the answer to yes.

## What deliberately did not change

- **The weights are not shipped.** Distributing weights whose training data the publisher does not document is the ADR-0001 criterion-2 problem restated as redistribution, and the same thing that blocks commercial use of the model stack.
- **Nothing is fetched without consent.** The daemon does not download on first start; `models enable` keeps its stricter type-the-name gate for anyone arriving without context.
- **Deny-only** is unchanged: the cue can refuse a presentation the built-in gate accepted, never approve one it rejected.
- `flir` still fails criteria 2 and 3 and is still not warranted. Offering a model is not certifying it.

## The cost, in the ADR

The operating window is narrow and thinly measured: highest genuine 0.702, lowest attack 0.941, both one subject on one camera, threshold 0.9. Genuine-side failures are mapped rather than absent (dim strobe frames, direct sun) and a blown frame drops p_fake into an abstain band (#237). More users now inherit that false-rejection risk; every such failure falls back to the password, which is why the trade is acceptable, and widening the evidence past one subject is what would justify shipping rather than offering.

## Testing

- `fetch_and_enable` extracted so exactly one code path installs weights and applies the checksum pin; the two callers differ only in how consent is obtained.
- A test pins the offer: the catalog is non-empty, the entry carries the license and provenance strings the step prints before asking, and `pad_model_offer` resolves to it or correctly steps aside when one is already enabled.
- Full CLI suite green (323), clippy clean, fmt clean.

## Not tested

The interactive path end to end, which needs a TTY, root and a network fetch. The download, checksum and settings write are the pre-existing `models enable` code, unchanged apart from being lifted into a function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,12 +286,16 @@ The current gaps:
   brightness ratio on a 2D infrared sensor, and a life-size print held at an
   angle produces the same falloff a face does (1.02 to 1.58 over those 70
   presentations, against 1.26 to 1.49 for the live user), so no threshold
-  accepts the user and rejects the print. `sudo irlume models enable flir` adds
-  a trained, deny-only cue that flagged all 6 presentations of the same print
-  that reached it, at p_fake 0.941 to 1.000; that evidence covers one subject
-  and one print instrument. It is opt-in because its publisher documents
-  neither the training data nor a way to reproduce the model, which fails
-  [ADR-0001](docs/adr/0001-liveness-pad-strategy.md). Passive blink liveness also
+  accepts the user and rejects the print. **`irlume setup` offers a trained,
+  deny-only cue (`flir`) for this, and recommends it**: it refused the same
+  print at p_fake 0.941 to 1.000, and still refused at 0.998 to 0.999 when the
+  print was enhanced with an infrared-absorbing patch that carries the built-in
+  ratio into the genuine range. That evidence covers one subject and one print
+  instrument. irlume does not ship or warrant those weights, because their
+  publisher documents neither the training data nor a way to reproduce the
+  model, which fails [ADR-0001](docs/adr/0001-liveness-pad-strategy.md); setup
+  shows the license and provenance before fetching anything, and
+  `sudo irlume models enable flir` does the same later. Passive blink liveness also
   **doesn't cover glasses-wearers** (IR lens reflections hide the eyelid). Every
   miss falls **safely to the password**. See
   [ADR-0002](docs/adr/0002-challenge-response-liveness.md) and the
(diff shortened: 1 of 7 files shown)


archledger:
Codex round complete: all five findings applied in 9416d18.

**The consent bypass was real, and it was exactly what the ADR claimed could not happen.** `yes_no` returns its default *without printing the question* when stdin is not a terminal, so with `default_yes: true` a `sudo irlume setup </dev/null` would have fetched third-party weights with nobody having seen the license. A default may only apply to someone who was actually asked; a non-TTY stdin now declines and points at `models enable`. The repository's own `setup_walks_every_step_noninteractively` drives precisely that path, which is how it should have been caught before review.

**Reporting a config key as enabled.** Correct and worse than it looks: the daemon refuses an unknown catalog name, missing weights, or a checksum mismatch and runs without the cue in each case, so setup would have told such a user they had a print defence they did not have. That is worse than knowing they have none, because it stops them taking other precautions. `pad_model_offer` now re-checks what the daemon checks, a new `ConfiguredButBroken` state says so plainly, and a failed fetch now states the model was not enabled.

**My green-suite claim was wrong.** I ran `--bin irlume`, which skips the integration tests, and two of them assert the old `[1/6]` labels. Confirmed failing before the fix. They now assert `[1/7]`, `[3/7]` and `[7/7]`, and the setup test additionally asserts a non-interactive run does not fetch. The unit test's `AlreadyEnabled` arm did accept any ambient string; it now verifies the invariant the fix establishes, that `AlreadyEnabled` implies present weights matching their pin, with a separate discriminating case for a broken configuration.

**The ADR numbers.** I conflated two measurement sets. The six `padcapture` presentations with the model disabled carried the ratio to 1.31-1.53; the four occluded daemon-path runs with it enabled read 1.32-1.44 at p_fake 0.998-0.999. Both are now cited separately with their corpus paths. The `flir` scores were genuinely uncommitted on this branch when reviewed, because they were on #273; that has since merged and this branch is rebased on it, so `2026-08-04-flir-vs-occluder.jsonl` is present. Every figure in the ADR and README was recomputed against a committed corpus before pushing.

Full `cargo test -p irlume-cli` green across all six test binaries, clippy clean, fmt clean.

PR #275, opened 2026-08-04
models: support measured models irlume does not fetch, and document them

archledger:
Builds the two-tier model support and the user-facing page for it.

## The problem

The catalog conflated two different questions: whether irlume has **measured** a model, and whether irlume will **fetch** it. A model whose licence makes obtaining it the user's decision could not be listed at all, so a user had no way to learn which weights irlume actually supports and has tested. The alternative, free-form bring-your-own, fails for a reason this repository already documents: a deny-only cue firing where neither population was observed "guesses against the user", and every false fire costs a real login.

## The design

`url` becomes `Option`. `None` means bring-your-own: measured identically, pinned identically, but installed by the user with

```sh
sudo irlume models add <name> /path/to/weights.onnx
```

which verifies the file against the sha256 irlume measured and refuses anything else. That pin is the point: it is how irlume knows *which* model is loaded rather than trusting whatever file appeared in the directory, and it means a bring-your-own model still gets a measured threshold rather than a guessed one.

- `install_verified` is the single installer for both tiers, so a future caller cannot install unpinned weights by forgetting the check; `pin_matches` is a value so the refusal is testable.
- `setup` only ever offers a fetchable entry, since it cannot act on a model it may not download.
- The listing states per model how to obtain it, and its header no longer claims everything is fetchable.

## The page

`docs/THIRD-PARTY-MODELS.md` lists what irlume has measured, the threshold it measured and why the publisher's default is never used, the mapped genuine-side failures, and what is deliberately absent: any model this project has not measured. The bring-your-own section is empty today and says so, because nothing non-permissive has been measured yet. The page also states that whether a licence permits a given user's use is that user's determination.

## Testing

- Catalog invariant test now covers both tiers: a fetchable entry must name an https origin, a bring-your-own entry names none, and **both** must carry a 64-char pin.
- `pin_matches` test asserts unpinned bytes are refused, with the fixture checked not to collide with the real pin.
- Setup-offer test updated to require a fetchable entry rather than any entry.
- Exercised on the machine: a random file is refused showing both hashes; the genuine artifact passes the pin and stops at the consent prompt, which correctly refuses a non-TTY stdin.
- Full `irlume-cli` and `irlume-common` suites green, workspace clippy clean, fmt clean.

## Not tested

An end-to-end `models add` that actually enables, which needs a TTY; the install path past consent is `install_verified`, shared with the fetch path that is exercised today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,6 +464,7 @@ see [Honest limitations](#-honest-limitations).
 | **Write software that drives irlume** | [`docs/INTEGRATION.md`](docs/INTEGRATION.md) |
 | Look up the versioned **machine API**, field by field | [`docs/MACHINE-API.md`](docs/MACHINE-API.md) |
 | Understand the **architecture** | [`docs/ARCHITECTURE.md`](docs/ARCHITECTURE.md) |
+| Choose a **third-party liveness model** irlume has measured | [`docs/THIRD-PARTY-MODELS.md`](docs/THIRD-PARTY-MODELS.md) |
 | Read the **threat model** and standards mapping | [`docs/THREAT_MODEL.md`](docs/THREAT_MODEL.md) · [`docs/STANDARDS.md`](docs/STANDARDS.md) |
 | **Verify the claims** on my own machine | [`docs/VERIFY.md`](docs/VERIFY.md) |
 | **Debug** a login or trace every stage | [`docs/DEBUGGING.md`](docs/DEBUGGING.md) |
--- a/crates/irlume-cli/src/commands.rs
+++ b/crates/irlume-cli/src/commands.rs
@@ -1449,7 +1449,15 @@ pub fn setup(args: &[String]) -> ExitCode {
         PadOffer::ConfiguredButBroken(name, why) => {
             println!("  '{name}' is configured but the daemon will not load it ({why}),");
             println!("  so this machine has NO trained print defence right now.");
-            println!("  repair with: sudo irlume models enable {name}");
+            // The repair differs by tier: irlume can re-fetch one it fetches,
+            // and can only ask for the file again for one it does not.
+            match irlume_common::thirdparty::by_name(&name).and_then(|m| m.url) {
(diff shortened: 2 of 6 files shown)


archledger:
Codex round complete: all four findings applied in 97149de, and the mutation check that followed found a fifth problem in my own fix.

**The install regression was mine and the consequence is worse than a failed update.** I replaced the rename-from-temp with `std::fs::write`, which truncates. `settings.conf` keeps naming the model, so an interrupted write leaves a checksum mismatch the daemon answers by running *without* the cue, and the gate that remains accepts the print attack. That is a quiet fail-open of the only defence that stops it. The install now goes through `write_atomic_mode`, which the common crate already tests for durability and for leaving no stray temp file.

**Both new tests passed for the wrong reason, and I only found out by mutating.** As an unprivileged process they could not write to the real state dir, so `install_verified` returned FAILURE whether the pin guard existed or not: deleting the guard entirely left the suite green. The refusal test now sandboxes the state dir so the write would genuinely succeed, and carries a positive control with matching bytes, because a refusal that might come from an unwritable directory proves nothing about the pin. Verified after the change: the guard-deleting mutant is killed by exactly that test.

That test also called `restart_daemon` on its success path, so it restarted the real daemon and took 50 seconds. `install_verified` is now split: `place_verified` does the pin, the atomic write and the settings record, and the restart stays with the caller.

**The tier test exercised nothing**, correct: the catalog has no bring-your-own entry, so the `None` branch was never reached. A fixture supplies one, asserting that a urlless entry is still pinned and that `fetch_and_enable` refuses it before any download rather than trusting call sites to check.

**The repair advice** now names `models add` for an entry irlume does not fetch.

One thing deliberately left uncovered, stated rather than papered over: no test proves the call site picks the atomic path over a truncating one, because a unit test cannot crash mid-write. That mutant survives. The property belongs to `write_atomic_mode` and is tested there.

Issue #276, opened 2026-08-04
Bring-your-own models beyond liveness: which pipeline stages can safely be opened


archledger:
irlume's shipped model stack is chosen for a commercially-clean bill of materials: YuNet (MIT) for detection, AuraFace (Apache-2.0) for recognition. That constraint is irlume's, not necessarily its users'. Someone running this on their own machine may reasonably want a model irlume cannot ship, and today the only stage they can change is the liveness cue.

#275 built the mechanism for that one stage: a catalog entry can be measured and pinned without irlume fetching it, installed with `models add <name> <path>` and verified against the sha256 irlume measured. This issue is about extending that to other pipeline stages, and about the fact that the stages are not equally safe to open.

## Why the PAD case was easy

The third-party liveness cue is **deny-only**: it can refuse a presentation the built-in gate accepted, and can never approve one it rejected. A hostile, broken, or simply mismatched model there causes false denials, and every denial falls back to the password. That property is what makes an unaudited artifact tolerable.

No other stage has it.

## The stages, by risk

**Recognition (the embedder) is grant-capable, and has a second, independent failure.**

A swapped embedder that produces similar vectors for many faces does not fail loudly. It authenticates strangers while the legitimate user's own logins keep working, so nothing surfaces the problem.

Separately, and true today regardless of this issue: `Enrollment` records no embedder identity. Templates are vectors in one model's embedding space, and cosine similarity between spaces is meaningless. If the embedder changes, irlume compares new embeddings against old templates and produces a number that has no interpretation, which may sit either side of the threshold. That is a correctness bug waiting for the first person who swaps a model or for a future BOM change, and it should be fixed on its own merits.

**Detection and landmarks are mixed.** Most failures are non-detections, which deny safely. But the mesh feeds the liveness cues, and bad landmarks do not raise an error, they produce confident numbers from the wrong pixels. The 2026-08-04 relief work has a worked example: chin landmarks fell outside a 400-row frame, the sampler clamped at the border, and two strobe phases "disagreed" for a purely geometric reason. A gate reading that would have scored garbage.

**Liveness (PAD): already open**, for the reason above.

## The measurement bar differs too, and that is the harder constraint

This project's rule is that a threshold comes from where the two populations were actually observed on real hardware, never from a publisher's default. That is satisfiable for a PAD cue with one subject and a print: the question is "does it deny this attack".

[...]

PR #277, opened 2026-08-04
core+auth: tag templates with the recognizer that produced them

archledger:
Step 1 of #276, and a latent bug worth fixing on its own merits.

## The bug

Cosine similarity is only meaningful *inside* one embedding space. `Enrollment` recorded nothing about which recognizer produced a template, so if the model ever changed, irlume would compare fresh probes against vectors from a different space. That does not error: it produces a number with no interpretation, free to land either side of the threshold. A silent grant or a silent denial, with nothing to indicate which.

This bites on any change to the shipped weights, not only on a user-supplied model. It simply has to be true before #276 can consider opening the recognition stage at all.

## The fix

`FaceScan` gains `embed_space`, `"embed:<sha256>"` of the recognizer's weights (the full 64-hex digest; a truncated hash halves collision strength per dropped character, and this tag exists to resist an adversarial model). One `std::fs::read` feeds both the digest and the ONNX session, so the tag always describes the weights that are running; the IR adapter loader got the same single-read treatment.

Compatibility is decided in one place, `storage::recognizer_space_matches`:

- A tagged template matches only its own recognizer's hash.
- An untagged template (every enrollment written before this PR) is comparable **only when the loaded recognizer is the one that could have produced it**: the shipped `glintr100.onnx`, whose digest is pinned as `LEGACY_RECOGNIZER_SPACE` and matches `models/SHA256SUMS`. Grandfathering untagged scans into *any* space would hand every existing enrollment to whatever model gets loaded, which is the hole this PR closes.

Every comparison site applies it:

- `rgb_scans_in` at the single point both RGB match paths gather templates.
- `ir_match_in`: the recognizer produces the raw IR embedding, so its identity gates IR matching too. One filter covers fusion, IR fallback, the calibrated centroid, and dark IR-only auth.
- `refit_profile_calib` skips foreign-recognizer pairs rather than fitting one calibration across incompatible spaces.
- `colliding_profile` / `enroll_merge_target` take the engine's space, so a re-enrollment under a new model can never merge into or be rejected by templates from the old one.

## Testing

- An engine test asserts `Engine::load`'s tag of the real shipped weights equals `LEGACY_RECOGNIZER_SPACE`, tying the loader, the constant, and `models/SHA256SUMS` together.
- Storage, IR-match, collision, and refit tests each prove their filter excludes a foreign space and keeps the same space, including a foreign tag on a vector byte-identical to the probe (the case where only the filter, not the score, can refuse).
- Seven mutants, each deleting or weakening one guard (blanket `None` grandfathering back, `Some` arm to `true`, each of the four call-site filters removed, digest truncated back to 12 hex): all seven fail named tests that match the mutant's semantics, run with `--no-fail-fast`, tree restored byte-identical and the suite re-run green afterwards.
- Full workspace suite green (1084 tests), workspace clippy clean with `-D warnings`, rustfmt clean.

## Not covered

- The check-then-act window itself (a file swap between hash and session load) is closed structurally by the single read but has no test; producing that race deterministically needs filesystem fault injection.
- No authentication was scored across two genuinely different recognizers on hardware; the exclusion is established at the template-gathering layer.
- The user-visible half: irlume does not yet *tell* someone their templates were made by a different recognizer, it just declines to match them, which presents as a failed login falling back to password. A message naming re-enrollment as the fix belongs with the `stage` work in #276, where there is a reason for the situation to arise.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -35,6 +35,10 @@ pub struct Engine {
     /// Stored on every new scan and matched against at verify, so an adapter
     /// swap/removal degrades to "re-enroll" instead of scoring across spaces.
     ir_space: String,
+    /// The recognizer's own embedding space, `"embed:<sha256 prefix>"` of its
+    /// weights. Stamped onto every scan enrolled and required to match at
+    /// verification: cosine scores are only meaningful inside one space.
+    embed_space: String,
     /// Optional MediaPipe FaceMesh: dense landmarks for the passive EAR blink
     /// liveness (ADR-0002). Loaded iff the model file is present; `None` disables
     /// the opt-in passive-liveness gate (it can't run without landmarks).
@@ -528,11 +532,22 @@ fn top_detection(faces: &[Detection]) -> Option<&Detection> {
 
 impl Engine {
     pub fn load(det_path: &str, model_path: &str) -> irlume_common::Result<Self> {
+        // Identify the recognizer by its weights, not its path: a file swapped
+        // in place under the same name is a different embedding space and must
+        // not silently score against templates from the old one.
+        let embed_space = match std::fs::read(model_path) {
+            Ok(b) => format!("embed:{}", &irlume_common::thirdparty::sha256_hex(&b)[..12]),
+            // Unreadable here means Embedder::load_from_file below will fail
+            // too; keep a value that can never equal a real tag rather than
+            // one that compares equal to everything.
+            Err(_) => "embed:unknown".into(),
+        };
         Ok(Self {
             det: Detector::load_from_file(det_path)?,
             emb: Embedder::load_from_file(model_path)?,
             ir_adapter: None,
             ir_space: "raw".into(),
+            embed_space,
             mesh: None,
             blaze: None,
             tp_pad: None,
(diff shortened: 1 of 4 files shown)


archledger:
**Do not merge.** The review round found this fix incomplete in ways that matter, and the session ended before they were addressed. Recorded here so the next session starts from the right place.

Two Critical, both meaning cross-space authentication is still reachable:

1. **Legacy templates stay comparable.** Grandfathering `embed_space: None` was meant as a migration concession, but it means every pre-existing enrollment on every machine remains matchable against a different recognizer. That is the whole population today, so the guard protects almost nothing in practice.
2. **Only the RGB path was filtered.** IR matching still scores templates from another recognizer; `ir_space` guards the adapter, not the model beneath it.

Two High:

3. **The digest and the loaded session can come from different files** — the weights are read once to hash and again by `Embedder::load_from_file`, so the tag need not describe what is actually running.
4. **Enrollment collision and merge decisions** still compare across spaces.

One Medium: a 12-hex-character prefix is short for a guard that is adversarial rather than merely accidental.

The direction stands, and the underlying bug this PR describes is real. The implementation is not sufficient.

archledger:
All six review findings are addressed in 90d577a.

- **Legacy templates under a future model (Critical):** untagged scans now match only when the loaded recognizer is the shipped `glintr100.onnx`, pinned as `LEGACY_RECOGNIZER_SPACE` (full digest, re-verified against `models/SHA256SUMS` and the file itself). The blanket grandfathering is gone.
- **IR paths (Critical):** `ir_match_in` takes the recognizer space and filters `embed_space` before `ir_space`, covering fusion, IR fallback, the calibrated centroid, and dark IR-only auth; `refit_profile_calib` applies the same filter to the pairs it fits.
- **Digest/session double read (High):** `Engine::load` reads the model once, hashes those bytes, and builds the session from them via `Embedder::load_from_memory`. The `embed:unknown` branch is gone; an unreadable file is now the I/O error it always was. The IR adapter loader had the same shape and got the same fix (its 12-hex `ir_space` format is unchanged because existing enrollments carry it).
- **Collision and merge (High):** `colliding_profile` and `enroll_merge_target` take the engine's space; a test proves a foreign-tagged template identical to the probe is invisible to both, and visible again under its own recognizer.
- **12-hex tag (Medium):** the tag is the full 64-hex digest.
- **Test fixture (Low):** the misleading assertion is replaced; a fully tagged foreign set now yields an actually empty result, and the legacy-only case is asserted by name.

Each new guard was mutation-checked individually: seven mutants (blanket grandfathering restored, equality arm to `true`, each call-site filter deleted, digest re-truncated), all killed by named tests, tree restored byte-identical and the suite re-run green. Not covered: the hash/load race has no test (needs fs fault injection; the single read closes it structurally), and no cross-recognizer authentication was scored on hardware.

PR #278, opened 2026-08-04
models: every catalog entry names its pipeline stage, and doctor reports per stage

archledger:
Stage 2 of #276, after #277 closed stage 1.

## Why now

The catalog was implicitly PAD-only. Nothing recorded which pipeline stage an entry plugs into, so the first recognition or detection entry added to it would have been installable by the existing commands and wired by the daemon as an anti-spoof cue. The stages are deliberately not equally open (#276): PAD is deny-only, so a bad model there costs retries or the password; a bad recognizer authenticates strangers while the legitimate user's logins keep working. That difference has to be something the code checks.

## What this adds

- `ThirdPartyModel.stage`, with `Stage::open()` as the single gate. Only PAD is open. Detection, landmarks, and recognition are named but closed, so the catalog, CLI, docs, and machine API share one stage vocabulary before those stages open.
- Refusals at three layers: `models enable`/`add` early (for the message), the installer choke point `place_verified` (so no future caller can bypass it), and the daemon's settings.conf wiring (root-editable text the installer never saw).
- Per-stage reporting from one builder, so the human and machine views cannot disagree:
  - `irlume doctor` gains a pipeline stages section: each stage's resolved model file, labelled `shipped` or `env-override`, with missing required files as failures. New check ids `stage-detection-model`, `stage-landmarks-model`, `stage-recognition-model`, registered in the MACHINE-API.md table.
  - `irlume models list --json`: new `models-list-json` capability. Reports every stage in order, whether it is open, the active model and origin, and for the open PAD stage the catalog with each entry's tier (`fetched` vs `user-supplied`) and weight state. Needs no daemon, so it still answers when the daemon will not start.
- `third_party.enabled.known` is honest about observability: settings.conf is root-only, and an unprivileged caller that read no name has not established that nothing is enabled. `known: false` is not "disabled".

## Testing

- Workspace suite green (1088 tests), clippy `-D warnings` clean, rustfmt clean.
- `scripts/machine-api-conformance.py` passes 36/36 against this build, including the new capability and fixture; the fixture was captured from the real engine and validates against the updated schema.
- Five guard-deletion mutants, each killed by a named test written for that guard: every stage forced open, the installer gate deleted, `third_party` attached to closed stages, unknown enabled-state collapsed to disabled, the env-override origin mislabelled as shipped.
- The closed-stage installer test carries a positive control: the same bytes install under an open stage, so the refusal is proven to come from the stage gate and not the sandbox.

## Not covered

- The daemon-side refusal (a closed-stage name hand-written into settings.conf) has no test; the decision it applies (`Stage::open`) is pinned, but the daemon branch itself only logs and skips.
- No closed-stage entry exists in the shipped catalog, so the gates guard a future addition; a catalog test says exactly that and fails the day one lands, pointing at the refusals to verify.


archledger:
All three merge-blocking findings and the follow-up are addressed in 14d7ffa.

- **"active" overclaimed (High):** the field is now `candidate`, resolved from this process's search order with `observed` and `readable` established by opening the file and checking it is a regular file — a directory or unreadable path reports `readable: false` and stops the search instead of falling through to a candidate the daemon would never reach. Origins are `caller-env` and `shipped`. The docs, doctor output, and schema all say candidate and state that a systemd drop-in is invisible to a shell, so an authoritative loaded-model report can only come from the daemon. (Your proposed `File::open`-only check would still pass a directory on Linux; the fix also checks `metadata().is_file()`.)
- **Failed read reported as disabled (High):** `config::observe_kv` classifies Value / Absent / Unknown; `read_kv` wraps it unchanged for existing callers. The enabled state is now keyed on what the read established, for root and non-root alike: observed absence is `known: true, name: null` (the config directory is world-readable, so a missing file is a real observation from any caller), a failed read is `known: false`. Your ENOTDIR test is adopted nearly verbatim and passes without root.
- **Schema too permissive (Medium):** `minItems: 4` plus `contains` for the four current stages, `third_party` tied to `open` in both directions, and `enabled` as the proposed `oneOf`. All six counterexamples I could derive from yours are rejected by `jsonschema.Draft202012Validator` while the real engine output validates. One deliberate divergence: `stage` stays an open string rather than a closed enum, because the documented contract says the stage list may grow within contract 1 and a validating consumer must not break the day it does; the four current names are guaranteed present by the `contains` constraints, which is the half of that counterexample that was a real contradiction.
- **Env restore (Low):** both installer tests now restore the two variables independently.

Two new guard-deletion mutants (readable computed as `exists` again; a failed read reported as absence) each fail a named test. Conformance passes 36/36 against the rebuilt binary and the re-captured fixture.

PR #279, opened 2026-08-05
auth+daemon: wire a third-party recognizer, dark-launched behind the stage gate

archledger:
Stage-4 plumbing for #276, landing while `Stage::Recognition` stays **closed**: every new path refuses today, and opening the stage is a deliberate one-line flip that ships with the first measured catalog entry. The measurement side is done — the split-source protocol proposal and buffalo_l's full FAR tables are on #276 — so this is the wiring that makes an entry loadable once the live genuine floor is captured.

## What it wires

**Engine policy, one builder call.** `with_thirdparty_recognizer(threshold)` does two inseparable things: the entry's measured RGB threshold replaces the shipped constant (`rgb_grant_threshold` is now the single place both RGB match paths and the enrollment collision decision get their bar), and IR matching dies wholesale. The IR thresholds and the fusion Platt calibration are measurements of the shipped model's cosine scale; no catalog entry carries IR-side measurements, so fusion, IR fallback, calibrated centroid, and dark IR-only auth are all off under a foreign recognizer. `ir_match` is the choke point — no present or future caller can score an IR template — and the dark path refuses with a reason naming the real limitation ("its IR matching is unmeasured") instead of a generic miss that would send the user debugging their enrollment. `refit_profile_calib` stops fitting calibrations that would imply dark support which cannot exist. The liveness gate is untouched: it reads pixels, not embeddings.

**Daemon resolution, pure and refusal-first.** A new `third_party_recognizer` settings key (per-stage keys — one key naming entries of two stages would make "what is enabled" ambiguous) resolves through `recognizer_override`, a pure decision function: not-in-catalog, wrong-stage, and stage-closed each refuse with their own reason. Any refusal falls back to the **shipped** recognizer, unlike the PAD cue's fall-back-to-nothing, because a daemon with no recognizer cannot authenticate anyone. The file read and pin check mirror the PAD block; `IRLUME_THIRDPARTY_RECOGNIZER` is the sandbox env override, same shape as the PAD one.

**Installer.** `place_verified` writes the stage-appropriate key via `settings_key_for`, so a recognition entry can never be named under the PAD key.

**Template safety needs nothing new**: #277 already tags every template with the full digest of the loaded weights, so a foreign recognizer's templates are isolated automatically, and #278's stage gate already refuses installation.

## Testing

- Workspace green (1094 tests), clippy `-D warnings` clean, rustfmt clean.
- Eight guard-deletion mutants, each killed by a named test: threshold field ignored, builder forgetting the IR shutdown, the `ir_match` choke deleted, the refit gate deleted, the collision threshold reverted to the constant, the override's stage check deleted, its open check deleted, and the recognizer key collapsed onto the PAD key.
- The engine policy test carries positive controls both directions: IR matching proven working before the flip and again after the restore.

## Not covered

- The daemon resolution closure runs only at startup and is not unit-tested; the decision it composes (`recognizer_override`, the pin check) is. Same standing as the PAD wiring block.
- No end-to-end authentication ran under a third-party recognizer: no catalog entry exists yet, which is the point of the dark launch. The full path gets exercised on hardware when the buffalo_l entry lands with the stage flip.
- `models disable` and the human `models` listing still speak only the PAD key; generalising the disable UX belongs with the first recognition entry, when there is something to disable.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -39,6 +39,18 @@ pub struct Engine {
     /// weights. Stamped onto every scan enrolled and required to match at
     /// verification: cosine scores are only meaningful inside one space.
     embed_space: String,
+    /// The RGB match threshold for THIS recognizer. The shipped constant for
+    /// the shipped model; a third-party recognizer brings its own measured
+    /// value (#276), because a threshold is a property of one model's cosine
+    /// scale and applying another model's number to it is a guess.
+    rgb_threshold: f32,
+    /// Whether IR matching (fusion, IR fallback, calibrated centroid, dark
+    /// IR-only auth) may run. False under a third-party recognizer: the IR
+    /// thresholds and the fusion Platt calibration are measurements of the
+    /// SHIPPED model's cosine scale, and no catalog entry carries IR-side
+    /// measurements yet. RGB-only matching stays; the dark path refuses with
+    /// its own reason and falls back to the password.
+    ir_matching: bool,
     /// Optional MediaPipe FaceMesh: dense landmarks for the passive EAR blink
     /// liveness (ADR-0002). Loaded iff the model file is present; `None` disables
     /// the opt-in passive-liveness gate (it can't run without landmarks).
@@ -563,6 +575,8 @@ impl Engine {
             ir_adapter: None,
             ir_space: "raw".into(),
             embed_space,
+            rgb_threshold: irlume_core::RGB_MATCH_THRESHOLD,
+            ir_matching: true,
             mesh: None,
             blaze: None,
             tp_pad: None,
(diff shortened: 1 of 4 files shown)


archledger:
Both findings accepted and fixed in 7b43f62.

- **Silent fallback (Critical):** the fallback-to-shipped policy is gone. An explicit `third_party_recognizer` selection that cannot be honoured — unknown name, wrong stage, closed stage, unreadable file, pin mismatch — now refuses to start, with the message naming the consequence ("face auth falls back to the password"). Your premise was right and mine was wrong: the daemon already exits on shipped-model failure and PAM already treats an unavailable daemon as password fallback, so fail-closed costs nothing and substituting a different grant-capable recognizer against retained templates was exactly the hole you described. The resolver is extracted (`resolve_thirdparty_recognizer`) so a re-exec test can observe both refusal exits; mutating `exit(1)` to `exit(0)` is caught.
- **Threshold/bytes binding (Critical):** `Engine::load_with_recognizer_bytes` takes the caller's verified buffer (your proposed shape, adopted nearly verbatim); `Engine::load` wraps it, and the daemon retains the verified bytes for the engine factory, closing both the initial-load window and the post-panic rebuild that repeated no pin check. Your regression test is in as `verified_recognizer_bytes_are_the_loaded_embedding_space`; a wrong-buffer digest mutant is killed by four tests.
- The follow-up note is agreed: `models disable` and the listing stay PAD-only until the commit that opens `Stage::Recognition`, and that commit must generalise them.

One pre-existing flake surfaced during validation (the identify rate-limiter test failed once in four full-suite runs, unrelated file); three consecutive daemon-suite runs are clean with the new tests in.

PR #280, opened 2026-08-05
models: open the recognition stage with buffalo_l, measured under the split-source protocol

archledger:
The final commit of #276's staging: `Stage::Recognition` opens, carrying its first measured entry.

## The entry

`buffalo` — buffalo_l's `w600k_r50.onnx`, pinned to the artifact inside the official InsightFace v0.7 release (verified by downloading the publisher's zip and matching it byte for byte), threshold **0.55**, bring-your-own because the licence is non-commercial research only and the training data is scraped without consent.

The measured basis is committed as `docs/recognition-results/2026-08-05-buffalo-l.md`, produced under the split-source protocol recorded on #276:

- **False-accept side**, three legs through irlume's own pipeline with shipped-model controls that reproduced every documented FAIRNESS.md figure to the digit: full identity-grouped LFW (87.3M pairs, EER 3.9% vs the shipped 4.2%), all seven FairFace groups individually (spread 4.5× vs 10×, and stated honestly: the worst-served group *shifts* to Middle Eastern and reads slightly worse there than under the shipped model), and SFHQ with its exact subset named.
- **False-reject side**, live floors on two cameras (Zenbook built-in and the Logitech BRIO), both sessions anchored by shipped-model control runs that landed on the documented live floor (0.71/0.85).
- **Why 0.55 and not 0.60**: at 0.55 the worst demographic group's FAR matches the shipped stack's own worst group at its operating point; 0.60's cross-lighting margin measured zero (side-lamp production-shaped minimum 0.553).

## The flip and its follow-up, same commit as required by the #279 review

- `Stage::Recognition.open()` flips, arming the dark-launched #279 wiring: the entry replaces the recognizer for RGB matching only, IR matching/fusion/dark login stay off (unmeasured for this model), and an unsatisfiable selection refuses to start rather than substituting.
- `models disable [name]` clears exactly the selected entry's stage key; bare `disable` keeps its old meaning while only one thing is enabled.
- The listing and both consent prompts derive from one `role_line` per stage — a recognizer entry says REPLACES, IR off, re-enroll — because wrong text there is wrong consent.
- The machine API's `third_party` object is per-stage: its own settings key, its own catalog subset. Fixture recaptured; conformance 36/36 in both modes.
- The PAD cue invariant tests were silently asserting P(fake) semantics over the whole catalog; they now sweep PAD entries only, with non-emptiness asserted so the narrowing cannot go vacuous.

## Testing

Workspace green (1096 tests + the additions), clippy `-D warnings` clean, four new guard-deletion mutants killed by named tests (per-stage catalog filter, per-stage key, recognition consent text, enabled-entry key pairing), on top of the ten from #279's branch. Conformance and fixtures green in strict and published modes.

## Not covered

- No end-to-end authentication has run under buffalo_l yet: enabling it on a live machine (models add → daemon restart → re-enroll → login) is the first-use validation, and doing that on a machine whose login the user needs is deliberately left as an explicit decision, not a side effect of merging.
- One subject, glasses on, two cameras, one night of live capture — stated in the measurement doc's scope section.


archledger:
All five findings addressed in a92f72e.

- **Evidence not auditable (High):** committed. `docs/recognition-results/data/` now holds the seventeen raw irbench outputs (aggregate distributions only — irbench emits no per-image data), `MANIFEST.json` with the exact commands, datasets, subsets and digests, and `floor-aggregates.json`: the live sessions reduced to counts, brightness ranges and cosine statistics by a committed derivation script, verified byte-identical to a regeneration. The raw capture records hold biometric embeddings and deliberately never enter the repository — the committed reduction is the same pattern the PAD relief work settled on. On your alternative of re-closing the stage: the requirement behind it was auditable evidence, which now exists in-tree, so the flip stands.
- **Contradictory fairness claims (High):** your arithmetic was right and the prose was wrong. Every claim is now scoped to its threshold — 3.9× vs 6.1× with worst-group parity at the enabled 0.55 (parity by construction: the operating point was chosen for FAR parity with the shipped stack), 4.5× vs 10× with the worst group halved at the 0.50 point FAIRNESS.md keys on. The catalog summary, the THIRD-PARTY-MODELS entry, and FAIRNESS.md all updated; FAIRNESS.md's training-data attribution corrected to WebFace600K per the model zoo.
- **Silent restart failure (High):** `restart_daemon` returns the failure and both callers refuse to claim activation, with the enable path explicitly warning not to enroll until a manual restart succeeds. A PATH-faked `systemctl` test observes both outcomes.
- **Deletion errors as absence (Medium):** only `NotFound` counts as deleted; anything else leaves the selection untouched. Tested with a directory-shaped weights path.
- **Doctor/TUI PAD-only (Follow-up Medium, partially taken now):** `doctor_line` reports every enabled stage, tested against a sandboxed recognition selection. The TUI's model screen remains PAD-shaped and is deferred as you allowed; the machine API was already per-stage.
- **Stale comments (Low):** module and variant docs corrected.

Workspace 1101 tests green, clippy `-D warnings` clean, strict conformance 36/36.

PR #282, opened 2026-08-05
auth: the device selection carries IR availability, not a load-time probe

### LINKED ISSUE #281: Engine snapshots ir_available independently of the daemon's pair selection, and can lose a startup race into permanent convenience tier
On archhost (NexiGo N930W, daemon at c3f15ab), every `auth test` denies with `convenience(RGB-only) … (Unknown)` even though startup logs `cameras rgb=/dev/video0 ir=/dev/video2 (secure tier)` and IR capture demonstrably works — enrollment through the daemon captured IR frames minutes earlier on the same boot.

The mechanism, established tonight:

- The daemon's own `select_pair` finds the IR pair and logs secure tier.
- `Engine::load` takes its own, independent snapshot: `ir_available: irlume_camera::capabilities().ir_pair`, once, at load time.
- At startup, load runs immediately after the emitter setup has been holding the IR node ("IR emitter ready" logs one second before the engine's model-load line). `capabilities()` → `list_pairs()` appears to lose that race and returns no pair, so the engine snapshots `ir_available: false` and sits in convenience tier for the daemon's whole life.
- The probe is not the sandbox and not the hardware: the identical `status --json` probe inside a `systemd-run` sandbox with the unit's hardening returns `ir: true` when the cameras are idle, and does the same from a plain shell.

Consequence: on a machine that loses this race deterministically, the daemon logs secure tier but enforces the convenience policy — every non-screen-unlock authentication denies before matching, and the dark path is unreachable. On the Zenbook the race apparently never loses, which is why this was invisible until the first `auth test` ever run on archhost (tonight's buffalo first-use validation).

Suggested direction: the engine should not take an independent second snapshot of a fact the daemon has already established — `with_devices` (or the daemon explicitly) should derive IR availability from the selected pair, with the existing `NO_IR` sentinel convention preserved for tests. A lazy re-probe on first use would also close it, at the cost of keeping two sources of truth.

Also observed on the same boot, likely related to archhost's environment rather than this bug: the emitter journal lock refusal (`/run/lock/irlume-emitter-….lock: Permission denied`) and the #207 consumer-scan warning — both also reproduce under the packaged 0.7.0 unit, and the identical sandbox writes `/run/lock` fine at idle, so the lock refusal may share the startup-timing character of the capability race.

archledger:
Fixes #281 — full mechanism and the archhost evidence are on the issue. `Engine::load`'s one-shot `capabilities()` probe can lose a startup race against the emitter setup holding the IR node; the engine then enforces convenience-tier policy for its whole life while the daemon logs secure tier from its own selection. Found by the first authentication ever run on archhost, during buffalo_l's first-use validation.

The fix makes the device selection carry IR availability: `with_devices` and `set_devices` derive it from the selected IR path existing — the daemon's own definition of usable, applied to the same pair its tier log uses — so a runtime camera switch retiers as well. The `NO_IR` test sentinel stays a nonexistent path and keeps reading unavailable; an engine that never receives devices keeps the load-time probe.

Tested: a new engine test drives both directions through both entry points (existing path retiers Secure, nonexistent reads Convenience); two stale-snapshot mutants (each assignment deleted) fail it. Workspace 1102 tests green, clippy clean. After merge, archhost re-runs the buffalo grant test that this bug blocked.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -548,6 +548,26 @@ pub fn thirdparty_abstains(p_fake: Option<f32>, threshold: f32) -> bool {
         .is_some_and(|p| p >= irlume_common::thirdparty::MEASURED_GENUINE_CEILING && p < threshold)
 }
 
+/// IR availability for a caller-selected IR device path.
+///
+/// The path half answers #281 (the selection, not a racy probe, is the truth);
+/// the forced-off half preserves `IRLUME_FORCE_NO_IR=1`, the documented
+/// drop-to-convenience override, which must outrank the selection exactly as
+/// it outranks `capabilities()` — the first cut of #282 overwrote it and
+/// silently re-secured a forced-convenience machine whose IR node existed.
+fn selected_ir_available(ir: &str) -> bool {
+    ir_selection_available(
+        std::path::Path::new(ir).exists(),
+        irlume_camera::ir_forced_off(),
+    )
+}
+
+/// The decision itself, as a value: testable without touching the
+/// process-wide override the engine test suite keeps set.
+fn ir_selection_available(ir_exists: bool, forced_off: bool) -> bool {
+    ir_exists && !forced_off
+}
+
 /// Highest-scoring detection: the face every pipeline stage keys on when a
 /// frame holds more than one.
 fn top_detection(faces: &[Detection]) -> Option<&Detection> {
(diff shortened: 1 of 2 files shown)


archledger:
The finding is accepted and fixed in a1bc515 — and it was sharper than stated: the first version of the new test only passed *because* of the cancellation (this suite keeps `IRLUME_FORCE_NO_IR=1` set process-wide, and the test read `/dev/null` retiering to Secure as the feature under test).

Adopted your shape: `irlume_camera::ir_forced_off()` with `capabilities()` reading it, and the selection rule extracted as the pure truth table `ir_exists && !forced_off`, since the positive arm is unreachable through a real engine under the suite's env. The engine-level test asserts the override survives both entry points with an existing IR path, and additionally discriminates the assignments themselves by pre-forcing the field true — the first mutant run showed a deleted `with_devices` write was invisible to every forced-off assertion, exactly the same-value-on-reachable-inputs trap. Four mutants killed: forced-off dropped, exists dropped, each entry point's write deleted.

PR #285, opened 2026-08-05
tui+cli: the model screen and help speak per-stage, not PAD-only

archledger:
The #280 review deferred this follow-up; buffalo enabled on a real machine made it urgent. The TUI's settings row read the PAD key only, so the primary interactive surface showed "none" while a replacement recognizer was selected.

What changed, after the review round:

- The daemon publishes its loaded recognizer. `Engine::with_thirdparty_recognizer` carries the name, `EngineBits` and `Response::Health` expose it (`#[serde(default)]`, so an older daemon deserializes as None), and the Settings row renders every loaded stage. Before the fix, a loaded PAD cue hid a loaded recognizer entirely.
- `TuiState` is structured. Each enabled entry carries its name, stage, and weight state, so the Repair check names exactly the model whose checksum failed and states the stage's real consequence: a refused PAD cue is off, a refused recognizer stops the daemon from starting. The previous joined string attributed one model's failure to every enabled model.
- The `[m]` toggle decides from enabled state, not installed files. Orphaned weights (the partial state `install_verified` itself admits) no longer produce a disable offer that `models disable` would refuse. With several stages enabled it lists each with its explicit disable command instead of guessing which authentication policy to remove.
- The section heading and help text stop calling every model a liveness cue.

Tests: a renderer regression drives both loaded stages through the draw harness (health constructed explicitly, because `test_app` leaves it None and an `if let Some` mutation silently skips the state under test); a sandboxed-config test drives recognizer-only and both-stages selections through `tui_state()`. Mutants killed: the renderer's recognizer half dropped, and `tui_state` truncated to its first entry. Workspace suite 1104 tests, clippy clean with `-D warnings`.

Not covered: the Repair rows have no test harness in this repo, so the per-stage wording arms are unexercised like the rest of that method, and the `[m]` handler is TUI event code without tests. The tested part is the state logic both consume.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -44,6 +44,10 @@ pub struct Engine {
     /// value (#276), because a threshold is a property of one model's cosine
     /// scale and applying another model's number to it is a guess.
     rgb_threshold: f32,
+    /// Name of the loaded third-party recognizer (None = shipped). Display
+    /// and Health reporting only; the POLICY lives in rgb_threshold and
+    /// ir_matching.
+    thirdparty_recognizer: Option<String>,
     /// Whether IR matching (fusion, IR fallback, calibrated centroid, dark
     /// IR-only auth) may run. False under a third-party recognizer: the IR
     /// thresholds and the fusion Platt calibration are measurements of the
@@ -611,6 +615,7 @@ impl Engine {
             ir_space: "raw".into(),
             embed_space,
             rgb_threshold: irlume_core::RGB_MATCH_THRESHOLD,
+            thirdparty_recognizer: None,
             ir_matching: true,
             mesh: None,
             blaze: None,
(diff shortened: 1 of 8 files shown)


archledger:
All three findings fixed in the pushed commit.

- PAD hides the recognizer (High): taken structurally as you proposed. The engine carries the recognizer name (`with_thirdparty_recognizer(threshold, name)`), `EngineBits` and `Response::Health` publish it with `#[serde(default)]` so older daemons deserialize as None, and the Settings row renders every loaded stage from health, falling back to per-entry filesystem state. Your renderer regression test is in, with one correction to my own first attempt at it: `test_app` leaves health `None`, so an `if let Some` mutation silently opted the test out. Health is constructed explicitly.
- Smeared checksum failure (Medium): `TuiState::Enabled` now carries typed entries (name, stage, weight state). The Repair check loops per entry, skips stages the daemon actually loaded, and words the consequence per stage: a refused PAD cue is off, a refused recognizer refuses daemon startup. The unknown-name case is its own variant instead of a fake Enabled.
- [m] decides from files (Medium): the handler now decides from enabled state; orphaned weights and root-only-unknown states get guidance rather than a disable offer `models disable` would refuse.

Mutants: dropping the renderer's recognizer half and truncating `tui_state` to its first entry each fail a named test. Not covered, stated plainly: the Repair rows have no test harness in the repo, so the per-stage wording arms are unexercised like the rest of that method; the `[m]` handler remains TUI event code without tests.

Workspace 1104 tests, clippy clean with `-D warnings`.

Issue #207, opened 2026-08-01
The packaged daemon cannot see cross-uid camera holders; decide CAP_SYS_PTRACE vs a confined observer


archledger:
PR #206 (issue #169) added a `/proc` scan: `ir_emitter::enable` stands down from the emitter write when another process holds the camera node. The scan has a production limit that the code now reports instead of hiding, and closing that limit needs a decision this issue exists to record.

## Mechanism

Reading another process's `/proc/<pid>/fd` targets is gated by `PTRACE_MODE_READ_FSCREDS`, not ordinary file permissions (proc_pid_fd(5), ptrace(2)). The packaged unit runs the daemon as root with `CapabilityBoundingSet=CAP_DAC_OVERRIDE CAP_FOWNER`, and neither capability satisfies the ptrace check. When PipeWire or a browser in the user's session holds `/dev/video2`, the daemon's scan gets a permission error on that process and cannot see the hold.

Since the fix on the #206 branch, the scan carries that fact (`ConsumerScan::permission_denied`), `enable` logs the degradation once per process and proceeds, and the threat model states the limit. The stand-down still fires for every consumer the daemon can see, which under the packaged capability set means same-privilege ones.

## The decision

Two ways to make the scan see the real case:

1. **Add `CAP_SYS_PTRACE` to the unit.** One line in `packaging/systemd/irlumed.service`. It grants the daemon read access to ptrace-gated proc files of every process on the machine, trips the systemd exposure ratchet, and widens what a compromised daemon can read.
2. **A separately confined observer.** A small helper holding only that capability (or another inspection interface), reporting consumer pids to the daemon over a narrow channel. No new capability on the authentication daemon itself, at the cost of another component to package, confine, and version.

Option 1 is a privilege expansion on an authentication daemon. Option 2 is more code and a new trust boundary to specify. Neither is obviously right, so the daemon currently proceeds with the degraded scan and says so in the journal.

## Acceptance for whichever option wins

- A `/dev/video2` holder under a non-root uid is detected by the scanner running under exactly the packaged unit's capability set.
- With the chosen mechanism removed, the scan reports itself incomplete rather than returning an empty successful result.

archledger:
Decided: keep the degraded scan and the report of its limit. No `CAP_SYS_PTRACE` on the daemon, and no separate observer for now.

[...]

PR #286, opened 2026-08-05
apparmor: the emitter locks and /dev discovery need rules the 0.7 soak predates

### LINKED ISSUE #283: archhost: daemon cannot take its emitter journal lock (/run/lock EACCES as root); startup warns it will not drive the emitter, yet IR capture works
Split out of #281's investigation notes so it survives that issue's close.

Every daemon start on archhost logs:

> irlume: could not check whether an interrupted emitter setup left this camera changed (lock the camera: open /run/lock/irlume-emitter-….lock: Permission denied (os error 13)). irlume will not write to this camera's emitter from this process until the lock error is resolved and the journal can be checked, so IR face authentication will not light here.

followed two seconds later by `IR emitter ready`, and IR capture demonstrably works (enrollment and authentication both captured lit IR frames on the same boot on 2026-08-05). So either the warning's stated consequence is not what actually happens — worth understanding on its own, because the message promises a fail-closed behaviour that something else is bypassing — or the emitter is being driven by a path that does not go through the lock, or it stays lit from a previous holder (the NexiGo's control survives process exit, per the measurement recorded in `IrCamera::session`).

What was established on 2026-08-05, all on the same boot:

- The failure reproduces under both the packaged 0.7.0 unit and the current repo unit.
- It is not the unit sandbox: a `systemd-run` unit with the same hardening properties creates and removes a file in `/run/lock` without error at idle.
- It is not filesystem permissions: `/run/lock` is root-owned 0755 and plain `sudo touch` works.
- The lock file named in the error does not exist afterwards, so the failing call is the `O_CREAT` open itself.
- The #207 consumer-scan warning (`could not inspect every process (permission denied)`) fires in the same startup second, suggesting a shared cause with a startup-timing character, the same family as the #281 capability race.

Not yet done: an `strace` of the daemon's startup on archhost to see the exact failing syscall and its context, and a check of whether the same warning appears on the Zenbook or ThinkPad.

archledger:
Fixes #283 and corrects the recorded mechanism of #281.

The strace on the enforcing Arch box settled it: the emitter-lock EACCES was never the unit sandbox or filesystem permissions. The `irlumed` AppArmor profile attaches by binary path (which is why running the bare binary reproduced it and a `/bin/sh` probe did not), and it predates three daemon needs from 0.8.x:

- `/run/lock/irlume-emitter-*.lock` open+flock: the #283 warning on every start.
- `/dev/` readdir: `capabilities()` saw no cameras inside the daemon while sysfs selection worked, so the engine mis-tiered. That is the deterministic mechanism behind #281; the "startup race" framing in that issue was wrong, and the #282 fix stands for a better reason than the one it was written under, because the engine must not depend on a probe the platform's MAC can scope differently from the daemon's own selection.
- `/var/lib/irlume/ir-emitter-stream/*.lock`: the state-dir `**` rule grants `rw` without the flock `k` permission. The first fix unmasked this one on the next restart.

Validation is the before/after on the enforcing machine: every daemon start logged the refusal warning and audit DENIED lines for both lock paths; after the profile reload, a restart produces zero warnings, the emitter reports ready, buffalo loads, and the only remaining denial is the #207 consumer scan's `sys_ptrace`, which that issue documents as an accepted fail-safe degradation and which this PR deliberately does not grant.

No code changes; there is no CI lane that exercises AppArmor, so the enforcing-machine measurement above is the validation.

### DIFF
--- a/packaging/apparmor/usr.bin.irlumed
+++ b/packaging/apparmor/usr.bin.irlumed
@@ -70,11 +70,13 @@ profile irlumed /usr/bin/irlumed flags=(attach_disconnected) {
 
   # The emitter undo-journal's per-camera lock (flock over the whole
   # check/act window; emitter_journal.rs). Lives under /run/lock because a
-  # lock's lifetime is a boot. 'k' is the flock permission; without these
-  # rules the daemon refuses to drive the IR emitter and logs the #283
+  # lock's lifetime is a boot. 'k' is the flock permission; without this
+  # rule the daemon refuses to drive the IR emitter and logs the #283
   # warning on every start. The profile's 2026-07-09 soak predates the
-  # journal (0.8.x), which is how the gap shipped.
-  /run/lock/ rw,
+  # journal (0.8.x), which is how the gap shipped. No rule on the /run/lock
+  # directory itself: creation is authorized on the new file's pathname, the
+  # recorded denials named only the file path, and enumerate/modify authority
+  # over every service's lock is nothing this daemon needs.
   /run/lock/irlume-emitter-*.lock rwk,
 
   # The per-stream emitter bookkeeping lock is a flock too, and it lives
--- a/packaging/apparmor/usr.bin.irlumed
+++ b/packaging/apparmor/usr.bin.irlumed
@@ -55,6 +55,11 @@ profile irlumed /usr/bin/irlumed flags=(attach_disconnected) {
 
   # Cameras: capture nodes + the sysfs identity walk of the anti-injection
   # camera pinning (canonicalized device path, USB ids, privacy switch).
+  # The /dev/ directory read is the discovery walk (discover_nodes lists
+  # /dev to find video nodes); without it, capabilities() sees no cameras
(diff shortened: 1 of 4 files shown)


archledger:
All three findings fixed in the pushed commit.

- usr.local.bin variant (High): carries the same three rules now, with its reload instruction corrected to point at itself. The parity script greps all three runtime rules in both variants exactly as you proposed; deleting a rule from either profile fails the script, proven by mutating one out and watching MISS plus exit 1.
- /run/lock/ rw (Medium): deleted from the one profile that had it, on your criterion exactly. The recorded denials named only the file path, and the A/B ran on the enforcing machine: the trimmed profile parses, the daemon restarts with zero warnings, and the only remaining audit line is #207's sys_ptrace. Both lock rules stay file-scoped with 'k'.
- nfpm comment (Low): now says the profile enforces by default.

Both variants parse under apparmor_parser on the enforcing box (the local variant with -Q, since only the packaged attachment path runs there).

PR #287, opened 2026-08-05
auth: the dark-login shape falls through the Uncertain short-circuit

### LINKED ISSUE #284: Dark IR-only login appears unreachable: the top-level Uncertain early-return intercepts 'no face in RGB' before the dark path runs
Observed 2026-08-05 on archhost (c0fd862, debug tracing on), in a genuinely dark room with the screen dimmed:

```
assess: rgb 640x480 in 1868ms, faces=0 top-det=0.00
assess: ir 640x360 in 1806ms, faces=1 top-det=0.92
assess: RGB has no face but IR does; recapturing RGB alone (dim overlapped frame?)
assess: rgb (recaptured) 640x480, faces=0 top-det=0.00
liveness(cross-spectrum): Uncertain (no face in RGB; present your face); ir_bright=180 … glint=253.00 ambient=0
face auth 'archledger': granted=false live=false score=0.0000 (liveness Uncertain: no face in RGB; present your face)
```

The emitter lit, IR found the face at 0.92, and the room was dark — the exact condition the dark IR-only path exists for. But `authenticate_once` returns on `a.verdict == Verdict::Uncertain` before either matching path, and the cross-spectrum gate reports Uncertain whenever RGB has no face. The dark branch (`if let Some(probe) = a.ir_embedding`, which re-derives its own verdict via `evaluate_ir_only`) is therefore unreachable in the situation it was written for.

The early return was added in the #238 round so an unreadable frame is reported before anything derived from it, and its comment scopes the intent to exposure problems ("a blown frame that hides the corneal glints"). It catches all Uncertain reasons, including the no-RGB-face one that must fall through to the dark path.

Two follow-on facts:

- The third-party recognizer's explicit dark refusal ("dark unlock unavailable…", #279) sits inside the dark branch and is currently dead code behind this. Under a third-party recognizer the OUTCOME in the dark is still the designed one — deny, password — but via the generic Uncertain reason rather than the honest one.
- Grants logged as "ir-fallback, dim light" are the RGB-path fallback (RGB face found but dim) and are unaffected; this is specifically about RGB finding no face at all.

Deciding the fix needs care: the Uncertain early-return exists for a real reason (#238's retryability semantics), so the likely shape is distinguishing "Uncertain because RGB is unreadable/blown" from "no RGB face, IR face present" and letting only the latter continue into the dark branch — plus a dark-room capture to prove a dark grant works again on the shipped stack before touching the third-party behaviour. Validation requires a person in a dark room, so it pairs with the next capture session.

archledger:
Fixes #284. The full evidence is on the issue: in a genuinely dark room the emitter lit, IR found the face at 0.92, and authentication denied "no face in RGB" at the top-level Uncertain early-return, before the dark IR-only branch ever ran. That early-return arrived in the #238 round for unreadable frames; it caught every Uncertain reason, so Windows-Hello-style dark login has been unreachable since 2026-08-02, and the third-party recognizer's explicit dark refusal from #279 was dead code behind it.

The fix keys the short-circuit on the structural shape instead of the verdict alone. `uncertain_short_circuits(verdict, has_rgb_embedding, has_ir_embedding)` denies every Uncertain except the dark-login shape (no RGB embedding, IR embedding present), which falls through. The safety argument for the fall-through: the RGB branch is skipped because there is no embedding, and the dark branch derives its own verdict via `evaluate_ir_only`, which shares the exposure refusal, so a blown IR frame in the dark still denies there, with the dark path's own retryability kinds. The #238 case (RGB face present, frame unreadable), the no-face-anywhere denial, and every non-Uncertain flow are byte-for-byte unchanged.

Tested: a truth table pins all six shapes, and three condition mutants (blanket short-circuit restored, never short-circuits, shape inverted) each fail it. Workspace 1105 tests, clippy clean with `-D warnings`.

Not covered yet, and deliberately gating the merge: the two newly reachable live outcomes need a dark room on hardware. Planned on archhost: with buffalo enabled, a dark attempt must produce the explicit "dark unlock unavailable with a third-party recognizer" refusal; and on the shipped stack, a dark attempt must produce an actual dark grant. This PR waits for those two observations.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -572,6 +572,25 @@ fn ir_selection_available(ir_exists: bool, forced_off: bool) -> bool {
     ir_exists && !forced_off
 }
 
+/// Should a top-level Uncertain verdict deny before either matching path?
+///
+/// Yes for every Uncertain EXCEPT the dark-login shape: no RGB embedding
+/// while an IR embedding exists. The cross-spectrum gate cannot say Live
+/// without an RGB face, so that shape always arrives as Uncertain, and
+/// short-circuiting it made the dark IR-only path unreachable in the exact
+/// condition it exists for (#284; observed live 2026-08-05: rgb faces=0, ir
+/// faces=1 at 0.92, emitter lit, denied "no face in RGB"). The dark branch
+/// re-derives its own verdict via evaluate_ir_only, so nothing is granted on
+/// the strength of the Uncertain that fell through.
+fn uncertain_short_circuits(
+    verdict: Verdict,
+    has_rgb_embedding: bool,
+    has_ir_embedding: bool,
+) -> bool {
+    let dark_login_shape = has_ir_embedding && !has_rgb_embedding;
+    verdict == Verdict::Uncertain && !dark_login_shape
+}
+
 /// Highest-scoring detection: the face every pipeline stage keys on when a
 /// frame holds more than one.
 fn top_detection(faces: &[Detection]) -> Option<&Detection> {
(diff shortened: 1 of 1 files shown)


archledger:
Both hardware gates passed on archhost (NexiGo N930W, this branch built at dc4fc5c, debug tracing on, dark room with the monitor powered off).

Gate 1, third-party recognizer (buffalo enabled). Three attempts, all identical: `assess: rgb faces=0` even after the recapture retry, `assess: ir faces=1 top-det=0.91`, and the request reached the dark branch to produce the explicit refusal that had been dead code since #279:

```
face auth 'archledger': granted=false live=false score=0.0000
(dark unlock unavailable with a third-party recognizer (its IR matching is unmeasured);
 use the password, or switch back to the shipped recognizer)
```

Gate 2, shipped recognizer. Six attempts, all granted through the dark path:

```
assess: rgb 640x480, faces=0 top-det=0.00
assess: ir 640x360, faces=1 top-det=0.93
assess: rgb (recaptured) 640x480, faces=0 top-det=0.00
liveness(ir-only/dark): Live (live (dark/IR-only): IR-reflective, 3D)
gate(per-user IR center/edge floor, dark): live 1.49 vs floor 0.97
match(ir/dark): best 0.888 vs thr 0.600 (10 scans, adapter=false, calib_centroid=Some(0.8961067))
face auth 'archledger': granted=true live=true score=0.880 (match: BEN (ir/dark))
```

That is the full newly-reachable sequence: the cross-spectrum gate still reports Uncertain, the dark branch derives its own Live verdict independently, the per-user IR floor is applied, and the IR match decides. Scores ranged 0.816 to 0.888 across the six.

One honest note on process: an earlier attempt at gate 2 granted four times through the RGB path (rgb faces=1 at 0.92, `match: BEN (rgb)` at 0.65 to 0.73), because the room was not actually dark. Those runs prove nothing about the dark path and are not counted; the gate was re-run with the monitor physically powered off.

Issue #288, opened 2026-08-05
Per-model enrollment: switching recognizers should not require re-enrollment (and ir_calib is keyed wrong for it)


archledger:
Switching recognizers should not cost a re-enrollment. Most of this already works, one part of it is a latent correctness bug, and the rest is management and display.

## What already works

Every scan carries `embed_space`, the digest of the weights that produced it (#277), and matching filters to the loaded recognizer's space. So scans from two recognizers coexist in one enrollment file today, and switching models silently changes which ones are live.

Observed on archhost, 2026-08-05, with no code changes: an enrollment holding `BEN` (10 scans, shipped recognizer) and `Face Profile 1` (5 scans, buffalo). With buffalo enabled, only the buffalo scans matched. `sudo irlume models disable buffalo` restarted the daemon on the shipped recognizer, and BEN authenticated immediately, in the dark, at 0.816 to 0.888. No re-enrollment, no data loss.

## The bug that has to be fixed first

`FaceProfile.ir_calib` is a single value, and `refit_profile_calib` fits it from scans in the loaded recognizer's space and then overwrites the field unconditionally. In a profile holding scans from two recognizers, adding scans under model B replaces the calibration belonging to model A; switching back to A then applies B's calibration to A's templates inside `ir_match_in`. That is wrong numbers reaching the matcher, not a display problem.

Fix: key the calibration by embedding space, migrating an existing value under `LEGACY_RECOGNIZER_SPACE`. This should land before multi-model profiles are encouraged anywhere in the UI.

## Design

**Keep the model dimension on the scan.** It is already there. A container layer would restate the same fact in a second place, which is the failure pattern behind both #281 (the engine's own capability probe disagreeing with the daemon's selection) and the first cut of #285 (Health's PAD-only field disagreeing with the state the TUI computed).

**Identity across models is asserted by the operator, never inferred.** irlume cannot determine that BEN and Face Profile 1 are the same person; comparing their vectors across spaces is the meaningless operation the recognizer-identity work exists to prevent. The flow is `irlume enroll --profile BEN` with the new model loaded, appending scans in that model's space to the named profile. The existing collision guard keeps working within the loaded space.

Security note to document rather than discover later: that guard stops a second person's face being added to a profile, and it can only do so within one embedding space. Under operator-asserted linkage, a different person's face added under a different model cannot be caught automatically. Liveness gates still apply and enrollment is root, so the exposure is bounded, but the docs must say the linkage rests on the operator's assertion.

[...]

PR #289, opened 2026-08-05
core+auth: calibrations are keyed by recognizer, not one slot per profile

archledger:
Step 1 of #288, and a correctness fix on its own merits.

`FaceProfile` held a single `ir_calib`, and `refit_profile_calib` fitted it from scans in the loaded recognizer's space and then overwrote the field. A profile holding scans from two recognizers is already possible, since templates started carrying their producer's digest in #277 and archhost demonstrated the two-model state last night. In that state, adding scans under model B replaced the calibration belonging to model A; switching back to A applied B's calibration to A's templates inside `ir_match_in`, which puts uninterpretable numbers into the matcher rather than failing.

`ir_calibs` keys calibrations by embedding space. `calib_for` reads the map and falls back to the legacy slot for the shipped recognizer, so a profile written before this change keeps its calibration. `set_calib_for` records one model's value without touching another's, and mirrors the shipped recognizer's value back into the legacy slot so an older irlume reading the same file still finds it. The on-disk shape is additive in both directions, matching how `embed_space` and `ir_space` were added.

Tests: a storage truth table over legacy-only reads, cross-model isolation, the mirror, and per-model clearing; a deserialization test that loads a pre-change enrollment; an engine test running the switching scenario, refitting under a second recognizer and asserting the first survives; and a matcher test proving the calibration is chosen by space, discriminated by presence, because a loaded recognizer with no calibration of its own must skip the calibrated-centroid protocol that another model's calibration would run. The existing engine assertions on the legacy field now also exercise the mirror.

Five mutants, each failing a named test: clearing other models on set, ignoring the space on read, dropping the legacy fallback, dropping the mirror, and the matcher reaching for any calibration on file. That last one survived the first pass and is why the matcher test exists.

Workspace 1109 tests, clippy clean with `-D warnings`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ __pycache__/
 # Mutation-harness backups. One was committed by a `git add -A` run while the
 # harness had files out, which put a 1938-line copy of a source file in a commit.
 *.rs.bak
+.aider*
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -3800,6 +3800,39 @@ mod tests {
     }
 
     #[test]
+    fn the_matcher_picks_the_calibration_by_recognizer_space() {
+        // #288: a profile can hold calibrations for several recognizers, and
+        // ir_match_in must apply the LOADED one. Discriminated by presence:
+        // with the calibration filed under a different space, the loaded
+        // recognizer has none, so the calibrated-centroid protocol does not
+        // run at all. Reaching for any available calibration instead of the
+        // keyed one puts another model's transform on these templates.
+        let (mut prof, probe) = calibrated_profile(16);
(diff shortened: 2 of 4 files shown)


archledger:
Accepted and fixed in the pushed commit. The finding was exactly right: `n_templates` is counted before the calibration lookup, so it proves only the space filter, and my tests covered rejecting another recognizer's calibration without ever covering applying the loaded one's. Your proposed mutant would have passed the entire suite.

The switching test now asserts `centroid.is_some()` under model B and under the shipped recognizer after switching back, which is observable only once a calibration has been retrieved and applied. I ran your mutant verbatim against the suite: it fails `a_refit_under_one_recognizer_leaves_another_models_calibration_alone`. The inaccurate comment about the restored engine is gone too.

Workspace 1109 tests, clippy clean with `-D warnings`.

PR #290, opened 2026-08-05
auth+cli: scan limits count per recognizer, and add-scan takes a count

archledger:
Step 2 of #288. Adding a second recognizer's templates to an existing profile already worked in principle: each scan carries its producer's digest (#277), and the anti-mixing check filters by space, so the operator naming a profile is the identity link. Two things stopped it being usable.

**The scan limit counted every scan in the profile, regardless of recognizer.** A profile that had reached the limit under one model could never gain a template for another, which is the case this feature exists for. `FaceProfile::scans_in` counts per space and both limit sites use it. The rationale is the limit's own: it bounds false-accept inflation from taking the best of N templates, and a comparison only ever ranges over one embedding space, so ten scans under each of two recognizers is two independent best-of-ten operations rather than one best-of-twenty.

**`add-scan` captured exactly one scan.** It now takes a count, so a second model's templates arrive in one session. The request field is optional and defaults to the previous behaviour, so an older CLI keeps working against a newer daemon. The anti-mixing check now runs on every capture rather than only the first, matching what enrollment already does, so a second person stepping into frame partway through is caught. The CLI states plainly that scans are recorded for the recognizer the daemon has loaded, and that this is how a profile gains a second model's templates without re-enrolling as a new person.

## Testing

Per-space counting over a mixed profile; a profile full of another recognizer's scans reaching the capture boundary instead of refusing, which is the regression this step exists to prevent; and the capture-collision decision extracted as a value so the every-capture rule can be observed at all, since the loop it guards sits behind a camera.

Four mutants, each failing a named test: `scans_in` ignoring the space, the limit counting all scans, the collision check narrowed to the first capture, and the earlier keying mutants still covered. Workspace 1111 tests, clippy clean with `-D warnings`.

## Not covered

No hardware run: a real second-model add-scan session needs a camera and a second installed recognizer, and archhost currently has buffalo disabled. The pieces it would exercise are the capture loop and the calibration refit, both of which have unit coverage; the end-to-end version belongs with the next hardware session.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -2888,11 +2888,17 @@ impl Engine {
 
     /// Add one scan to an existing profile ("improve recognition"). Errors if the
     /// profile is missing or already at MAX_SCANS_PER_PROFILE.
+    /// Add `count` scans to an existing profile, in the LOADED recognizer's
+    /// space. This is also how a profile gains templates for a second
+    /// recognizer without re-enrolling as a new person: the operator names
+    /// the profile, which is the only way the link can be made, since
+    /// comparing vectors across embedding spaces is meaningless (#288).
     pub fn add_scan(
         &mut self,
         user: &str,
         profile_name: &str,
-    ) -> irlume_common::Result<(String, usize)> {
+        count: usize,
+    ) -> irlume_common::Result<(Vec<String>, usize)> {
         use irlume_core::storage::{self, FaceScan, MAX_SCANS_PER_PROFILE};
         let mut enr = storage::load(user)?
             .ok_or_else(|| irlume_common::Error::Protocol(format!("'{user}' is not enrolled")))?;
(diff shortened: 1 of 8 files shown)


archledger:
All four findings fixed, in ee4c17d and the commit before it.

- **Merge-path underflow (High):** correct, and it was a bug this branch introduced. Permitting the limit per recognizer means one profile may legally hold more scans in total than the limit, and both enrollment merge sites still subtracted the total. My PR body's claim that "both limit sites use" per-space counting was false; only `add_scan` did. Every site now goes through one saturating `scan_room_in`, so a later site cannot inherit the old arithmetic, and the fixture in the new test holds more than the limit in total while staying under it in the loaded space, which is exactly the underflow case.
- **Short capture reported as success (Medium):** refused before anything is written, matching enrollment. The decision is extracted as a value because the capture sits behind a camera; a mutant relaxing it back to "only zero is short" survived until that extraction, and now fails.
- **Malformed or zero `--scans` (Medium):** now a usage error, exit 2, before any daemon request. CLI tests cover `0`, `-1`, `abc` and empty, and a valid count building a request.
- **Docs and comments (Low):** `COMMANDS.md`, the request doc, and the engine doc comment no longer describe a one-scan command.

One process note worth stating: the fmt lane failed on my previous commit because I formatted before adding the CLI test and did not re-check. Two entries went into my own defect list from this round: a formatting check proves nothing about code written after it ran, and a limit split by a new dimension underflows at every site that did not learn about it.

Workspace 1114 tests, clippy clean with `-D warnings`, `cargo fmt --all --check` clean.

PR #291, opened 2026-08-05
profiles: report which recognizer each scan belongs to and which is live

archledger:
Step 3 of #288, the part that makes the state visible.

A profile can hold templates from several recognizers at once, since each scan records the model that produced it, and only the loaded recognizer's templates can match. The flat scan count cannot say whether a profile is usable right now: after switching models, a profile with ten scans and none for the loaded recognizer looks exactly like one that works. That is the state archhost was in last night, and nothing on screen said so.

`ProfileSummary` now carries per-recognizer counts and the loaded space. Both fields are additive with serde defaults, so an older daemon reports nothing rather than breaking a newer CLI.

**Human listing** prints the breakdown only when it says something the total does not: more than one recognizer, or a loaded recognizer that does not own every scan. When none of a profile's scans belong to the loaded recognizer it names the `add-scan` command, which is the fix for that state.

**Machine API** gains a `recognizers` array per profile, each entry carrying the space, the scan count, and whether it is the loaded one. Schema, fixture, and MACHINE-API.md updated; conformance passes 36/36 in strict mode.

## Testing

A daemon test pins the summary mapping, including untagged scans counting under the recognizer that predates tagging, which is the rule matching already applies. A machine-API test pins the live flag per recognizer, that exactly one is live, and that an older daemon's absent fields yield an empty array rather than a fabricated one.

Three mutants, each failing a named test: untagged scans dropped from the counts, the loaded recognizer not reported, and every recognizer marked live. That last one survived the first pass, which is why the machine-API test exists.

Workspace 1116 tests, clippy clean with `-D warnings`, `cargo fmt --all --check` clean.

## Not covered

The TUI's profile screen still shows a flat scan count. It is the next surface to teach, and it belongs with the `forget-model` step rather than here, since both touch the same screen.


archledger:
Both findings fixed in c0aa6b3, plus the cause of a CI failure this branch kept hitting.

- **Shell injection (Medium):** correct and worth catching. `shell_single_quote` escapes the name the only way a POSIX shell accepts inside single quotes, and the regression test confirms the result by asking `sh` what it parses back, rather than by comparing my own escaping to my own expectation of it. Your exact hostile name is the fixture.
- **Compatibility half never touched the wire (Low):** also correct, and the same class as a finding two PRs ago: the literal supplied what the test claimed serde was supplying. It now decodes `{"name":"P","scans":["s1"]}` and asserts both fields default before passing it on.

Separately, while this PR was in review the archhost lane failed for the third time on `update_probes_query_the_package_manager_and_the_ppa_index`. It was not flakiness. `gz_lists_irlume` discarded the result of writing the index to gzip's stdin and returned a bare bool, so a failed or partial write on a loaded machine read as "the package is absent", which in production tells a current-LTS user that their own archive does not serve them. It returns `Option` now: a write failure or a nonzero gzip means "cannot tell", which `ppa_serves_via` already distinguishes from a 404. The undecompressable-input case in its test changed from `false` to `None` accordingly.

Workspace 1117 tests, clippy clean with `-D warnings`, fmt clean.

archledger:
Correction on the CI failures, since I got this wrong twice before landing on it.

The archhost lane failed again on c0aa6b3, this time with two tests down: the update probe as before, and the new shell-quoting test failing to spawn `sh` at all. Both tests pass when run alone on that machine; the full suite reproduced the failure three runs out of three. That pattern is shared-state interference, not a loaded runner.

The cause is mine, from #285: the test that fakes `systemctl` sets `PATH` to a temp directory holding the fake. Rust runs tests in parallel threads of one process, so for the duration of that test every concurrent test loses `sh`, `gzip`, and everything else on the real path. `ppa_serves_via` then cannot run its fake and returns None where the fixture expects Some(false), which is exactly the failure I had twice attributed elsewhere. It now prepends instead of replacing, so the fake `systemctl` still wins and every other binary stays resolvable. Five consecutive full-suite runs are clean locally.

The tri-state `gz_lists_irlume` from the previous commit stands on its own merits, since a failed write to gzip genuinely is not "the package is absent", but I was wrong to present it as the cause of these failures.

PR #292, opened 2026-08-05
cli+daemon+tui: profiles forget-model, per-model profile view, disable note (#288 step 4)

archledger:
Step 4 of #288, the last of that design.

## What this adds

**`irlume profiles forget-model <model>`.** `models disable` deletes a recognizer's weights and deliberately keeps its templates, so re-enabling needs no re-enrollment. This command is the counterpart for when the operator wants that biometric material gone: it removes every scan a user holds in one recognizer's embedding space, plus the per-profile calibrations fitted from those scans. A profile left with no scans is deleted with them, and deleting the last profile erases the enrollment file; the same never-orphaned rule `DeleteScan` upholds, via the same `mutate_enrollment` path.

The CLI resolves the model name (`shipped`, a catalog name, or a raw `embed:<sha256>` tag as `profiles list` prints it) and the daemon only ever sees the space tag, so a recognizer the catalog no longer carries stays forgettable. A catalog entry from another stage is refused by name: detection, landmarks, and the PAD cue store no enrollment data. Untagged pre-#277 scans are matched through `recognizer_space_matches`, so forgetting `shipped` covers them. When the forgotten space is the one the daemon has loaded, the reply says so and names the fix.

**The TUI stops reporting misleading flat scan counts.** Only the loaded recognizer's scans can match. The profile screen now shows "(N scans, M for the loaded recognizer)" when the two differ, with a warning line naming Improve Recognition when M is 0; the Welcome and Done enrollment badges warn instead of showing green in the same case. An old daemon reports neither field and keeps the flat count, since nothing can be said. Wording follows the CLI listing from #291.

**`models disable` on a recognizer now prints the asymmetry**: scans are kept, re-enabling needs no re-enrollment, and `profiles forget-model` erases them.

## Testing

- `cargo test --workspace` green; clippy clean; fmt checked at commit time.
- The two save-path daemon tests self-skip on TPM hosts, so they were also run under `unshare -Umr` with tmpfs over `/dev` (the repo's no-TPM reproduction from #226): all three forget tests execute their real paths there, including the erase-the-file case.
- 8 hand-run mutants, each killed by the test written for it, applied and reverted by inverse patch: CLI stage guard inverted, TUI breakdown condition inverted, badge warn arm disabled, daemon no-data error disabled, scan retain filter disabled, calibration clearing dropped (kill confirmed on both the counting and the clearing half), emptied-profile retain disabled, loaded-recognizer note disabled.
- Wire test: a fake daemon asserts the request carries the resolved `embed:<pin>` for a catalog name, so a resolver regression fails in CI rather than on an enrollment.

## Not tested

- No hardware run yet. archhost holds the two-model fixture (`BEN` 10 shipped-space scans, `Face Profile 1` 5 buffalo-space scans); I will validate `forget-model buffalo`, the TUI view against the mixed enrollment, and the `models disable` note there before merge, per the container-first protocol.
- An old daemon meeting a new CLI answers `bad request` to `ForgetRecognizer` (unknown variant), the same behavior every added request variant has during an upgrade window; not exercised by a test.

### DIFF
--- a/crates/irlume-cli/src/main.rs
+++ b/crates/irlume-cli/src/main.rs
@@ -327,6 +327,18 @@ fn shell_single_quote(value: &str) -> String {
 /// to-3 face profiles and their scans via the daemon.
 fn profiles(sub: Option<&str>, args: &[String]) -> std::process::ExitCode {
     use irlume_common::{Request, Response};
+    // A supplied `--user` must carry a real username. `user_arg` falls back
+    // to SUDO_USER/$USER when the flag is ABSENT, which is right for the bare
+    // commands, but a dangling `--user` inheriting that fallback would point
+    // a destructive subcommand (forget-model, delete) at the invoking user's
+    // own enrollment. Checked here, not in `user_arg`: the fallback semantics
+    // of an absent flag belong to every caller, this omission does not.
+    if args.iter().any(|a| a == "--user")
+        && !matches!(flag(args, "--user"), Some(u) if !u.is_empty() && !u.starts_with("--"))
+    {
+        eprintln!("[profiles] --user requires a username");
+        return std::process::ExitCode::from(2);
+    }
     let user = user_arg(args);
     let req = match sub {
         None | Some("list") => Request::ListProfiles {
--- a/crates/irlume-cli/src/main.rs
+++ b/crates/irlume-cli/src/main.rs
@@ -367,6 +367,25 @@ fn profiles(sub: Option<&str>, args: &[String]) -> std::process::ExitCode {
             }
             None => return usage_profiles(),
         },
+        Some("forget-model") => {
+            // Positional: `irlume profiles forget-model <model>` (args[0] is
+            // "profiles", args[1] the subcommand). A flag must not be read as
+            // the model name when the positional is missing.
+            match args
+                .get(2)
+                .map(String::as_str)
+                .filter(|a| !a.starts_with("--"))
+            {
+                Some(name) => match crate::models::recognizer_space_for(name) {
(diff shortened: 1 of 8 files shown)


archledger:
Codex round (one round per the merge policy) raised two findings, both real, both fixed in fe808bd.

**Merge-blocking, confirmed: a dangling `--user` targeted the invoking user.** `flag()` returns nothing when `--user` is the last argument, and `user_arg` then falls back to `SUDO_USER`/`$USER`, so `sudo irlume profiles forget-model shipped --user` would have deleted the invoking user's own shipped-space enrollment. Every `profiles` subcommand now refuses a supplied `--user` that carries no username (or a flag where the username should be) with exit 2 before any request is built. The guard sits in `profiles()`, not `user_arg`, so the absent-flag fallback other commands rely on is unchanged. Proven test-first: `a_dangling_user_flag_never_reaches_the_daemon` fails on ec5bdec by reaching the dead socket, passes on fe808bd.

**Follow-up, taken now rather than later: uppercase raw tags could never match.** `recognizer_space_for` accepted `embed:` with uppercase hex and passed the case through, while stored tags are lowercase `sha256_hex` output compared case-sensitively; fail-closed, but a documented input form that always answers "no enrollment data" is a trap. Tags now lowercase on resolution, with the round-trip pinned in the resolver test.

The reviewer could not run the suite (read-only workspace at the parent commit); on fe808bd: `cargo test --workspace` green, clippy clean, `cargo fmt --all --check` clean, and the two no-TPM save-path tests re-verified under `unshare -Umr` with tmpfs over `/dev`.

PR #293, opened 2026-08-05
vision+auth: landmark failure modes measured, and garbage geometry now abstains (#276 stage 3)

archledger:
The measurement #276 stage 3 requires before detection/landmarks can open to third-party models, plus the guards it showed were missing. A broken or mis-converted model there does not raise an error; it emits coordinates, and this PR measures what every consumer answers when those coordinates are garbage, then makes the garbage abstain.

## The instrument

`examples/landmark_failure_probe.rs`: synthetic frames, hand-built pathologies (swapped labeling, background-locked, off-frame, far, collapsed, NaN, tiny, mis-indexed 468 topology), no camera, deterministic. Full before/after tables and analysis in `docs/pad-results/2026-08-05-landmark-failure-modes.md`.

## What it found on main, worst first

- Rust's saturating float-to-int cast turns a NaN landmark into pixel (0,0). With a bright frame corner (emitter bloom is a realistic stand-in), `eye_glint` scored 255 and `both_eyes_open`, the gate behind require-eyes-open, answered **true** from landmarks that do not exist.
- `align_to_arcface` turned NaN geometry into a NaN transform and returned an all-black chip as `Ok`; the embedder then embeds a chip that came from no face.
- `FaceMesh::landmarks()` ran on any box: zero-area, inverted, off-frame, NaN, and million-pixel boxes all returned a full 478-point set as `Ok`, including 478 NaN points.
- A NaN detector coordinate with a finite score survived YuNet decode and NMS.
- Measured fail-closed already, recorded, unchanged: off-frame and far coordinates sample nothing; the stream consumers (blink, deliberate closure with a valid calibration, nod, EAR calibration) refuse NaN and stuck-constant windows; swapped left/right labeling changes nothing.

## The guards

- `detection_is_finite`, applied in YuNet's `detect()` and to the Blaze rescue box: non-finite detections drop at the source rather than being guarded at every consumer.
- `mesh_box_valid`: the mesh refuses non-finite, non-positive-area, no-frame-overlap, and larger-than-4x-frame boxes by named reason.
- `mesh_output_plausible`, applied inside `map_checked_mesh_output`, the same call that maps raw output into frame coordinates, so a refactor cannot skip the check without losing the result.
- The glint cues and `both_eyes_open` fail closed on non-finite eyes; `align_to_arcface` refuses non-finite landmarks.

All are validity bounds, not tuned thresholds; the 25% border slop keeps a tilted chin from being refused, and every caller of `landmarks()` already treats an error as "no landmarks", the fail-closed answer.

## Testing

- Unit tests assert each refusal REASON, not just rejection, plus the benign cases (a partly out-of-frame box, border overshoot inside the slop) so the gates cannot creep.
- A model-backed test drives the real shipped mesh through `landmarks()` with garbage boxes, pinning the wiring, not just the pure functions.
- 13 hand-run mutants, each killed by its named test, applied and reverted by inverse patch: every branch of both mesh gates, both wirings, the align guard, `detection_is_finite`, and the three cue guards. One first-draft test could not kill its mutant (a uniform corner block reads contrast 0.0 with or without the guard); the test now uses a spike over darker neighbors and the mutant dies.
- `cargo test --workspace` green, clippy clean, fmt checked at commit.

## Not tested

- No live capture: the findings are about arithmetic (casts, fits, mapping), which synthetic frames exercise fully; nothing here claims sensor behavior.
- The detector-side finite filters cannot be exercised through the real models (the shipped ones never emit non-finite output); their decision function is unit-tested and the wiring is a one-line retain/check.
- `head_pose` still answers frontal on degenerate-but-finite geometry; recorded in the doc with the reasoning for leaving it.

Stage 3 remains closed after this: the measurement and guards are the prerequisite, and opening still needs a measured candidate entry (detection rate and landmark accuracy on this project's cameras, live).


archledger:
Codex round (one, per policy) raised four findings. All four were real, all four are fixed in 7e74d49; the High one was a false claim of mine that the review caught exactly right.

**High, confirmed: mesh refusals aborted authentication windows.** I wrote "every caller already treats a mesh error as no landmarks" having checked the refine path (`.ok()?`) and not the two streaming callers, which propagated with `?`. That operator was dormant while the mesh could only fail on ONNX runtime errors; the new validity refusals made it live, so one refused frame lost the whole passive-liveness capture or consent watch instead of one observation. Fixed structurally: the EAR pipelines now consume refusals through `mesh_min_ear`, which returns an `Option`, so a caller has no error to propagate; the model-backed test drives it with refused boxes and asserts `None`, and a nominal box still measures.

**Medium, confirmed: one stray point defeated the collapse gate.** The review's exact example (477 stuck points plus one 3px away) passed the extrema-based extent. The gate now judges the central 80% span, the one-outlier case is a unit test, and the model-backed test asserts the real mesh's central span sits far above the 2px floor even on a faceless crop, so the bound cannot cost genuine capture.

**Medium, accepted as a claim correction: the model-backed wiring test pins the box gate only.** The output gate cannot be exercised through the real mesh (it never emits implausible output); its wiring is structural instead, since the check lives inside `map_checked_mesh_output`, the only mapping implementation, and skipping it means reimplementing the mapping. The test comment and this PR's description now say exactly that instead of the overclaim.

**Low, taken now: the glint helpers let a bright valid eye vouch for a set with a NaN eye.** Both now score the whole set 0.0 on any non-finite eye, and the test stages the case the per-eye skip passed: a placeable eye on a bright glint plus a NaN second eye, asserting both helpers read 0.0 where the unguarded value is provably nonzero.

Four new hand-run mutants, each killed by its named test (refusal-swallowing in `mesh_min_ear`, the central-span check, both whole-set guards). Workspace green, clippy clean, fmt checked at commit. The measurement record carries a review-round section with all of this.

PR #294, opened 2026-08-05
bench+docs: live stage-3 detection/landmarks measurement on both cameras (#276)

archledger:
The live half of the #276 stage-3 measurement: fresh captures of the enrolled user on both cameras, benched offline against the shipped YuNet, the shipped BlazeFace short-range (the only converted third-party detection candidate), and the 478-point mesh. Measurement doc with full tables: `docs/pad-results/2026-08-05-stage3-live-detection-bench.md`; raw per-frame CSVs committed beside it.

## Headline numbers

- YuNet: 100% detection on RGB in every segment on both cameras (scores 0.91 to 0.95); on lit IR it misses only the 30cm partial-face regime (75 to 92%) and sweep extremes (88%).
- BlazeFace short-range: 100% within ~50cm, and **0 of 9 lit IR frames at 1m on the NexiGo** where YuNet holds 100%. Promoted to primary it would break far-distance dark login; its shipped rescue role is unchanged and now has a measured distance boundary.
- The #293 geometry gates refused **none** of the 196 genuinely detected frames; central spans measured 50 to 206px against the 2px validity floor.
- Mesh EAR stability on held poses: sd 0.006 to 0.040, glasses on or off.

## Decision recorded

The detection stage stays closed, now on measurement rather than the absence of one: the candidate loses the login envelope, and #276 allowed declining as an outcome. The landmarks stage has no distinct candidate to open for (the measured Face Landmarker mesh is the shipped artifact byte for byte; the Holistic alternative failed the clean-BOM bar). Full-range BlazeFace was not converted: its design range sits outside the login envelope and no measured gap called for it.

## Instrument honesty

Two defects in my own bench were found and fixed during the run, and the doc records both because the first numbers were wrong: a strict UTF-8 parse of the PPM header prefix silently dropped most bright RGB frames (dark IR corners parsed by luck), and a fixed lit-frame brightness floor mislabeled bright-ambient bursts, halving two segments' apparent detection rate. The committed instrument carries the fixes; the tables come from the corrected run.

## Not tested

- No dim/dark captures (the room could not be darkened): low-light detection floors are not measured here.
- One subject, one session per camera; no occlusion beyond glasses, no saturation (the 2026-07-15 bursts cover that).
- Landmark accuracy against ground truth is not measured, only stability and geometric plausibility.


archledger:
Codex round (one, per policy) raised four findings; all four were real, all addressed in c9b48c1, and the third changed the decision.

**Critical, fixed: inference errors were recorded as misses.** Detector execution errors now abort the run with the frame named, mesh errors are logged per frame, and the rerun over the same corpus produced zero errors with byte-identical CSVs, so the published rates carry no hidden runtime failures.

**High, fixed: model-byte binding.** The bench now refuses to run on anything but the shipped artifacts, checked against `models/SHA256SUMS`, and prints the digests; all three are recorded in the doc.

**High, accepted, and the review was right on the substance: the full-range dismissal was unsupported.** The first draft's "several meters, outside the login envelope" claim came from a card nobody had read; the card actually says in-scope to 5 meters. So it got measured, through Google's own runtime to keep hand-rolled decode out of the answer (mediapipe 1.0.0, floor 0.5, pinned artifact, sha256 recorded, card read: Apache-2.0, consented first-party data): **full-range detected 100% of every segment on both cameras, including all the far-IR frames short-range misses at 0%.** The decision section now names full-range as the detection-stage candidate and lists what opening requires (pinned conversion, a parity-tested 160x192 decoder since the shipped one cannot load it, and an irlume-pipeline threshold measurement). The stage stays closed today for lack of that integration, not by declination.

**Low, fixed: the brightness ranges the CSVs contradicted** are replaced by the per-burst midpoint rule.

One new fact the cross-check surfaced, recorded in the doc as an open instrument note: irlume's parity-tested BlazeRescue wiring reads higher short-range far rates than the official runtime on this corpus at the same score floor (preprocessing, not thresholds). It changes no conclusion here, and it means the Rust bench's Blaze columns measure irlume's wiring rather than the model's ceiling.

Issue #295, opened 2026-08-05
Native LiteRT runtime support: run .tflite models unconverted, full-range BlazeFace first


archledger:
Decision (2026-08-05): irlume gains native LiteRT/TFLite runtime support, so Google's `.tflite` artifacts run unconverted, byte-for-byte as published. The alternative (a pinned tf2onnx conversion pipeline) was considered and set aside: conversion keeps a converter version inside the trust chain, and for the MediaPipe family the published artifact is the thing the model card and hash actually describe. The costs are accepted knowingly: a second inference runtime in the daemon, a vendored native library across all five packaging lanes, and the SSD decode still being irlume's own code either way, since MediaPipe implements decode outside the model file.

## Integration shape

Same contract as the existing onnxruntime integration, established from a source-level review of the candidate bindings:

- **Binding: `edgefirst-tflite`** (classic TFLite C API side only; its LiteRT-Next side stays unused). It is the one candidate whose mechanism matches `ort` load-dynamic: `libloading` dlopen at runtime, an env/explicit-path override that slots into irlume's resolver, `Result`-based absence handling (`is_litert_unavailable()`), and pre-generated bindings, so distro builds need no network, no bindgen, no Bazel. Known risk, accepted: the crate is months old with a single corporate maintainer; pin the exact version and audit its discovery/loader module once before shipping. Fallback if it misfits: a hand-rolled `libloading` FFI against the ~25 stable `TfLite*` C-API symbols, the same resolver pattern irlume already maintains for ort.
- **Native library:** build `libtensorflowlite_c.so` from a pinned TensorFlow tag (v2.19.0 is the working candidate) with the documented CMake path (`tensorflow/lite/c`), XNNPACK on. The CMake tree fetches ~15 third-party deps at configure time, so the Fedora/Debian lanes carry them as extra pinned source tarballs with `FETCHCONTENT_SOURCE_DIR_*` overrides; AUR and Nix fetch natively. Bundle and sha256-pin the built .so exactly like the bundled onnxruntime. To measure before committing the lane: build time and .so size in mock, and whether the C-API path auto-applies the XNNPACK delegate or needs `TfLiteXNNPackDelegateCreate`.
[...]

PR #296, opened 2026-08-05
vision: native TFLite runtime plumbing, dlopen'd and pin-gated (#295 stage 1)

archledger:
Stage 1 of #295: the runtime plumbing that lets irlume run Google's `.tflite` artifacts unconverted, byte-for-byte as published. No shipped behavior changes; nothing loads a model that was not sha256-pinned first.

## Shape

Mirrors the ort integration deliberately:

- The C library (`libtensorflowlite_c.so`) is dlopen'd at runtime from an explicit, auditable path: `IRLUME_TFLITE_LIB` override wins alone (a broken override never falls through to a packaged copy), else the packaged locations. The crate's own `discovery()` is not used: its bare-soname fallback lets the dynamic linker pick up whatever is on the search path, and a root daemon loads code only from paths it chose.
- Absence is a typed, recoverable answer (`TfliteUnavailable`) naming what was tried, distinguished by cause: a broken override reads differently from "install the runtime". Only success is cached, so installing the library works without a daemon restart.
- `TfliteSession::from_pinned_bytes` takes verified bytes, the same one-buffer rule as `Engine::load_with_recognizer_bytes`: the digest that was checked and the model that runs cannot diverge.
- The candidate-path policy is a pure function (`tflite_lib_candidates`), tested without any dlopen.

## The adoption audit paid off twice

The dependency is `edgefirst-tflite = "=0.9.0"` exactly, default features off, because the audit covers those bytes. It found:

- **An upstream soundness gap:** the TFLite C contract (read from the v2.19.0 header, not recalled) requires the model buffer to outlive every interpreter created from it, but the crate's `build(&model)` returns an interpreter that does not hold the model, so safe code can drop the `Model` and invoke into freed memory. `TfliteSession` owns both with the field order documented as load-bearing. Worth reporting upstream.
- **XNNPACK is not auto-applied** by the C API path; the session applies the delegate explicitly, so inference speed does not depend on how the shared library was built.

## Measured for the packaging lane

`libtensorflowlite_c.so` from TF v2.19.0 via the documented CMake path: **210 seconds at -j8 on the 2.2GHz-capped Zenbook, 8.2MB artifact**, one flag needed on current CMake (`-DCMAKE_POLICY_VERSION_MINIMUM=3.5`, old vendored deps declare pre-3.5 minimums). Both numbers are comfortably below the bundled onnxruntime, so the per-lane bundling plan from #295 holds. The lane itself (specs carrying the ~15 dep tarballs for the no-network builds) is the next PR.

## Testing

- Resolver policy unit tests (override-wins-alone including the nonexistent-path case, empty-override-reads-as-unset, existing-paths-only).
- End-to-end smoke: the pinned full-range BlazeFace artifact (sha256 verified in-test) loads through the freshly built library, runs a zero tensor, and answers the measured contract, input `[1,192,192,3]`, outputs `[1,2304,16]` and `[1,2304,1]`; a wrong-length input is refused. Self-skips loudly without the library and model (every CI runner); the skip was proven to fire without the env and proven absent with it.
- `cargo test --workspace` green, clippy clean, `cargo deny check` fully ok after one scoped exception: webpki-roots (CDLA-Permissive-2.0) enters only as a build-dependency of edgefirst-tflite-sys via ureq, whose code runs solely under the `vendored` feature irlume keeps off. Scoped to the crate, not the license, so a future runtime CDLA dependency still trips the check.

## Not tested

- No decode, no detection: stage 2 (#295) builds the full-range decoder and parity gate on this.
- The packaged paths in `PACKAGED_TFLITE_LIBS` are exercised only through the pure candidate function; no distro package exists yet to load from them.
- Threads parameter and XNNPACK performance are unmeasured; stage 2's parity/latency work covers that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,7 +136,7 @@ dependencies = [
  "quote",
  "regex",
  "rustc-hash",
- "shlex",
+ "shlex 1.3.0",
  "syn 2.0.118",
  "which",
 ]
@@ -240,12 +240,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8f1fe948ff07f4bd06c30984e69f5b4899c516a3ef74f34df92a2df2ab535495"
 
 [[package]]
+name = "bytes"
+version = "1.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc652a48c352aef3ea3aed32080501cf3ef6ed5da78602a020c991775b0aff04"
+
+[[package]]
 name = "castaway"
 version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dec551ab6e7578819132c713a93c022a05d60159dc86e7a7050223577484c55a"
 dependencies = [
  "rustversion",
+]
+
+[[package]]
+name = "cc"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5add81bb678e6cb321aff7fa0dc7689ad82b112dbc032cea19f91d6b8e3582b9"
+dependencies = [
+ "find-msvc-tools",
+ "shlex 2.0.1",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


archledger:
Codex round (one, per policy): three merge-blocking findings, all real, all fixed in 32e2c12, plus the follow-ups.

**High, confirmed, and it contradicted my own module doc: a bare-soname override was a loader search.** `IRLUME_TFLITE_LIB=libtensorflowlite_c.so` would have sent dlopen walking LD_LIBRARY_PATH and the loader cache as root, the exact "library the daemon did not choose" the doc promised to refuse; and `env::var().ok()` collapsed a non-UTF-8 override to "unset", falling through to a packaged copy against the override-wins-alone rule. The override must now be an absolute path (`OverrideInvalid` names the reason), the value is read losslessly via `var_os`, and both behaviors are pinned by tests.

**High, confirmed: the pin was a naming convention, not a check.** `from_pinned_bytes` now takes `expected_sha256` and verifies before anything parses the bytes; the digest check precedes even the runtime load, so the refusal is testable on machines with no TFLite library, and a stage-2 caller cannot reach the parser with unverified bytes. The comparison I made to `Engine::load_with_recognizer_bytes` was wrong in the way the review said: that function computes an identity, it does not enforce a pin, and this constructor now does.

**Medium, confirmed: tensor types were trusted.** The crate's typed views only check that `T` fits, so an f32 view of an Int64 tensor silently writes half of each element. Input and outputs must now be Float32 with exactly one input, refused by name; `run_f32` re-runs the contract check so a caller that never asked for the shape is equally covered.

**Follow-ups taken:** packaged-path attempts keep per-path causes ("does not exist" vs "failed to load"), the deny.toml comment now says the build script does compile ureq (its code path just never runs and nothing links into shipped binaries), and the packaged-path comment says no package installs the bundled location yet.

**Follow-up declined, with the reasoning on the record: the self-skipping smoke test stays.** Skip-with-eprintln when hardware or a runtime is absent is this repo's standing convention (every TPM-gated test works this way), the skip was proven to fire and proven absent in both directions, and the container/local validation lanes run the real path. Making it fail or vanish on CI would diverge from that convention for no added coverage.

On 32e2c12: workspace green, clippy clean, `cargo deny check` fully ok, fmt clean, and the end-to-end smoke re-run against the built library with the new constructor signature, zero skip lines.

PR #297, opened 2026-08-05
packaging: bundle the TFLite C runtime on every lane (#295)

archledger:
The #295 packaging lane: every distribution bundles the TFLite C runtime the same way it bundles onnxruntime, consumed as a pinned artifact.

## The artifact

Google publishes no prebuilt Linux x86_64 C-API library at stable URLs, so irlume builds its own and plays Microsoft's role in the established pattern: `scripts/build-tflite-runtime.sh` clones the pinned tensorflow tag (v2.19.0, refusing a mismatched checkout), builds `tensorflow/lite/c` via the documented CMake path with XNNPACK on, and produces a tarball carrying the .so, the TensorFlow LICENSE, and a PROVENANCE file naming the exact source commit. Published on the `tflite-runtime-v2.19.0` release (signed tag, signed SHA256SUMS, not marked latest so `irlume update` still resolves v0.8.1), sha256 `04282fce406ac3408478f5559e96b0ac225046f89ffde6a85d1f1bef41554867`, verified to round-trip by download. Measured: 210s at -j8 on the 2.2GHz-capped Zenbook, 8.2MB.

## Per lane

All four install at `/usr/share/irlume/tflite/`, the first path the daemon resolver probes, so unlike onnxruntime there is no environment drop-in to drift; a missing library stays a recoverable "runtime not installed" answer.

- **Fedora:** Source6 with the same verify-then-unpack pattern as Source1, digest pinned beside `tflite_ver`.
- **Arch:** an extra pinned source in the PKGBUILD; no usable system package exists (AUR's `libtensorflow-lite` is 2.4-era).
- **Debian/nfpm:** fetched and sha256-checked in `build-deb.sh` staging, shipped by `nfpm.yaml`.
- **PPA:** vendored into the self-contained orig tarball by `build-ppa-source.sh` (Launchpad builders have no network), installed by `debian/rules`.
- **Nix:** not wired, and the README says so; a flake user sets `IRLUME_TFLITE_LIB`.

## Testing

- The build script ran end to end and produced the published artifact; the spec/PKGBUILD/scripts changes are exercised by the Packit rpm-build lanes on this PR (the srpm fetches Source6 from the live release).
- `bash -n` on every touched script; the deb and PPA container builds are not run here and will be exercised at the next release cut, which is when those lanes always run.

## Not tested

- No mock/koji build of the full RPM locally; the PR's Packit builds are the check.
- The Arch PKGBUILD change follows the AUR-on-hold status: it is the source of truth for later, not pushed anywhere.


archledger:
Codex round (one, per policy): all five findings were real, fixed in 34b8dbf, and the two High ones would have shipped a lane that was dead on arrival exactly where the package declares support.

**High, confirmed by the reviewer's own objdump of the published artifact: the runtime could not load on the declared systems.** Built on a current Fedora, it demanded GLIBC_2.43 and GLIBCXX_3.4.32 against Ubuntu 22.04's 2.35 and GCC-12-era libstdc++. The artifact is rebuilt inside ubuntu:22.04 (new `build-tflite-runtime-container.sh`; the minimal image needed make and, for XNNPACK's microkernel generator, python3, both now baked in), the build script refuses any output whose symbol floors exceed GLIBC 2.35 / GLIBCXX 3.4.30, and `build-deb.sh` asserts the same floor on the staged library beside its existing binary checks. The rebuilt library measures GLIBC_2.34 / GLIBCXX_3.4.30, passes the end-to-end smoke test, and has replaced the release assets with a re-signed SHA256SUMS; every lane's digest pin now reads `dd3abcdb…17bd311`. It also came out 0.9MB smaller.

**High, confirmed: AppArmor would deny the dlopen.** Both profile variants gain the `mr` rule for the bundled path, the rule joined the packaging-parity runtime list, and the guard was proven to fire: removing the rule from one profile prints MISS and exits nonzero.

**Medium, confirmed: the tag guard proved a name, not a commit.** The script now pins and verifies the tensorflow commit (`e36baa30…`), which matches the commit the review independently cited.

**Medium, confirmed: the PPA copyright misattributed the runtime.** `tflite-prebuilt/*` now carries its own stanza (The TensorFlow Authors, Apache-2.0, with the source commit), placed with the other per-file stanzas.

**Low, taken: the README now says four FHS lanes and names the Nix exception in the same sentence** rather than six lines later.

Both Packit rpm-builds also failed on the pre-fix sha with null srpm logs, on this PR and on #298 whose spec is untouched main; retriggered, and the new push runs them against the corrected Source6 digest.

PR #298, opened 2026-08-05
vision: full-range BlazeFace decoder, parity-gated against the official runtime (#295 stage 2)

archledger:
Stage 2 of #295: the full-range BlazeFace decoder on the native TFLite plumbing, parity-gated against Google's own runtime before anything cites it.

## The decoder

Follows the measured contract and MediaPipe's own graph config, not the model card (which misstates the input size): 192x192 input, a 48x48 stride-4 anchor grid with one anchor per cell (2304), box scales 192, the sigmoid clipped at 100 like the short-range path, and the publisher's 0.6 floor as the default. The operating threshold for any catalog entry still comes from stage 3, through irlume's pipeline, live.

The whole decode decision is a pure function (`decode_full_range_best`), unit-tested with synthetic tensors: anchor placement (row-major, centers, bounds), the floor, the clip, and the non-finite rule from the short-range decoder, where a NaN regressor holding the best logit must neither become the answer nor shadow a finite runner-up. Preprocessing (square zero-pad letterbox, center-of-pixel sampling, [-1,1]) matches the short-range rescue line for line.

## Parity, measured

Over the committed stage-3 corpus (320 frames, both cameras) against the official mediapipe runtime, with the bench script extended to emit boxes:

- **199 of 199 detection agreements at matched floors; zero one-sided detections.**
- **Mean IoU 0.932** (min 0.782, p10 0.881); mean absolute score delta 0.038.
- The one raw disagreement dissolved on inspection: the official harness ran at floor 0.5 and the frame scored 0.587, which the 0.6 floor refuses on both sides once matched.
- Reference bar: the short-range decoder shipped on a 0.94-mean-IoU parity bench; this sits at 0.93 with independently written preprocessing.

The harness (`examples/blaze_full_parity.rs`) reruns the comparison from the stored corpus and the committed official-runtime CSV whenever either side changes.

## Not tested

- No wiring into detection selection, the daemon, or the catalog: the stage stays closed until stage 3's live threshold measurement.
- Latency and thread scaling are unmeasured.
- The parity corpus is one subject, two cameras, no dim segment, per the #294 record.


archledger:
Codex round (one, per policy): both merge-blocking findings were real, fixed in 8f5b19c.

**High, confirmed in full: the parity gate verified nothing.** The repo held a CSV dumper plus a claim; the comparison lived outside the tree, the instruments ran different floors (0.5 vs 0.6), and the harness had silent-empty paths that could shrink the corpus to zero rows and still exit clean. Now the harness is strict (unreadable directory, malformed frame, empty segment, or zero rows is a loud failure), the reference bench takes `--floor` and the parity run matches the decoder's 0.6, and `scripts/compare-blaze-parity.py` is the executable, fail-closed gate: identical frame sets required on both sides, a 300-frame minimum, regression pins from the measured run, and zero disagreements outside an explicit floor band. Both CSVs are committed to `docs/pad-results/`, so every published number recomputes from repo material with one command.

The matched-floor rerun also surfaced what my post-hoc aggregation had hidden: one floor-straddling frame (official 0.5774, rust 0.6233) where a few hundredths of cross-preprocessing score noise legitimately flips a binary decision at any threshold. The gate tolerates exactly that shape, inside floor+0.05, capped at five frames, every instance printed; anything else fails. Final gated numbers: 320 frames, 198 both-detect, mean IoU 0.9354, min 0.7821, mean score delta 0.0383, one tolerated frame.

**Medium, confirmed: a NaN class logit became a NaN-scored detection** because `NaN < floor` is false. The decoder refuses non-finite scores; the test stages a NaN logit alone (no detection) and with a finite runner-up (runner-up wins), and the guard is mutant-proven.

**Follow-up notes:** the raw frames stay in the session archive per the #294 convention (128MB; the committed CSVs are the recomputable evidence, and the doc says which is which); the parity measurement's reference runtime is mediapipe 1.0.0 as recorded in the doc; and the PNM readers stay capture-format-only, which the strict harness now converts from a silent skip into a named failure for anything else.

PR #299, opened 2026-08-05
detection: open the stage and wire full-range BlazeFace into the rescue slot (#295 stage 3)

archledger:
Stage 3 of #295, the last one: the detection stage opens, and full-range BlazeFace becomes a catalog entry wired into the rescue slot.

## The measurement that decides it

Through irlume's own pipeline this time, not the publisher's runtime, over the stage-3 corpus extended tonight with the two conditions it lacked: genuine dark (Zenbook lights-out, archhost dusk) and, for the first time in this project, **empty scenes** with the subject out of frame in every lighting state. 512 frames, scored with the floor dropped to 0.01 so the sub-threshold distribution is visible at all. Record: `docs/pad-results/2026-08-05-fullrange-threshold.md`, per-frame CSV beside it.

**The empty-scene population refuted the obvious threshold.** An empty room is not a zero: on near-black IR frames the model reads sensor noise and reaches **0.5293**, so inheriting the shipped rescue's 0.5 would admit a false detection. From 0.53 up, none are admitted. The genuine side is flat across 0.45 to 0.6 (60 to 61 misses at every candidate, because those frames score about 0.10, non-detections rather than a threshold effect), so the choice costs nothing genuine. **0.55 enabled**, mid-flat, 0.02 of margin.

Two findings worth the record: a false detection here is not an authentication event (the rescue box still has to survive mesh refine, alignment, recognition, and liveness, and an empty-room box matches no template), and true-dark detection is uneven per frame on the Zenbook's saturating strobe, 0.85 and 0.10 alternating inside one burst, which is a per-frame observation and not a per-attempt rate since an attempt captures many frames.

## The wiring

- `Stage::Detection.open()` flips, and the deliberate-act test that guards stage opening changes with it, as its comment demands.
- The `fullrange` catalog entry: Google's published `.tflite` unconverted, pinned by sha256, Apache-2.0 with consented first-party training data per the card.
- **Rescue slot only.** YuNet stays primary; the entry runs when YuNet finds nothing, so a bad entry's worst case is a missed rescue, a denial. One slot means one enum (`Rescue::ShortRange | FullRange`), not a second `Option` field that could disagree with the first, which is the #281/#285 shape.
- `third_party_detector` settings key with the same fail-closed resolve as the recognizer: an invalid selection, unreadable file, or pin mismatch refuses to start, and PAM reads an absent daemon as password fallback.
- The consent text for a detection entry states the rescue-only seat and the runtime dependency; `stage_override` is now the shared pure decision behind both per-stage wrappers.

## Live validation on this hardware

The daemon loads the entry (`third-party detector 'fullrange' loaded into the RESCUE slot (threshold 0.55; YuNet stays primary)`), and its rescue-status line now names the actual occupant, since the old line said "BlazeFace" whatever was loaded. With the TFLite runtime removed it refuses to start and names the cause, which is the fail-closed behavior the docs promise.

## Not tested

- Per-attempt dark rescue rates, latency, and throughput with the entry enabled.
- One subject, two cameras, one session per condition; the negative population is empty rooms, not other people's faces, which is the right question for a detector but is not a population false-accept rate.
- The degenerate-exposure exclusion (frame mean 8 to 235) is stated in the doc so it is auditable; including those frames does not move the threshold.


archledger:
The round was right on the finding that matters, and it changes this PR's outcome: **the detection stage stays closed.** Fixed in 0ddcb28.

**High, confirmed from the control flow before acting.** My load-bearing claim, that a bad detection entry's worst case is "a missed rescue, a denial", is false. `rescue_detect` fills `rgb_top` when YuNet returns nothing, and that box is aligned, embedded, matched against enrolled templates, and can reach a grant. The daemon's own startup warning says the built-in liveness gate does not stop a life-size print of the enrolled face without the opt-in PAD cue, so a detector that accepts presentations YuNet declines widens precisely that path. My corpus is empty rooms: no prints, no screens, no other faces. It establishes an operating point and says nothing about false grants, which is the evidence a grant-capable stage needs.

So `Stage::Detection.open()` returns to false, the deliberate-act test is restored, and the `fullrange` catalog entry is withdrawn (the catalog's own invariant forbids closed-stage entries, and an unselectable entry must not be listed). What stays, dormant behind the stage gate: the decoder, the `Rescue` enum, the settings key, and the fail-closed resolve, with fixture-driven tests pinning the refusal that keeps them dormant.

**Medium, confirmed and worse than reported.** The comment beside `open()` quoted "genuine minimum 0.6038" and "nothing above 0.01" from a superseded first run and I never updated it when the corrected numbers landed; the CSV says 0.0956 and 0.5293. The guard's comment now carries the real figures and the real reason. The threshold document's own finding 2, which generalised the empty-room case into "a false detection here is not an authentication event", is corrected in place and says so.

**Both Lows taken.** Health and the TUI now carry the loaded detector, since the daemon is the only authority on it and a TUI that cannot show it cannot verify what is running; the settings-key coverage is now fixture-driven so it pins the refusal rather than a name that no longer exists.

Remaining work is recorded in the threshold document: an end-to-end corpus measured on frames where YuNet returns nothing, carrying prints, screens, other faces, and genuine users through to the authentication outcome. That needs the vinyl print from the #235/#237 work and a session at the cameras.

Workspace green, clippy clean, fmt clean.

PR #300, opened 2026-08-05
tui: the daemon owns the cameras, so the TUI stops opening them behind it (#187)

archledger:
Fixes the enrollment failure in #187, whose cause the reporter's own diagnostic finally caught.

## What their trace showed

The fuser loop they ran during an enrollment attempt caught **two processes holding /dev/video0 at once**:

```
/dev/video0:  dale  69700 F.... irlume
/dev/video0:  root   2708 F...m irlumed
/dev/video2:  root   2708 F...m irlumed
```

`irlume` there is the TUI, not the daemon. Reproduced on the reference Zenbook with nothing but `irlume tui` running: `/dev/video0: wisbfime 350409 F.... irlume`, the same signature. It is transient, held during the startup probe window rather than leaked for the process lifetime, which is why it took a live sampling loop to see.

## Cause

Classifying a video node opens it (`classify_node` → `Device::with_path` → `enum_formats`). Both background poll paths did that on every refresh: `LightState::gather` called `discover_nodes`/`list_pairs` before it had even checked whether the daemon was up, and `Probes::gather` called `capabilities` plus both enumerations. So the TUI opened every camera node repeatedly while the daemon might be streaming those same nodes.

A second open is harmless on tolerant UVC modules, which is why this was invisible on my hardware for months. On strict ones it is EBUSY, and it lands on the user as a failed enrollment naming a holder they cannot close. This is the same family as #242 and #243, one layer out: those were the daemon colliding with itself, this is the TUI colliding with the daemon.

## The rule

While the daemon is reachable it is the authority on cameras: it already has them open and reports its tier and devices through Health, so the TUI takes capabilities from there instead of looking for itself. With the daemon down the local probe is both the only source and safe, since nothing else holds the cameras. That is one pure function, `enumerate_cameras_now`, pinned by its own test.

The Cameras picker genuinely needs the full listing, so it enumerates on **entering that screen** rather than in background polls: an explicit act that cannot coincide with an authentication the user did not start.

Two absent-vs-not-observed guards ride along, because the change creates exactly that hazard: a skipped enumeration returns empty vectors that must not overwrite a known listing, and the sweep's capabilities are adopted only when actually probed, with `caps_probed` carrying the distinction rather than an all-false default pretending to be an observation. The second one made an existing test fail honestly, and the fix was to say what the fixture meant.

## Verified on hardware

- **Before:** sampling while `irlume tui` runs catches the CLI holding /dev/video0.
- **After:** 40 samples over 20 seconds, daemon up, TUI running: **zero** holds.
- The tier still reads `Face (IR) · secure` from Health, and the picker still lists `video0+video2 built-in [3277:0059]`, so neither half of the capability story regressed.
- Workspace green, clippy clean, fmt clean.

## Not tested

- I cannot reproduce the EBUSY itself: this module tolerates the second open. The claim is structural rather than empirical on the reporter's failure mode, and it is the right kind of claim, because if the CLI never opens the node while the daemon may be using it, the collision cannot occur on any module.
- The reporter should confirm on their hardware before this issue closes.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -23,6 +23,10 @@ pub use irlume_camera::{
 /// devices without depending on the camera crate directly. See
 /// [`irlume_camera::select_pair`].
 pub use irlume_camera::{capabilities, device_identity, select_pair};
+/// Enumerate the Hello camera pairs. Re-exported for the daemon's
+/// camera-class `ListCameras` arm: clients must not enumerate for themselves
+/// (#187), so this is the only path to a listing.
+pub use irlume_camera::{list_pairs, privacy_engaged, CameraPair};
 
 /// Loaded models + camera device selection. Build once, reuse per request.
 pub struct Engine {
--- a/crates/irlume-cli/src/fingerprint.rs
+++ b/crates/irlume-cli/src/fingerprint.rs
@@ -334,7 +334,7 @@ fn enable(user: &str, args: &[String]) -> ExitCode {
     // both stay active and the user unlocks with whichever is convenient.
     // `--fingerprint-only` forces the old fingerprint-only mode (face disabled).
     let fingerprint_only = args.iter().any(|a| a == "--fingerprint-only");
-    let coexist = !fingerprint_only && irlume_camera::capabilities().rgb;
+    let coexist = !fingerprint_only && crate::caps().rgb;
     // Enroll a finger first if the user has none.
     if !fp::has_enrollment(user) {
         println!("[fingerprint] no finger enrolled yet; enrolling one now");
--- a/crates/irlume-cli/src/machine.rs
+++ b/crates/irlume-cli/src/machine.rs
@@ -417,7 +417,7 @@ pub fn status(args: &[String]) -> ExitCode {
     // Secure-tier hardware, that does not exist. The daemon's own Health reply
     // has always paired the capability probe with an existence check; this now
     // agrees with it.
-    let caps = irlume_camera::capabilities();
+    let caps = crate::caps();
     let (rgb, ir) = irlume_camera::select_pair();
     let camera = json!({
         "rgb": caps.rgb && std::path::Path::new(&rgb).exists(),
--- a/crates/irlume-cli/src/main.rs
+++ b/crates/irlume-cli/src/main.rs
@@ -274,7 +274,7 @@ fn enroll(args: &[String]) -> std::process::ExitCode {
 fn offer_blink_challenge(user: &str) {
     use std::io::{BufRead, IsTerminal, Write};
     // Only meaningful on IR-capable (Secure-tier) hardware.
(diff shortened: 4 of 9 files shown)


archledger:
All four findings were real, and the first one exposed that my verification could not have detected the difference. Fixed in fd17951.

**My "zero holds" was not evidence.** I sampled with `fuser` every 0.5 seconds and reported zero. Running strace over the same window counted **12 camera opens**, and **160** once the Cameras screen was visited. Sampling at half-second intervals for a millisecond-scale open cannot distinguish absence from bad luck, and I presented it as proof. Every number below is now from strace, which counts syscalls rather than looking for them.

**High 1, confirmed:** `App::new` probed before any daemon contact, so a TUI started during an authentication opened every node. Also `refresh_camera_listing` on picker entry, which my "explicit act" comment defended and which says nothing about a concurrent PAM authentication, exactly as you wrote.

**High 2, confirmed and the bigger one:** `Health` probed cameras on a connection thread, outside the camera worker, contradicting the Status class's documented "touches no camera" contract. Even a perfect TUI could not have avoided that. The camera facts are now published into `EngineBits` once at engine load and Health answers from memory, which makes the arbiter's contract true again.

**High 3, confirmed, and the sharpest:** a Ping that times out read as "daemon down" and licensed local opens. A timing-out Ping is precisely what a daemon busy with the camera produces, so the fallback fired when it was most dangerous. There is now no daemon-state gate at all: the TUI never opens a node under any condition.

**Medium 4, confirmed:** the stale-listing guard could not distinguish an empty answer from an unobserved one. Enumeration moved behind a camera-class `ListCameras` request that the arbiter serializes against captures, so the answer is authoritative, and a `pairs_known` flag separates "the daemon said none" from "the daemon has not answered". The picker no longer prints "no camera found" while the active-pair line right below it names two devices, which is what an older daemon produced.

Two further defects I found by measuring rather than reading, both mine:

- `draw_cameras` called `select_pair()` **per rendered frame**, and that helper falls through to discovery when no pair is configured. That was the bulk of the 160.
- My source-property test passed while the behavior persisted, because it scanned only `tui.rs` while the opens came through `pamwire::wants`. The CLI's capability lookups now route through a single daemon-first cached accessor, and the guard bans `select_pair` too.

Privacy-shutter reads moved to the daemon as well, since reading that control also opens the device; each `CameraPairInfo` carries its pair's state.

**Verified:** 0 camera opens across a 26-second session including a Cameras-screen visit, against a daemon built from this branch (the installed 0.8.1 daemon predates `ListCameras`, so the first run of the picker was inconclusive until I ran the branch's own daemon). Picker still lists `video0+video2 built-in [3277:0059]`; tier still reads secure. Workspace green, clippy clean, fmt clean.

Not addressed, deliberately: `irlume doctor` still enumerates locally. It is a one-shot the user runs, and its job is to report hardware when the daemon is down, which is when a daemon-first path would be useless.

Issue #312, opened 2026-08-06
Enrollment can store scans that only authenticate in the lighting they were captured in, silently


archledger:
Found on the Brio during the #187 hardware session, and it is a lockout path a real user can walk into.

Enrollment completed with 10 scans while daylight supplied the infrared the camera's own emitter never produces (see the Brio compat issue). Every scan passed the live gate legitimately. The very next `identify`, in conditions without ambient IR, was denied `no face in IR`, fail-closed and correct, and would be denied every time thereafter in those conditions.

So the sequence "enroll at noon, locked out at night" is real on any camera whose emitter does not fire: enrollment measures a property of the room and stores it as a property of the user. Nothing at enrollment time says so.

Fix direction: enrollment already sees per-scan `ir_bright` and the ambient reading. When scans succeeded but the ambient contribution is what lit them (high ambient with the emitter's own contribution unproven), finish the enrollment and SAY it: a completion note that dark-room login is unverified on this camera, with the one-command check. A stronger variant probes one frame with the scene ambient subtracted, but the honest warning alone would have turned this session's confusion into one sentence.

archledger:
The honest-warning half shipped in #330. Enrollment now measures, per scan, the room's share of the IR burst's lit-frame brightness, and only when the burst actually observed an emitter-off frame (a camera-classified dark frame; the burst-minimum fallback never counts, so a steady emitter in a dark room cannot be misread as room light). Scans at share 0.5 or above count as ambient-lit; anchors are measured (a working emitter reads near 0, this issue's daylight case near 1). The count reaches every completion surface with the same sentence: N scans were lit mainly by the room, dark-room login is unverified, check it with the lights off via irlume identify. AddScan carries the count through an opt-in structured reply, and the TUI's merge continuation folds the first scan's count in, so multi-scan enrollments are covered end to end.

The stronger variant this issue sketched, probing one frame with the scene ambient subtracted at enrollment time, was not built; if the warning proves too quiet in the field (a camera that never classifies frames gets no warning at all, by design, because there is no observation to attribute), that variant is the next step and deserves its own issue.

Issue #311, opened 2026-08-06
Hardware notes: Logitech Brio 4K needs ambient IR (emitter never strobes) and cannot dual-stream on USB2


archledger:
Hardware compatibility record from the #187 session: Logitech Brio 4K (046d:085e) on a USB2 link, CachyOS, kernel 7.1.6.

Three findings, each measured:

1. **Dual streaming starves RGB completely.** With the IR stream armed, RGB STREAMON succeeds and then no frame ever arrives; the queue dies with QBUF EINVAL and dmesg logs `Failed to resubmit video URB`. The fix on fix/brio-sequential-capture makes sequential mode arm one stream at a time, which restores capture. Untested whether a USB3 link removes the starvation; the camera negotiated USB2 on the test box.

2. **The IR illuminator never strobes.** Its Hello extension control (unit 12, selector 0x06) reads as already holding the enable value, the alternate selector 0x0a is not advertised, and the LED stays dark through every capture. The IR sensor itself works: under daylight (850nm-rich) the IR path detected and enrolled; in ambient-IR-free conditions the same path correctly reads a black frame. Consequence: secure-tier authentication on this camera works only when the environment supplies infrared, which is a coincidence, not a capability.

3. **The IR sensor answers 340x340 GREY** regardless of the requested size (asked 640x400, granted 340x340), right at the Windows Hello minimum.

Suggested handling: a hardware-notes entry for the Brio 4K stating RGB works, IR works only with ambient IR, and recommending convenience tier or an external Hello camera with a working emitter (NexiGo N930W on the same box: emitter fires, `ambient=0`, `ir_bright=94`, full enroll and match). Whether `ir-setup` can learn to drive this emitter through some other control is an open question; nothing the camera advertises admits to it.

archledger:
Recorded in docs/PLATFORMS.md via #327 (e758c36), with today's USB3 measurements added to the #187 session's USB2 ones: the emitter strobes from bare STREAMON on a USB3 link only (dark-phase mean 0.6, lit 126 to 224 in an ambient-IR-free room, after a cold re-enumeration with no extension-unit write), held dual-streams starve RGB on BOTH links (EINVAL three of three held rounds on USB3), sequential capture works on both, and per-frame illumination metadata works on USB3 once #310's pairing fix reads the right queue (0/20 to 20/20). Verdict per link is in the doc; full enrollment on USB3 is the remaining unexercised piece.

Issue #310, opened 2026-08-06
Illumination-metadata sibling lookup arms the wrong node on single-interface cameras


archledger:
Found on a Logitech Brio during the #187 hardware session; refuted as that session's queue-killer but real on its own.

`metadata_node_for` picks the first sibling on the same USB interface that offers UVCM, lowest node number first. That rule was written for cameras whose RGB and IR functions sit on separate interfaces. On a camera whose four nodes register against ONE interface (the Brio: video0 RGB, video1 RGB-metadata, video2 IR, video3 IR-metadata), the IR camera's search walks video0 and video1 first and arms video1, which is the RGB stream's metadata queue.

The observable symptom is the reader classifying nothing, forever: every capture logs `illumination metadata closed after 0 classified frames`, because RGB-stream metadata timestamps never match IR frames. The cost is one doomed open, format probe, buffer arm, and teardown of the wrong node on EVERY capture, plus the illumination signal silently never working on such cameras (they fall back to brightness, which is the guess the reader exists to replace).

Fix direction: resolve the metadata node by its video sibling's entity/function pairing rather than interface adjacency, or at minimum prefer node adjacency (videoN pairs with videoN+1 in uvcvideo registration order) and verify the first classified frame's timestamps actually correlate before keeping the reader. `IRLUME_NO_ILLUM_META=1` (added on fix/brio-sequential-capture) skips the reader as a diagnostic.

archledger:
Fixed in #328 (ad91739). The pairing rule is the immediately next same-interface node above the image node, kept only when it offers the format: a lower-numbered candidate is an earlier stream's queue, and scanning forward past a gap would borrow a later stream's (uvcvideo ignores a failed metadata registration, so a stream can lack its metadata node while a later stream has both). Measured on the Brio the issue describes: 0 of 20 burst frames classified before the fix, 20 of 20 after, with the illumination source flipping from brightness fallback to the camera's own metadata. The camera reported per-frame illumination all along; the reader was subscribed to the RGB queue.

Issue #309, opened 2026-08-06
TUI reports a capture that never answered as 'no face detected'


archledger:
Found during the #187 hardware session, and it cost the session an hour of chasing the wrong thing.

While the daemon's capture path was stuck cycling a broken stream (a multi-minute retry loop), the TUI's enrollment screen reported **"no face detected"** with the user's face plainly in frame. The daemon journal for the same window shows no detection ever ran: no frames were captured at all. The TUI's poll timed out (three `Broken pipe` entries as it hung up) and the timeout rendered as a biometric verdict.

"No face detected" and "the camera never answered" call for opposite user actions: one says move into frame and add light, the other says nothing you do in front of the camera can help. Reporting the first when the truth is the second sends users into lighting adjustments against a hung capture, exactly as happened live.

Fix: the TUI must distinguish assess-response-says-no-face from assess-did-not-answer (timeout, connection error) and say the second plainly, with the daemon journal pointer. Same misleading-message family as the fixed "camera busy, in use by irlumed" (#242/#243).

archledger:
Fixed in #327 (e758c36). The framing guide's polls now run a bounded 10s budget; a guide that stops answering replaces the whole live panel with "Camera guide not answering; this is not about your face or lighting", the transport error, and the journalctl pointer, and three consecutive misses (framing and countdown counted together) end the enrollment saying the guide never answered. Regression tests pin that the exact stale text from the #187 session can never render as current, and a worker-level scripted test drives the give-up path.

PR #313, opened 2026-08-06
True sequential capture, in-place stream recovery, honest capture errors (Brio hardware session)

archledger:
The camera-layer fixes from the #187 hardware session on a Logitech Brio over USB2, where enrollment failed with "no face detected" and the diagnosis ran four defects deep. Companion issues: #308 (camera-tune probe shape), #309 (TUI message), #310 (metadata sibling), #311 (Brio hardware notes), #312 (enrollment lighting warning).

## The defects, each syscall-proven on hardware

1. **"Sequential" only ordered the reads.** Held sessions arm their streams at creation, so sequential mode still had both streams live, and on the Brio the armed IR stream starves RGB of isochronous bandwidth entirely: STREAMON succeeds, zero frames ever arrive (a 6-second all-EAGAIN gap in the strace), the queue dies with QBUF EINVAL, and dmesg logs `Failed to resubmit video URB`.
2. **The capture retry collided with its own dead stream.** On a held-session failure, the retry re-opened the device while the caller's broken session still owned the buffer queue: QBUF EINVAL at .266366, the retry's S_FMT EBUSY at .269393, no close between. The user-facing result was "camera busy, close that app" with nothing to close, the same phantom-holder family as #242/#243.
3. **The retry log hardcoded "concurrent"** whatever mode actually ran.

## The fixes

- Sequential mode now vetoes the held-session fast path in `capture_scans`; the per-frame path opens one stream at a time and releases it before the other starts. One resolver (`sequential_capture_selected`) serves both the assess read-ordering and the enroll session decision, so the two halves of the answer cannot disagree again.
- The resolver's decision is a pure function over its two observations (env override, stored per-camera measurement), tested in both directions of every arm, three mutants run and killed (stored-arm flip, env-arm consulting stored, default flip).
- `RgbSession` recovers a broken stream in place: the stream field is an `Option` so the dead stream drops (STREAMOFF, buffer release) before its replacement negotiates on the same fd, the assign-order trap spelled out in the comment. A held-session capture failure gets one in-place recovery, then propagates; the standalone reopen retry runs only when this call opened one-shot.
- The retry log names the mode that ran. `IRLUME_NO_ILLUM_META=1` skips the illumination-metadata reader as a diagnostic (used to cleanly refute that reader as the queue-killer; its wrong-sibling defect is #310).

## Hardware validation (CachyOS mini PC, N100)

With all fixes: enrollment completed on the Brio (10 scans, daylight supplying IR per #311), and on a NexiGo on the same box, enrollment and a live identify match at 0.661 through the full liveness gate with the emitter at ambient=0. Before the fixes, the same box produced the reporter's exact failure signatures. The EINVAL-at-stream-start now propagates truthfully when recovery cannot save the capture, instead of masquerading as a phantom camera holder.

## Not tested

- `RgbSession::recover` and the held-failure propagation have no CI-reachable test (they need a camera whose stream breaks mid-capture); their validation is the hardware session above.
- The IR session side has the same held-stream shape; sequential mode sidesteps it, but an in-place IR recovery is not implemented here.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -645,12 +645,71 @@ fn top_detection(faces: &[Detection]) -> Option<&Detection> {
 /// live anyway, which on a bandwidth-starved camera is indistinguishable
 /// from concurrent (#187).
 fn sequential_capture_selected(rgb_dev: &str) -> (bool, &'static str) {
-    match std::env::var("IRLUME_SEQUENTIAL_CAPTURE") {
-        Ok(v) => (v.trim() == "1", "IRLUME_SEQUENTIAL_CAPTURE"),
-        Err(_) => match irlume_camera::stored_capture_mode(rgb_dev) {
+    capture_mode_decision(
+        std::env::var("IRLUME_SEQUENTIAL_CAPTURE").ok().as_deref(),
+        irlume_camera::stored_capture_mode(rgb_dev),
+    )
+}
+
+/// The decision itself, pure over its two observations so every arm is
+/// testable without a camera or an environment mutation: a set env var
+/// decides alone (even when it says concurrent, because setting it is an
+/// explicit instruction and the stored answer must not outrank it), then the
+/// stored per-camera measurement, then the concurrent default.
+fn capture_mode_decision(
+    env: Option<&str>,
+    stored: Option<irlume_camera::CaptureMode>,
+) -> (bool, &'static str) {
+    match env {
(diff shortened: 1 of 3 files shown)


archledger:
Review round applied: the capture-mode decision is snapshotted once per enrollment and carried through every held assessment (and the per-frame path, one step past the finding, so one enrollment runs under one policy throughout). Fresh reads remain only for one-shot calls with no caller snapshot. The archhost CI failure is the pre-existing PATH-fake test race in irlume-cli (models::tests::a_failed_daemon_restart_is_reported_not_swallowed), untouched by this branch and green on main minutes earlier; job rerun requested.

PR #314, opened 2026-08-06
Native MediaPipe benchmark: short-range parity gate, blendshapes probe, runtime latency

archledger:
The native-runtime benchmark of every MediaPipe model relevant to face authentication, per the session's direction: test them all, benchmark them, and say which is better. Full findings and the per-stage verdict live in `docs/pad-results/2026-08-06-mediapipe-native-benchmark.md`; per-frame evidence in the three committed CSVs, with the external corpus content-pinned by `2026-08-05-stage3-corpus.sha256` (both harnesses verify every frame against the manifest and refuse leftovers or differences; proven firing on a tampered hash and a missing entry).

## What was measured

1. **Short-range BlazeFace parity** (`blaze_short_parity.rs`): the one shipped model pair never compared on the native runtime. Both models run in one process on the identical input tensor through the identical decode, so the residual is the ONNX conversion itself. Over all 512 stage-3 frames: mean IoU 0.999999, min 0.999996, max score delta 1.967e-6, 139/512 scores bit-identical, zero one-sided detections. The harness is an enforcing gate (pinned 512/512 denominator, IoU bound 0.9999, delta bound 5e-6) with a skew self-test: a 2px shift injected into only the native input drops mean IoU to 0.757 and zeroes the bit-identical count.
2. **Blendshapes, first contact** (`blendshapes_probe.rs`): contract read from `face_blendshapes_graph.cc` ([1,146,2] pixel coords; the 146-index subset ends in iris points 468-477 ;  CORRECTION: an earlier body claimed the shipped ONNX mesh lacks these; measured, it emits all 478 points, so it can feed this model; what remains unvalidated is iris PARITY between the meshes, since #306 compares the 468 shared-topology points only). Over the 223 accepted frames: pooled Pearson r -0.938 between max eyeBlink and production min EAR, all outputs finite, open-eye segment means at most 0.37 (glasses-on IR, the same conditions that degrade EAR). The corpus holds no controlled eyelid closures, so this is an observational open-eye baseline, not evidence of blink sensitivity; the doc says exactly that.
3. **Latency** (`mp_latency_bench.rs`, Zenbook 258V, 100 iters after warmup): native 478-pt mesh 5.79ms at 2 threads vs 10.34ms shipped ONNX mesh in the committed run; short-range Blaze 1.23ms at 4 threads vs 4.23ms ort; full-range 4.29ms; blendshapes 1.72ms with per-frame input construction inside the timed closure. Across repeat runs the ort means moved (mesh 7.6 to 10.3ms) while every tflite row held within its own p95, so the doc claims direction plus tail tightness, not a point percentage.

## Production code touched

`BlazeRescue`'s preprocessing and decode are extracted as pure functions (`blaze_letterbox_input`, `decode_short_range_best`) so the harness exercises the production arithmetic instead of a copy, and `detect_top_at(floor)` mirrors the existing `FullRangeBlaze` measurement pattern. `detect_top` delegates with the unchanged threshold. New tests cover the decode's reject half (floor, NaN regressor) and apply half (known regressors decode to the expected bbox at a nonzero anchor) separately.

## Verdict (details and reasoning in the doc)

- **Primary detection**: YuNet stays; nothing here competes with its measured RGB coverage.
- **Rescue detection**: full-range remains the measured winner at range; the stage stays CLOSED pending the end-to-end false-grant corpus (#276). No accuracy number reopens it.
- **Landmarks**: recommend switching to the native 478-pt .tflite in a follow-up PR (#315). Fidelity (mean NME 6.9e-7, #306), latency (faster in both bench runs with a tighter tail), and supply chain (drop the unreproducible conversion, byte-pin Google's artifact) all point the same way, The mesh_parity gate is the acceptance test, extended to all 478 points (the iris tail was never compared).
- **Short-range runtime**: no move; the conversion is proven faithful.
- **Blendshapes**: not evaluated as an authentication cue; an all-open-eye corpus cannot support that. Open-eye baseline recorded; evaluation blocked on a labeled blink corpus (#316).

## Review round (applied in fb7826e)

One Codex round per policy; all four findings were real and are applied: the count-only corpus pin became the sha256 manifest above, the blendshape causal claim became an observational baseline, the blendshape latency row now times input construction, and the public decoder rejects malformed head/anchor lengths and a non-finite floor instead of panicking (tests for each).

## Not tested

- Latency on one machine only; the mini PC and archhost will differ.
- No closed-eye frames exist in the corpus, so blendshape open/closed separation is unmeasured.
- The parity gates ran the native runtime single-threaded; a production 2-thread mesh should re-run mesh_parity at that thread count.


archledger:
Review round applied in fb7826e; all four findings were real.

1. **Count-only corpus pin (High)**: `docs/pad-results/2026-08-05-stage3-corpus.sha256` now pins all 512 frame contents, and both harnesses consume it entry-for-entry: every frame's sha256 is verified as read, an unknown frame refuses, and leftovers fail the run. Proven firing both ways before trusting it: a tampered hash exits with "content differs from the manifest", a manifest missing one entry exits with "not in the corpus manifest". The doc header no longer calls the frames committed; it names the manifest as what binds the gates to this evidence.
2. **Blendshape causal overclaim (Medium)**: the section is retitled to open-eye behavior, the correlation is stated as observational with the confounds named, and the verdict entry now reads not-evaluated-as-an-authentication-cue. Issue #316 got the same correction.
3. **Blendshape latency contract (Medium)**: input construction moved inside the timed closure and the bench re-ran; the committed CSV is regenerated from the corrected code (blendshapes 1.72ms). The re-run also surfaced ort run-to-run variance (mesh mean 7.62 to 10.34ms across runs while the tflite rows held within their own p95), so the doc and #315 now claim direction and tail tightness instead of a point percentage.
4. **Decoder panic surface (Low)**: `decode_short_range_best` rejects mismatched head/anchor lengths, truncated regressors, and a non-finite floor, with a test per malformed contract plus a well-formed control.

Re-runs on the final code: parity 512/512 both-ran, min IoU 0.999996, gate exit 0, skew self-test exit 0; probe 223 compared, r -0.938, exit 0; both CSVs byte-identical to the committed ones.

archledger:
Correction, measured after the review round: the shipped face_landmark.onnx emits all 478 points including iris (verified on a corpus frame; count printed by a throwaway harness). The earlier claim that it cannot feed the blendshapes model was wrong and is corrected in the doc, this PR body, and issues #315/#316. The real gap is that mesh_parity compares only the 468 shared-topology points, so the iris tail was never parity-checked between the meshes; that extension lands with #315.

archledger:
Landed on main as the rebased commit range 214f076..eedad0f, merged through the GitLab-primary flow (MR !3 at gitlab.com/archledger/irlume, four-stage pipeline green, Codex review round applied there). The review corrections recorded in this PR's history all carried over; closing rather than merging because the commits reached main with different SHAs after the rebase.

PR #318, opened 2026-08-07
Document recovery.key_present in the machine-API schema

### LINKED ISSUE #305: status --json emits recovery.key_present that the machine-API schema does not allow (strict conformance fails off-CI)
Found while running the CI gate for #304 on a machine with real irlume state.

`scripts/machine-api-conformance.py --strict` fails on:

```
capability status-json -> irlume status --json: schema: data/recovery: Additional properties are not allowed ('key_present' was unexpected)
```

The failure reproduces identically against a main-built binary (471cf79), so it predates #304. It does not appear in CI because the runner's environment never produces a recovery object carrying `key_present`; the strict schema pass only meets that field on a machine with an enrollment and recovery state, which is exactly the machine a support tool would query. So the green CI lane is testing a narrower response shape than the field produces, which is the 0.9.0 lesson again: if this contract were broken for real users, no job would go red.

Two-part fix, in whichever order the schema policy prefers:
1. Document `key_present` in the MACHINE-API schema for `data/recovery` (it ships in 0.9.0, so the contract should say so), or stop emitting it if it was unintentional.
2. Extend the conformance fixtures with a captured response from a machine with enrollment and recovery state, so the strict pass exercises the full shape.

archledger:
Closes #305. The field has been documented prose-side in MACHINE-API.md since 0.9.0 and emitted by status --json; only the JSON schema never gained it, so --strict conformance failed wherever the daemon answers with real recovery state. Hosted CI never had such an environment; the self-hosted lane surfaced it on its first full run of main. The schema now carries the field with the orphaned-store semantics, additive within contract 1. Strict conformance passes 36/36 locally against the real binary.


archledger:
Review round applied: the schema and MACHINE-API.md wording contradicted the implementation (a missing key with passphrase_set=true is repaired by irlume recovery restore, per restore_from_recovery and the TPM lifecycle test), both corrected; the running-daemon fixture now exercises key_present, and the guard is proven to fire both ways (pre-PR schema: 34 passed 2 failed including the fixture; fixed schema: 36/36).

PR #319, opened 2026-08-07
Preflight: pin the MSRV toolchain the header already claims

archledger:
The preflight job ran the runner's rolling pacman rust despite its own header describing a pinned MSRV toolchain; clippy 1.97's question_mark lint under -D warnings failed it on code the pinned gate accepts, the same drift that failed the GitLab lane's first pipeline. ghrunner's rustup 1.88.0 now leads GITHUB_PATH with a guard step that refuses anything but the pin.

### DIFF
--- a/.github/workflows/preflight.yml
+++ b/.github/workflows/preflight.yml
@@ -35,6 +35,14 @@ jobs:
     name: preflight · fmt · clippy · test (archhost)
     runs-on: [self-hosted, arch]
     timeout-minutes: 40
+    env:
+      # The ACTUAL pin (review round): PATH alone selects rustup's proxies
+      # but leaves the toolchain to rustup's mutable default; this override
+      # outranks default and directory overrides, and auto-install off keeps
+      # the job on the pre-provisioned runner instead of a silent network
+      # install.
+      RUSTUP_TOOLCHAIN: "1.88.0"
+      RUSTUP_AUTO_INSTALL: "0"
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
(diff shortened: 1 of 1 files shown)


archledger:
Review round applied: RUSTUP_TOOLCHAIN=1.88.0 at job scope now pins the selection itself (outranks the mutable default per rustup's override hierarchy), RUSTUP_AUTO_INSTALL=0 keeps the job on the pre-provisioned runner, and the verify step prints the full toolchain identity. The PATH-only version would have failed every preflight after a runner-local default change.

archledger:
Correction to this PR's stated root cause, found by a runner-fleet audit. I wrote that preflight was running "the runner's rolling pacman rust". The real mechanism is worse and more specific: ghrunner's rustup default toolchain is **stable**, not the 1.88.0 I installed, because the dtolnay/rust-toolchain action in ci.yml (jobs at ci.yml:257 and :324, plus asan.yml:94 for nightly) runs `rustup default` and therefore MUTATES the persistent toolchain state of the self-hosted runner user on every run. Verified: `rustup toolchain list` for ghrunner shows stable as active/default with 1.88.0 merely installed alongside.

That makes any job relying on the rustup default hostage to whichever job ran last on that machine, which is a hazard that simply does not exist on ephemeral hosted runners. The fix in this PR (RUSTUP_TOOLCHAIN at job scope) is still exactly right, and is now the only thing standing between preflight and that shared mutable state; the merge commit will carry the corrected explanation.

archledger:
Superseded by #324, which deletes preflight.yml entirely. The pin this PR added was the right fix for a lane that no longer exists: preflight duplicated ci.yml step for step once both ran on the same machine, and its unique real-TPM coverage moved into the check job. The finding behind it survives and is recorded on #324: the dtolnay/rust-toolchain action runs `rustup default` and therefore mutates a self-hosted runner's persistent toolchain state, so no job should trust the rustup default.

Issue #315, opened 2026-08-06
Switch the landmarks stage to the native 478-point .tflite mesh


archledger:
Switch the landmarks stage from the shipped `face_landmark.onnx` (a 468-point conversion) to Google's published `face_landmarks_detector.tflite` (478 points, sha256 `c7d54204ce0448474c7f3fa9af494787c0965cbdd6f20fc72867e43046bd43d5`, from the face_landmarker.task pinned in mesh_parity) on the bundled TFLite runtime.

The decision evidence is complete (docs/pad-results/2026-08-06-mediapipe-native-benchmark.md):

- **Fidelity**: the two models agree at mean NME 6.9e-7, worst 1.5e-6, over the stage-3 corpus (#306). The mesh_parity harness is an enforcing gate and becomes the acceptance test for this switch.
- **Latency**: native at 2 threads beat the ONNX mesh in both bench runs on the Zenbook 258V (5.79ms vs 10.34ms mean in the committed run; the ort mean moved between runs, the tflite mean did not), with a tighter p95 (2026-08-06-mp-latency-zenbook.csv). 2 threads matches the FullRangeBlaze precedent.
- **Supply chain**: deletes a conversion step that cannot be reproduced from the shipped artifact alone; ships Google's file byte-pinned under the same `TfliteSession::from_pinned_bytes` discipline as full-range Blaze.

Scope:

- `FaceMesh` grows a TFLite-backed path (or a sibling type) behind the existing pin; every consumer keeps reading (x, y) in frame pixels via `map_checked_mesh_output`, which both runtimes already share.
- The artifact joins the model set on every lane (Copr spec, PKGBUILD, PPA, .deb, Nix; fetch script + SHA256SUMS). The 468-pt ONNX and its `.legacy-192` sibling retire once the switch survives a release.
- Re-run the mesh_parity gate at the production thread count before shipping: the recorded parity runs were single-threaded, and multi-thread determinism of the native runtime is unverified (named in the benchmark doc's not-measured list).
- EAR indices, consent-gesture cues, and alignment all read indices below 468, so the extra iris points change no existing consumer.

[...]

PR #320, opened 2026-08-07
Package the tflite mesh the spec already installs

archledger:
The native-mesh switch (#315) added Source7 and its buildroot install but never extended %files beyond the *.onnx glob, so Copr builds fail with installed-but-unpackaged on face_landmarks_detector.tflite. First surfaced by the Packit run on #318, the first RPM build since the sync. The %files glob now covers *.tflite; rpmspec parse and packaging parity pass locally, and the Packit lanes on this PR are the real proof.


archledger:

archledger:

PR #321, opened 2026-08-07
Per-process temp dirs in tests, closing the shared-/tmp collision

### LINKED ISSUE #317: TPM-gated tests seal into fixed /tmp paths and collide across CI users on a shared runner host
arm_and_unseal_roundtrip (and siblings in the TPM-gated set) hardcode paths like /tmp/irlume-kr-rt for IRLUME_KEYRING_DIR. On archhost, which now hosts BOTH the GitHub preflight runner (user ghrunner) and the GitLab runner (user gitlab-runner), whichever user runs first owns the directory; the other user's run fails with seal: Io("Permission denied") because the leading remove_dir_all is a `let _` and the failure is silent. The first panic then poisons ENV_LOCK and four more tests fail as PoisonError cascades, which is exactly how it presented on the first GitLab hardware-stage run (pipeline 2738433939).

Fix shape: derive the temp dir from std::env::temp_dir() plus the uid or a per-process unique suffix, and stop discarding the remove_dir_all error when the directory exists but belongs to someone else (that case deserves a named panic, not a mystery EACCES downstream).

Infrastructure mitigation already in place: the gitlab-runner service runs with PrivateTmp=true, so its jobs see a private /tmp and the collision cannot recur there. The repo fix is still worth it: any two users running the suite on one machine hit this, and /tmp is world-shared by design.

archledger:
Closes #317. Twenty-six hardcoded /tmp/irlume-* paths predated the test_tmp_dir helper that already appends the process id. With two CI users on one host (archhost runs both the GitHub and GitLab runners), whichever ran first owned the directory and the other failed with EACCES on seal; the first panic poisoned ENV_LOCK and five more tests cascaded as PoisonError. It took down the GitLab hardware stage and then the GitHub coverage job, on code neither PR touched.

All paths now derive from test_tmp_dir, so two users, two runners, or two concurrent processes cannot collide. Verified locally: the four TPM-gated tests that were failing pass against the real TPM, full irlume-core lib suite 120/120.

### DIFF
--- a/crates/irlume-core/src/keyring.rs
+++ b/crates/irlume-core/src/keyring.rs
@@ -541,10 +541,10 @@ mod tests {
     #[test]
     fn envelope_path_under_keyring_dir() {
         let _g = crate::testenv::ENV_LOCK.lock().unwrap();
-        std::env::set_var("IRLUME_KEYRING_DIR", "/tmp/irlume-kr-test");
+        std::env::set_var("IRLUME_KEYRING_DIR", crate::test_tmp_dir("kr-test"));
         assert_eq!(
             envelope_path("alice"),
-            PathBuf::from("/tmp/irlume-kr-test/alice.json")
+            PathBuf::from(format!("{}/alice.json", crate::test_tmp_dir("kr-test")))
         );
         std::env::remove_var("IRLUME_KEYRING_DIR");
     }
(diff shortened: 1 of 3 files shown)


archledger:
Review round applied: the uid now leads the temp-path suffix, so a pid reused after wrap can only ever hit our own stale directory (which is what makes the callers' ignored remove_dir_all safe rather than a hidden failure). Generated shape verified live: /tmp/irlume-test-probe-u1000-p366663. Full irlume-core lib suite 120/120, TPM-gated tests pass against the real TPM.

archledger:

archledger:

PR #322, opened 2026-08-07
Share the light CI jobs across both self-hosted runners

archledger:
minihost (Intel N100, CachyOS, NexiGo attached, headless, Tailscale-reachable) joins archhost as a second runner. Both now carry a shared `ci` label; cargo-deny and the systemd units job move to it so either machine can run them.

The split is deliberate. Heavy jobs (build+test, stable, coverage, fuzz, nix) stay on `arch`, which only the Ryzen carries: an N100 would stretch a 4-minute release build considerably and make CI latency depend on which machine happened to be free. And sharing beats pinning for these two, because both are required checks and a required check bound to one machine makes that machine a single point of failure for every merge.

minihost is also provisioned for hardware work (TPM, ir-camera labels, v4l2loopback nodes 8/9/10), so the hardware-suite workflow can already land on either box.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,12 @@ concurrency:
   group: ci-${{ github.ref }}
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
+# Runner routing: heavy jobs (build, test, coverage, fuzz, nix) stay on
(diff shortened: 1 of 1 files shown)


archledger:

archledger:

PR #323, opened 2026-08-07
Drop the GitLab lane; GitHub is the only platform

archledger:
Retires the GitLab standby by decision: one platform, one set of runners, less to maintain. Removes .gitlab-ci.yml and the mirror workflow. Already done outside the repo: the GitLab project is deleted, its runner is unregistered and the package removed from archhost (reclaiming 9GB), the GITLAB_MIRROR_TOKEN secret is revoked, and the local gitlab remote is gone.

Verified before deleting anything: every file on the GitLab-only branch already existed on GitHub main (the commits differed only because they had been rebased), all 41 tags matched, and no merge requests were open.

Beyond simplification this removes a genuine hazard. archhost hosted two runners under two different users that could run jobs at the same time, which is exactly what turned the hardcoded /tmp paths (#317) into an EACCES cascade and then into a TPM dictionary-attack lockout that failed unrelated jobs on unrelated branches. One runner per machine makes that class of failure impossible.


archledger:

archledger:

PR #324, opened 2026-08-07
Delete the duplicate preflight lane and fix what the CI audit found

archledger:
A workflow audit measured what a single push costs on the self-hosted fleet: **10 jobs, 265 timeout-minutes serialized on one machine**, with `cargo test --workspace` running five times and the workspace compiled seven times across five toolchain/flag combinations. Roughly 155 of those minutes were re-runs of identical work. That queue length, not a misconfigured concurrency group, is what produced the cancelled runs.

**preflight.yml is deleted.** It existed to give fast feedback before a GitHub-hosted gate; that gate now runs on the same machine, so it duplicated ci.yml step for step while its header still described the old topology. Its only unique coverage was the REAL TPM set (ci.yml had swtpm only), which moves into the check job where the workspace is already built, so nothing is lost.

Other audit findings fixed here:

- **The camera readiness poll could not fail.** The nodes are chmod 666 before the feeders start, so `[ -r /dev/video8 ]` always succeeded on the first iteration. It now waits for an actual frame from each node and fails loudly if a feeder dies.
- **hardware-suite could test no camera and pass.** It routed on `[self-hosted, tpm, ir-camera]`, labels minihost also carries, and its camera-absent path exits 0 — so the only real-IR-camera lane could land headless and report green. Pinned to `arch`.
- **The required-check list was wrong**: the header documented four, seven are enforced, and it described two of those as not required. Corrected against the live ruleset, with a note that five of the seven are archhost-only.
- **actionlint was silently half-blind**: without shellcheck it drops every shell finding. shellcheck is now installed on both runners and the stale comment claiming a hosted runner provides it is corrected.

Verified locally: all eleven remaining workflows parse, shellcheck reports zero errors across every inline `run:` block, and runner routing is as intended.


archledger:

archledger:

PR #325, opened 2026-08-07
CI integration: land the five interdependent fixes together

archledger:
Five open PRs each failed only because they lacked the others' fixes, a deadlock that would have taken five sequential CI cycles to unwind on a single-runner fleet. This lands them as one verified change; the individual PRs close pointing here.

**What it contains**

- **#321** per-process temp dirs in tests (uid+pid), closing the shared-/tmp collision that cascaded into EACCES, poisoned ENV_LOCK, and then tripped the TPM's dictionary-attack lockout, failing unrelated jobs on unrelated branches.
- **#320** package the tflite mesh the spec already installs (Copr proved both Fedora lanes).
- **#322** share the two light jobs across both runners via a `ci` label, heavy jobs stay pinned to `arch`.
- **#323** drop the GitLab lane; GitHub is the only platform.
- **#324** delete the duplicate preflight lane (absorbing its unique real-TPM coverage into the check job), make the camera readiness check actually prove a frame arrives, pin hardware-suite to the machine with the real camera, and correct the required-check list from four to the seven actually enforced.

**Why together**: #324 deletes preflight.yml, which made #319's toolchain pin moot (closed as superseded), and #322 and #324 both touch ci.yml's header. Merged in dependency order with one conflict resolved by hand, keeping the runner-routing note and replacing the stale four-check list with the verified seven.

**Verified locally**: eleven workflows parse, shellcheck reports zero errors across every inline `run:` block, `cargo fmt --check` clean, irlume-core lib suite 120/120, machine-API strict conformance 36/36, `rpmspec -P` clean, packaging parity clean. Net 390 lines removed against 134 added.


archledger:
Review round applied. CRITICAL accepted and fixed: the real-TPM step no longer selects seal_unseal_pcrlock_roundtrip_provisioned_nv, which hard-codes NV index 0x0181C0DE and undefines whatever occupies it with no ownership check. I verified that in the source (tpm.rs:1689 and the unconditional nv_undefine_space at 1714-1718) rather than taking it on trust: against the real chip that destroys persistent hardware state the test never created, and an interrupted run leaves the fixture for the next run to delete. Its pcrlock coverage stays on swtpm, where the whole TPM state is disposable.

HIGH accepted: uid+pid does not close same-user pid reuse, so the identity now includes the boot id and the process start time (a surviving directory can then only belong to a live process), and the /proc reads fail loudly instead of collapsing to a sentinel uid. Both LOWs fixed: the feeder liveness check watched the IR pid while naming the RGB one, and a per-job comment contradicted the header's required-check list. Core suite 120/120, workflows parse, shellcheck clean.

archledger:
Routing reworked after measuring two facts: this repo is public, so hosted runners are free, unmetered and PARALLEL; and self-hosted runners are not an outage hedge, since archhost's lane failed at "Set up job" during the 2026-08-06 incident exactly like the hosted lanes (they all need the Actions control plane to receive a job).

So routing is now by capability rather than availability. Everything that needs no real device is hosted and runs concurrently. The two machines keep only what they alone can do: real-TPM and pinned-TFLite-runtime tests in a new hardware-checks.yml on the shared `ci` label so either machine takes them, and the real IR camera in hardware-suite on archhost. A side effect worth having: with every ci.yml job hosted, the pull_request trigger returns, so fork PRs get the full required gate again rather than needing a maintainer to push the branch.

archledger:
Correcting my own diagnosis: the check and coverage cancellations were NOT a hosted-runner capacity problem, and the timeout raises I pushed were treating a symptom. Per-step timings from the cancelled jobs: build 72s, test 144s, swtpm TPM tests 31s, and `Start virtual cameras` **1501s** (1867s in coverage's copy) — the entire budget in one step.

The cause was the readiness probe I added earlier tonight. An ffmpeg frame read against a v4l2loopback node with no producer blocks forever, and it sat in a retry loop, so any job that could not get a frame burned its whole budget and was cancelled at exactly 30:00. Each probe is now bounded by `timeout 5`, which keeps what the fix was for (proving a frame actually arrives, which `[ -r /dev/video8 ]` never did) without the hang. The 70 and 75 minute budgets are reverted to 30.

archledger:
Followed the no-more-guessing instruction with measurements on real hardware rather than another timeout bump.

Three states of a v4l2loopback device, measured with v4l2loopback 0.15.4:

| state | behaviour |
|---|---|
| no writer attached | ffmpeg fails immediately (exit 251, 0s) |
| writer attached, healthy | frame obtained on the first probe |
| writer attached, STALLED | blocks in the kernel; plain `timeout 8` left it alive in futex_wait **8+ minutes** later, timeout process still waiting |

So the CI hang was never "no producer" and never a capacity problem. It is a writer that attaches but does not deliver, and SIGTERM cannot reach a task blocked in a V4L2 ioctl, which is why my first `timeout 5` fix would still have hung. `timeout -k 2 5` adds the SIGKILL it cannot ignore: re-measured, the stalled case returns exit 137 in 7s with no stray processes, and the healthy case still obtains a frame. The 70/75-minute budgets I set on the wrong theory are reverted to 30.

archledger:
Correcting a regression I introduced two commits ago, found by re-running the research against the pushed code rather than trusting it.

15c442c deleted the /dev/video9 IR feeder while removing stale pid captures: both surviving nohup lines fed video8, while the readiness gate still looped over video8 AND video9. video9 could therefore never get a producer, its sysfs state stayed `output` forever, no read was attempted, and wait_for_frame burned its full 30s deadline on every run. The job timing matches to the second: 76s = ~45s module build + ~1s for video8 + exactly one 30s deadline for video9.

Two smaller defects from the same commit are fixed with it: the failure branch tailed feeder logs that were still redirected to /dev/null, so the diagnostics it claimed to restore were still discarded; and wait_for_frame's variables were not declared local. Verified on hardware with both feeders and the real helper: both devices ready in 1s.

Note this is NOT the original 117s hosted failure, which ran the previous loop with both feeders present. That remains unexplained, and four reproduction environments (including a VM on the runner's exact 6.17.0-1021-azure kernel, which gets a frame in 45ms) rule out kernel, ffmpeg, userspace, module commit and load parameters. The restored diagnostics are how we settle it.

archledger:
Caught a second self-inflicted defect before merging, this one silent: the block replacement that synced coverage's camera step to the check job's copy swallowed the `deny` and `units` JOB DEFINITIONS that sat between them. They disappeared from the workflow, so neither ran, and branch protection blocked the merge on two required contexts that could never report.

Worth noting how quiet that failure is: GitHub does not flag a required check that is never created. The board looked entirely green (16/16) while two required contexts were simply absent, and `gh pr merge` refused with BLOCKED rather than naming them. Both jobs are restored verbatim from main and routed hosted with the rest; verified that all seven required contexts are defined, both camera gates survived the edit, the units job still carries its apparmor steps, and shellcheck is clean.

PR #353, opened 2026-08-07
chore: enforce the documentation lints that were never running

archledger:
## What this changes

`missing_errors_doc`, `missing_panics_doc` and `undocumented_unsafe_blocks` are
all `allow` in stock clippy, and this workspace had no `[workspace.lints]` and
no `clippy.toml`. None of the three had ever fired, so 252 sites across 45 files
carried no documented panic, error condition or unsafe precondition.

All three are now `warn` at workspace level, with each of the twelve member
crates opting in through `[lints] workspace = true`. That opt-in is required:
without it no workspace lint applies at all.

## Why the existing sites use `#[expect]`

CI runs `cargo clippy --all-targets -- -D warnings` and the tree was otherwise
clean, so switching the lints on without a plan would have turned the build red
on all 252 at once.

Each remaining site is marked `#[expect(...)]` instead of being given an
invented docstring. A plausible but unchecked `// SAFETY:` comment on FFI in a
daemon that authenticates people is worse than a visible gap, so the debt is
marked rather than papered over.

`#[expect]` self-cleans: documenting a marked site makes its expectation
unfulfilled, and `unfulfilled_lint_expectations` is `warn` by default, so
`-D warnings` rejects the change until the attribute is removed. The backlog can
only shrink. This is the migration path the Rust 1.81 release notes describe for
this exact case.

## Unsafe expectations are per-expression, and that is load-bearing

A single `#[expect]` is fulfilled by one matching diagnostic anywhere inside the
item carrying it. A function-level attribute would therefore let a later
undocumented unsafe block in that same function ride along on an
already-fulfilled expectation and reach CI unflagged.

The first revision of this PR had exactly that hole. Review caught it, and every
unsafe expectation now sits on the individual unsafe expression. Proved by
planting the failure: adding an undocumented `unsafe { libc::getpid() }` inside
`password_matches_login`, which already carries one expectation, is now rejected.

21 of the 95 unsafe sites turned out to be credential getters with no
preconditions for the caller (`getuid`, `geteuid`, `getgid`, `getegid`,
`isatty`, `umask`, and `mkfifo` on a live `CString`). Those carry real safety
comments rather than debt markers. 74 remain marked.

## The five panic sites

Only `estimate_similarity` can actually panic, on its `assert_eq!` over
mismatched slice lengths, and it gets a real `# Panics` section. The other four
are guarded and cannot panic, so a `# Panics` section would have stated
something untrue. Each `#[expect]` on those names the guard instead:

- `contrast_signature`: every `unwrap` sits behind the `ear.is_some_and(..)` filter
- `capture_ear_samples`: the `mesh.is_none()` early return
- `enroll_profile`: `target` comes from the same profile list the lookup searches
- `capture_ir_sequence`: the single `take()` reopens the stream in the same branch, and both failure paths in between return early

## Verification

Run on the 1.88 MSRV and again on stable 1.96, both green:

| check | result |
| --- | --- |
| `cargo clippy --all-targets -- -D warnings` | pass |
| `cargo fmt --all -- --check` | pass |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --locked` | pass |
| `cargo test --workspace` | pass |

Three gate directions were proved by planting the failure rather than assuming:

1. a new undocumented `pub fn` returning `Result` is rejected
2. documenting a marked site is rejected until its attribute is deleted
3. a new undocumented unsafe block inside an already-marked function is rejected

The 15 deletions are rustfmt expanding `unsafe { … }` one-liners that now carry
an attribute. No executable logic changed.

## What this does not do

It does not write the remaining docs. The ratchet stops the backlog growing;
clearing it is separate work. Hardware lanes were not exercised locally, though
nothing here touches a code path that reaches hardware.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,28 @@ license = "GPL-3.0-or-later"
 repository = "https://github.com/archledger/irlume"
 authors = ["archledger <archledger236@gmail.com>"]
 
+# Documentation lints for a crate that authenticates people. All three are
+# `allow` in stock clippy, so the coding standard requiring documented panics,
+# error conditions and unsafe preconditions was never actually enforced.
+#
+# The pre-existing sites carry `#[expect(...)]` rather than a hand-waved
+# comment, so CI stays green while NEW code is gated from today. `#[expect]`
+# self-cleans: once a site is documented the expectation goes unfulfilled and
+# `-D warnings` fails until the attribute is removed, so this backlog can only
+# shrink. Each member crate opts in with `[lints] workspace = true`, which is
+# required; without it none of these apply.
+#
+# Unsafe debt is expected on each INDIVIDUAL unsafe expression, never on the
+# enclosing function. One `#[expect]` is fulfilled by one matching diagnostic
+# anywhere inside the item it sits on, so a function-level attribute would let
+# a later undocumented unsafe block in that same function ride along on an
+# already-fulfilled expectation and reach CI unflagged. Per-expression
+# placement is what makes a new unsafe block a build failure.
+[workspace.lints.clippy]
+missing_errors_doc = "warn"
+missing_panics_doc = "warn"
+undocumented_unsafe_blocks = "warn"
+
 # Shared dependency versions. Heavy/native deps are listed but commented so the
 # scaffold compiles offline as pure stubs; uncomment per crate as you implement.
 [workspace.dependencies]
--- a/crates/irlume-auth/Cargo.toml
+++ b/crates/irlume-auth/Cargo.toml
@@ -18,3 +18,6 @@ irlume-core.workspace = true
 
 [dev-dependencies]
 serde_json.workspace = true
+
+[lints]
+workspace = true
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -806,6 +806,7 @@ mod capture_mode_decision_tests {
 }
 
 impl Engine {
+    #[expect(clippy::missing_errors_doc, reason = "doc backlog")]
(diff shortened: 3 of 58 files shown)


archledger:
Review round found one merge blocker, now fixed.

**The defect.** Every `undocumented_unsafe_blocks` expectation sat on the
enclosing function. A single `#[expect]` is fulfilled by one matching diagnostic
anywhere inside the item carrying it, so the two existing unsafe expressions in
`password_matches_login` fulfilled its attribute and a third undocumented block
added to that function later would have been suppressed too. The gate would have
caught new unsafe code in unmarked functions while silently accepting it in the
61 marked ones, which is the wrong half to miss in a privilege-dropping and PAM
path.

**The fix.** Every unsafe expectation moved to the individual unsafe expression.
Where the attribute could not be placed on an expression, the site got a real
safety comment instead, which is why 21 of the 95 unsafe sites are now genuinely
documented rather than marked: they turned out to be credential getters with no
preconditions for the caller (`getuid`, `geteuid`, `getgid`, `getegid`,
`isatty`, `umask`, and `mkfifo` on a live `CString`). 74 remain marked.

**Proof the hole is closed.** Adding `let _probe = unsafe { libc::getpid() };`
to `password_matches_login`, which already carries one expectation, now fails
the build with `unsafe block missing a safety comment`. Reverted after the
check.

Re-verified on the 1.88 MSRV and stable 1.96 after the change: clippy with
`-D warnings`, `fmt --check`, `cargo doc` with `-D warnings`, and
`cargo test --workspace` (1261 passed, 0 failed, 70 ignored).

One other point from the round was fair and is corrected in the description: the
earlier claim that the standard had been in force "since July" could not be
verified from anything in this repository, so it has been dropped in favour of
the checkable facts about what the lints were set to.

PR #371, opened 2026-08-08
An IR exposure that cannot be measured is refused, not passed (#358)

### LINKED ISSUE #358: The IR exposure gate is disabled on every IR camera that is not Grey8
Found by a codebase audit, then confirmed by reading each site.

`exposure_refusal` in `crates/irlume-liveness/src/lib.rs:651-653` collapses "not measured" into "measured and clean":

```rust
cues.ir_exposure_ok = s
    .ir_saturated_frac
    .is_none_or(|f| f <= IR_SATURATED_FRAC_MAX);
if cues.ir_exposure_ok {
    return None;
}
```

`ir_saturated_frac` is `None` whenever `clipping_white_level` returns `None`, and `crates/irlume-camera/src/lib.rs:1933` returns `None` for `IrPixel::Grey16 | Nv12Luma | YuyvLuma`. Those cover every IR node negotiating `Y16`, `Y10`, `Y12`, NV12 or YUYV, all of which are in `IR_CANDIDATES` and are supported hardware. So on that whole class of camera the #237 exposure gate is not merely lenient, it is off.

The rule this breaks is stated in three places in this repo and implemented correctly in a sibling function. `crates/irlume-camera/src/ir_metadata.rs:344` says it outright: "A missing per-frame entry is treated as fully clipped, never as clean: failure to observe must not authorize", and line 371 does `clipped.get(i).copied().unwrap_or(1.0)`.

Both mitigations disappear together on the same cameras, so nothing else catches it. `best_gate_frame` has no `clipped_fracs` either, so it falls back to `brightest_lit` and selects the most saturated frame; then `exposure_refusal` waves it through. #221 measured 78.8% of pixels at the ceiling on a close emitter, and the center/edge ratio on a blown frame was measured at 1.11 to 1.19 against a `MIN_CENTER_EDGE_RATIO` of 1.03. A frame carrying no facial structure passes.

The fix is one word, `is_some_and`, but it is not only a typo fix: it changes those cameras from always-pass to always-refuse, so it needs a decision about whether they belong at Convenience tier instead. That decision is the reason this is an issue and not a pull request.

Second consequence worth recording: `crates/irlume-cli/src/pad.rs:221` writes `cues.ir_exposure_ok` into the PAD corpus. Frames the gate never read are recorded as `ir_exposure_ok: true`, so the corpus used to tune these thresholds is contaminated on exactly the cameras where the question is hardest.

Not tested: no hardware run was made against a Y16 IR node. The finding is from reading the three sites and the format table, and the measurements quoted are the ones already recorded in #221.

archledger:
Closes #358.

`exposure_refusal` read "could not measure" as "clean":

```rust
cues.ir_exposure_ok = s.ir_saturated_frac.is_none_or(|f| f <= IR_SATURATED_FRAC_MAX);
```

`ir_saturated_frac` is `None` whenever the negotiated format defines no sensor ceiling, which is every IR node on Y16/Y10/Y12/NV12/YUYV. On those cameras the #237 exposure gate was not lenient, it was **off**, and every cue below it ran on a frame nobody had checked. This repo states the opposite rule in three places and implements it correctly in `ir_metadata.rs`: failure to observe must not authorize.

## Why not the one-word fix

`is_some_and` alone is wrong. `saturated_frac_of` also returns `None` when no **face** was found, so flipping the combinator would refuse every faceless frame on every camera, including the GREY ones that work today. The evaluators happen to require an IR face before reaching the gate, so the two cannot be confused there right now, but that is an invariant in a different crate that nothing pins.

The gate now states its own precondition instead: `Signals::ir_ceiling_known`, set from `ir_stats.white_level.is_some()`, defaulting to **false**. The default direction is deliberate; a permissive default is what the bug was.

## Refused, not tiered

Dropping these cameras to a lower assurance tier that still releases credentials would be the same fail-open with better paperwork. The honest outcome for "the gate could not run" is a refusal.

## The part that was easy to get wrong

The refusal is **not presence-retryable**. It arrives as `Verdict::Uncertain`, which is the retryable class, and that is the trap: the condition is a property of the camera and identical on every frame. Retrying would spend the entire grace window reaching the same answer, then fall back to the password anyway, while telling the user to move back, which cannot help. It is routed out by reason prefix in `liveness_deny_kind`, the same mechanism `no face in IR` already uses, and the message deliberately does not say "move back".

## The corpus fix, which stands on its own

`Cues::ir_exposure_measured` is now recorded beside `ir_exposure_ok`, so a row the gate never read is distinguishable from a row it read and found clean. Every retune of `IR_SATURATED_FRAC_MAX` was being fitted on data that claimed "clean" for unread frames, on exactly the cameras where the question is hardest to see.

## Blast radius, measured

Both IR cameras in the fleet negotiate `GREY`: Zenbook `/dev/video2` and archhost's NexiGo `/dev/video2`, checked with `v4l2-ctl --list-formats`. `IR_CANDIDATES` is walked in order with `GREY`/`Y8`/`Y800` first, so **any** camera offering 8-bit grey is untouched. Only a camera with no grey path at all changes behaviour, and it now gets a refusal naming the problem instead of a silent pass.

## Evidence

Both guards proven by mutation:

- restoring `is_none_or` and deleting the unmeasurable branch fails `an_unmeasurable_ir_format_is_refused_on_every_credential_releasing_path`
- deleting the routing arm fails `an_unmeasurable_exposure_is_not_retryable`

The retryability test also asserts a blown-out frame stays retryable, so this did not sweep the ordinary case out with the new one.

Gate: fmt, clippy `-D warnings`, doc lane, and the four touched crates green (`irlume-liveness` 51, `irlume-auth` 72, `irlume-cli` 34, `irlume-vision` 69).

## What was not done, and one thing to know about my local run

Implementing the ceiling for Y10/Y12 (raw maxima 1023/4095) and NV12/YUYV (via quantization) is achievable and is the better long-term answer. We own no camera in any of those formats, and writing a saturation ceiling that cannot be validated is how a gate goes wrong in the other direction. The refusal message asks for the camera to be reported, which is what would supply one. Generic Y16 may stay unmeasurable permanently, since V4L2 allows its real precision to be unknown.

The dev `liveness_probe` now stops at this gate, because it captures a single frame with no burst stats and never had a ceiling. Its measured values still print from its own locals; what it no longer does is claim a liveness verdict it was not in a position to make.

Local full-workspace `cargo test` currently fails 9 `irlume-vision` model tests in my checkout with `Io("No such file or directory")`. That is **not** from this change: it reproduces identically on `main`, only in the concurrent workspace run, and each crate passes on its own. CI runs the same command and is green on main, so it is local to this machine and I could not reproduce it per-crate. Flagging it rather than quietly working around it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,28 @@ All notable changes to irlume are documented here. This project adheres to
 
 ### Security
 
+- **The IR exposure gate was off, not lenient, on every IR camera that is not
+  8-bit grey.** `exposure_refusal` read "could not measure" as "clean", so on a
+  node negotiating `Y16`/`Y10`/`Y12`/`NV12`/`YUYV` the #237 gate never ran and
+  every liveness cue below it judged a frame nobody had checked.
+  `clipping_white_level` answers `None` for those formats, and `is_none_or`
+  turned that into a pass. The gate now states its own precondition,
+  `Signals::ir_ceiling_known`, defaulting to false, and a format that names no
+  ceiling is refused rather than passed: a gate that could not run is not a
+  gate that passed.
+
+  The refusal is not presence-retryable on either evaluator. It is a property
+  of the camera, identical on every frame, so retrying would spend the whole
+  grace window reaching the same answer while advising something that cannot
+  help; the message deliberately does not say "move back".
+
+  Blast radius, measured: `IR_CANDIDATES` tries `GREY`/`Y8`/`Y800` first, so any
+  module offering 8-bit grey is untouched. Every IR camera in the project's
+  record negotiates GREY, including both user-reported ones, and no camera
+  without a grey path has ever been reported. Such a camera can no longer
+  enroll or authenticate, and `docs/PLATFORMS.md` now says so instead of
+  listing it as untested.
+
 - **One socket request reached the keyring with an unscreened username.**
   The daemon screens the account name in every request against path
   traversal before it builds a `<user>.json` path, but the guard read a
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -2714,11 +2714,20 @@ impl Engine {
                 // Dark-path kinds: Uncertain retries under grace, any Spoof
                 // does not (the retryable RGB-yes/IR-no transient cannot occur
                 // here: this path only runs when RGB saw no face).
-                let kind = if verdict == Verdict::Uncertain {
-                    OutcomeKind::Uncertain
-                } else {
-                    OutcomeKind::Spoof
-                };
+                //
+                // Routed through the shared classifier rather than mapped
+                // inline. `exposure_refusal` is deliberately shared by BOTH
+                // evaluators, so the unmeasurable-format refusal arrives here
+                // as Uncertain too, and an inline map would leave it in the
+                // retryable class on exactly the camera this gate exists for:
+                // six full captures reaching the identical answer, every dark
+                // login, forever. The classifier holds the one prefix rule
+                // (#358 review).
+                //
+                // `reason` here is the raw liveness string; the "dark liveness"
+                // prefix is applied in the `format!` below, after this call, so
+                // the prefix match still sees what irlume-liveness produced.
+                let kind = liveness_deny_kind(verdict, &reason);
                 return Ok(Outcome::deny(
                     kind,
                     format!("dark liveness {verdict:?}: {reason}"),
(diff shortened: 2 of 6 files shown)


archledger:
Hardware validation, after merge. Zenbook, ASUS FHD IR module, `/dev/video2` GREY 640x400, merged main built at `325d7ae` and run as the release binary.

Three consecutive `liveness` probes, which is the path this change altered:

```
run 1  clipped 0.0%   glint 145  glint_readable=true   [GATE] Live
run 2  clipped 0.0%   glint 186  glint_readable=true   [GATE] Live
run 3  clipped n/a    glint n/a  glint_readable=false  [GATE] Spoof: no face in IR
```

Two things this establishes that the test suite could not.

**The ceiling is known on a real GREY camera, so the new refusal never fires here.** Runs 1 and 2 print a measured clipping fraction rather than `n/a`, which means `ir_stats.white_level` was `Some` and `ir_ceiling_known` was true. No run produced "IR exposure unmeasurable". That is the lockout risk this change carried, and it does not touch this module.

**Run 3 is the case the one-word fix would have broken.** No IR face, so `ir_saturated_frac` is `None` while the ceiling is still known. The verdict is "no face in IR", the retryable transient, not an exposure refusal. That is precisely the ambiguity `Signals::ir_ceiling_known` was added to separate: `is_some_and` alone would have refused this frame on a camera that measures perfectly well. Confirmed on hardware rather than argued from the source.

It also confirms the review-round fix to the probe itself. Before it, `ir_ceiling_known` was hardcoded false here, so all three runs would have printed three false cues and "this camera's IR format defines no sensor ceiling" about a camera whose ceiling is 255.

Not covered: no camera without an 8-bit grey path exists to test the refusal path itself, here or anywhere in the project's record.

Issue #340, opened 2026-08-07
Unmeasured capture-mode default should be Sequential, probed once at enrollment


archledger:
From the camera-stack research (~/irlume-research/2026-08-07-camera-stack-research/RESEARCH.md, kernel v6.16 sources cited): the unmeasured capture-mode default is Concurrent today, and the measured record prices that wrong. A wrong concurrent default cost an enrollment outright on the Brio (#308's history: STREAMON succeeds, zero RGB frames, QBUF EINVAL), and the NexiGo's silent 42-56% RGB retention hides in dim rooms as unexplained recognition weakness. A wrong sequential default costs 0.7 to 1.3 seconds of capture latency and nothing else.

Change: default Sequential when no measured verdict is stored, run the contention probe once at enrollment time (the natural moment: the user is present, the cameras are warm, and the verdict then governs exactly the shapes enrollment and auth use), and keep every capture path reading only the stored verdict as today.

Also in scope, same research: the failure signatures deserve telling apart in messages and dlog. Device-firmware refusal maps to EINVAL (uvcvideo maps a STALLed class request, uvc_video.c:148-150), while genuine bus exhaustion surfaces as EIO ("no fast enough alt setting") or ENOSPC (xHCI bandwidth admission). Both measured starving cameras are firmware refusals, not bus arithmetic, and the docs and diagnostics currently blur the two.

Explicitly not in scope, disproven by the same research: predicting dual-stream capability from descriptors (the bandwidth request is device-chosen at PROBE time, xHCI exposes no remaining budget, and both measured failures are device-internal; Microsoft's own stack uses declared concurrency tables instead of prediction, which is what cameras.conf already is with measurement in place of OEM declarations).

archledger:
Shipped in #342. The unmeasured default is Sequential with the measured cost argument in the code, enrollment runs the one-time contention probe for identified camera pairs with no stored verdict, only a fully-backed conclusive result persists (every requested round clean in both arms, stored under an exclusive config lock that yields to a concurrent writer), verdicts are keyed by the camera pair with legacy RGB-only entries honored solely for single-device pairs, and stream-start diagnostics now say only what each errno proves. Nothing about stored verdicts changed; camera-tune keeps its overwrite semantics.

Issue #100, opened 2026-07-26
Auto-calibrate the capture mode instead of requiring camera-tune


archledger:
## Problem

A camera fault that irlume can detect and work around currently requires the user to know a command exists. Most people will not run it. Issue #93 is the shape of it: face login stopped working after an upgrade, the reporter had no way to name the cause, and the exchange needed several rounds to get anywhere.

`irlume camera-tune` (added in the capture-mode work) has the same weakness. It measures whether a camera can stream both sensors at once and stores the answer, but nothing runs it, so a user with an affected camera gets a dim colour frame, worse recognition, and no reason to suspect the camera.

## What is already measured

On 2026-07-25, on two modules:

- The NexiGo HelloCam N930W stops tracking the scene while its infrared sibling streams. Its concurrent colour mean sat at 56-66 across four runs whose sequential arm ranged 62 to 143 with the room lighting. That is auto-exposure freezing, not proportional dimming.
- An ASUS FHD built-in measured in the same light kept 102%.
- Pairing the NexiGo's colour node with a *different* camera's infrared node kept 99%, so it is two interfaces of one module competing, not concurrency itself.
- No `uvcvideo` bandwidth error appears in `dmesg` for any of it. The hardware degrades silently.

A probe run in a dark room reports a clean result for an affected camera, because a scene that genuinely reads ~60 hides a fault that parks brightness at ~60. `camera-tune` reports that case as inconclusive rather than healthy.

## Proposal

**1. Trigger on real usage, not on a schedule.** The cross-spectrum self-heal in `Engine::assess_full` already fires exactly when this camera dimmed under concurrent capture, with a face present, in the light the user actually uses. Count those events per camera identity. After it fires repeatedly, switch that camera to sequential capture and log one line. No user action, no camera grab, and immune to the dark-room problem that defeats a probe.

**2. Let the probe run at enrol, and only there.** If a camera identity has never been measured, enrolment is the one moment the user is in front of the camera, lit, and already waiting. The streams are open by then anyway.

**3. Report what was decided.** `irlume doctor` should say which capture mode this camera uses and why, with the measurement date. Silent adaptation hides regressions: if irlume quietly works around a camera fault forever, nobody learns the camera is faulty, and a bug in our own code looks like normal behaviour.

## Rules this must not break

- **Never during an authentication.** Calibration holds the camera for tens of seconds. Firing it mid-login turns a working login into a failure, which is the exact class of problem this is meant to prevent.
[...]

PR #378, opened 2026-08-08
doctor: report which capture mode a camera pair uses and why (#100)

archledger:
## What & why

Part of #100.

A camera that dims when both sensors stream at once gets switched to sequential capture and works from then on, slower but correct. Nothing tells the user that happened. `irlume doctor` reported the emitter, the models, the TPM and the consent gate, but never the capture mode, so a camera silently parked in sequential mode looked identical to one that never needed it, and a bug in irlume's own capture path would read as normal behaviour. The issue makes this point directly: "Silent adaptation hides regressions... if irlume quietly works around a camera fault forever, nobody learns the camera is faulty."

doctor now names it, for the active camera pair:

- `concurrent` or `sequential`, measured for this pair, with a one-line reason;
- or the unmeasured `sequential` default, pointing at `sudo irlume camera-tune` to measure it;
- or, on an unprivileged run, that the verdict is root-only (in `cameras.conf`) and needs `sudo irlume doctor` to read, the same shape as the existing credential-release block.

The pair is read with `configured_pair_no_probe`, so doctor still opens no camera. Every case is `Info` or `Unknown`, never `Warn`/`Fail`: a capture mode is a strategy, not a fault, and the slower sequential mode is the correct choice for a camera that dims under concurrent capture. The wording of each case is a pure function (`capture_mode_report_line`) with a test; the block that gathers the observation reuses the existing `stored_capture_mode` reader.

## Scope

This is the "report what was decided" part of #100 (item 3). The other two parts, triggering the switch automatically from repeated concurrent-dim self-heals and running the probe at enrolment, set behaviour on the auth path and depend on a switch threshold and the on-hardware validation the issue asks for, so they are left for a separate change rather than guessed at here. The maintainer's own note in the issue is that pointing failures at doctor "may be worth more than more silent machinery," which is the part this delivers.

Because the mode is currently stored without a "why" or a measurement date, doctor reports the provenance it can derive (measured vs unmeasured default) rather than a stored date; carrying the date is part of the auto-calibrate storage in the deferred parts.

## Testing

Container-run.

- Added `capture_mode_report_names_the_mode_and_never_warns`, which checks each case: a measured mode is `Info` and names the mode and that it was measured; the unmeasured case names the default and points at `camera-tune`; the root-required case is `Unknown` and says to use sudo; the no-pin case names auto-selection.
- `irlume-cli`: 390 tests pass.
- `cargo fmt --check`, `cargo clippy -p irlume-cli --all-targets`, and `RUSTDOCFLAGS="-D warnings" cargo doc -p irlume-cli --no-deps` are clean.

One pre-existing failure in this container is unrelated and fails identically on `main`: `tui::tests::models_screen_unprivileged_shows_unknown_state_and_the_sudo_commands` self-panics with "this test needs a non-root job" because the container runs as root; it passes in CI, which runs non-root.

- [x] `cargo test -p irlume-cli` passes (390; the one skip-under-root failure above is pre-existing on `main`)
- [x] `cargo clippy -p irlume-cli --all-targets` clean
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc -p irlume-cli --no-deps` clean
- [ ] Docs/CHANGELOG updated if user-facing (adds a doctor line and a `capture-mode` machine-report id; no separate doc pins doctor's exact output)
- [ ] Hardware-validated (reporting only, opens no camera; the underlying capture-mode measurements are already recorded on #100 and #340)


archledger:
CI note: the only red check is the Fedora Packit build (`rpm-build:fedora-43/44`), failing at the **SRPM** stage (job `srpm/655811`), both targets in the same second they started. That stage builds the source RPM from the tree and the spec and doesn't compile the Rust, so this reporting-only diff can't affect it. The Rust CI is green (the ones already completed: cargo-deny, systemd, coverage, Analyze, hardware; `fmt · clippy · build · test`, `test (stable)`, ASAN, nix and fuzz were still finishing and green so far).

This is the same transient Packit SRPM failure hitting #377 right now, on PRs created after ~18:51; the identical Fedora builds passed on #376 at 18:16. It should clear on a Packit re-run (`/packit build` or a re-push); nothing in the diff to fix.

archledger:
Correction to my note above: the Rust CI was not fully green. `test (stable)`, AddressSanitizer, and `fmt · clippy · build · test` failed on a real issue in this change, not the SRPM.

Cause: the `machine_api` conformance test walks the check-id registry in `docs/MACHINE-API.md` against the ids `doctor` actually emits, in both directions, because a check id is public API. The new `capture-mode` check shipped without its registry row, so the test failed with `these check ids ship without a row in the MACHINE-API.md registry: ["capture-mode"]`. My local run missed it because I tailed only the last few test-result lines and the failure was in an earlier integration-test binary.

Fixed in c69f031 by adding the registry row (id, what it reports, and its states). Locally the full `irlume-cli` suite now passes except the one pre-existing `tui` test that requires a non-root job (it passes in CI). CI will re-run on the push.

The Fedora Packit SRPM failure is separate and still the transient infra one described above.

PR #379, opened 2026-08-08
Stop asking a camera to capture concurrently once it keeps failing at it (#100)

archledger:
## What & why

#100 item 1, the automatic trigger. **Stacked on #378** (base is `claude/doctor-capture-mode-100`), because it extends the doctor block that PR adds. Review #378 first; if you would rather this target `main`, say so and I will rebase and take the conflict in `cli/main.rs`.

Some Hello modules starve their own RGB interface while both sensors stream. The frame arrives without error, too dim for the detector to find a face, and the cross-spectrum self-heal recaptures RGB alone to rescue the login. That has worked silently since it was added: the camera is repaired one login at a time and never actually reconfigured, so every login keeps paying the recapture.

The daemon now learns from it. After three consecutive authentications where this pairing lost its RGB frame to contention, it stores `sequential` for that pairing and says so in the journal.

### What counts, and why it is narrower than "the self-heal fired"

The self-heal's own condition is not a fault signature. "RGB has no face, IR does" is equally the shape of a legitimate **dark login**, which this codebase supports deliberately (`uncertain_short_circuits`, observed live at rgb faces=0, ir faces=1 at 0.92), and the shape of a user still walking up, where the solo recapture succeeds ~700ms later because the person stopped moving. Counting either would demote a healthy camera, and in the dark case it would do so for everyone who logs in at night.

So a counted event needs both halves of the contention signature: the solo recapture **recovered** the face, **and** the overlapped frame arrived **measurably dimmer** than the solo one. "Measurably" is the capture-mode probe's own rule, reused rather than reinvented: `CONCURRENT_SIGNAL_FLOOR` (0.80) against a scene of at least `CONCLUSIVE_SCENE_BRIGHTNESS` (100.0). That brightness floor is what excludes the dark room, and it is not an arbitrary guard: the same NexiGo measured 0.42-0.56 retention against a sequential arm at 117-143, then 0.91 an hour later in a dark room against an arm at 62.

### The threshold is a decision, not a measurement

Three is your choice, and the constant's doc comment says so in those words. Every other number this rule leans on was earned on hardware; nothing in this repo has ever counted self-heal firings, because the only trace they leave is a `dlog!` that is off unless `IRLUME_LOG` is set, so there is no rate to fit a threshold to. That is also why every clause errs toward **under**-counting: not switching costs one extra RGB capture per login, which is what already ships, while switching wrongly taxes every later capture.

If you would rather earn the number the way the rest of this subsystem earned its constants, the strictly better landing order is: ship the counter, the dimming test and the journal line with **no write**, read the real firing rates off the journal, then land the write. I have kept the write in because you named the threshold, but the split is one commit away and I am happy to do it.

### What it refuses to do

- **An operator-forced mode is never learned from.** `IRLUME_SEQUENTIAL_CAPTURE=0` is the only env value that can reach a self-heal at all, and `capture_mode_decision` already states the rule: a set env var decides alone because setting it is an explicit instruction.
- **The write is guarded, not blind.** `store_sequential_if_still_concurrent` re-reads under the `cameras.conf` lock and writes only if the verdict is *still* concurrent. Evidence and write are separated by several logins, so an operator who ran `camera-tune` in that window measured the camera directly, and a measurement outranks an inference. This is also the first automatic writer that overwrites a measured verdict, which is why it is guarded at all: `store_capture_mode_if_absent` cannot express it, since an absent verdict already means sequential.
- **A failed write never fails a login.** It runs at the tail of an authentication that already has its verdict; a read-only `/etc` must not turn a successful login into an error.
- **No probe ever runs during an authentication**, the rule #100 names as the one that makes automating this safe. The switch infers from captures that already happened.

### Reporting

One journal line at the switch (an `eprintln!`, not a `dlog!`, because this writes `/etc` on the user's behalf and a debug-only record would be invisible by default). A sidecar `capture_mode_origin.<ids>` key records that irlume inferred this, and when; it is written **before** the mode so a crash between the two renames can only leave an inert stamp, never a switched verdict that reads as measured. The mode's own value is untouched, so an older irlume still sees exactly `sequential`. Doctor reports an auto-switched pairing **without ever using the word "measured"** and names `camera-tune` as the way to replace the inference with a measurement.

### Two side effects, stated rather than claimed away

A wrong switch is **not** the flat 700ms the issue calls safe. It is ~0.7s (ASUS) to ~1.3s (NexiGo) per capture inside *fixed* retry budgets: the 15s grace window is documented as roughly 10 attempts at 1.1-1.5s each, and the sudo window holds 3-4. The user loses attempts, not just seconds. It also stops enrolment holding both streams open.

Concurrent capture windows overlap while sequential ones merely abut, so the switch slightly widens the interval the cross-spectrum co-location cue binds. That is not a posture change; sequential is the shipped default for every unmeasured pair and the mode `camera-tune` itself selects, so this cannot take a machine below the product's own floor. Verified rather than assumed: no threshold, tier, gate or floor in `irlume-liveness` or `irlume-vision` is a function of capture mode.

## Testing

Container-run, with ONNX Runtime 1.24.4 and the `models-v1` weights fetched the way CI does.

Full suite with `--no-fail-fast`, every result line read (this is the check I skipped on #378, which is how its registry-row failure reached CI):

- `irlume-auth`: **80 passed, 0 failed**, which is the baseline 72 on this branch's base plus the 8 new tests
- `irlume-camera`: 266 passed, 1 failed *(known, pre-existing)*
- `irlume-cli`: 391 passed, 1 failed *(known, pre-existing)*, plus integration binaries 62/19/15/8/1 all green, including `machine_api` (19), so the check-id registry conformance stays satisfied; this change adds a doctor **variant**, not a new check id
- `irlume-common`: 78 passed. `irlume-daemon`: 104 passed
- `cargo fmt --check` and `cargo clippy --all-targets` clean across all five crates

The two failures are the container's root-only artifacts, verified earlier in this session to fail identically on the base branch: `ir_emitter::tests::a_visible_config_after_a_failed_save_keeps_the_record` and `tui::tests::models_screen_unprivileged_shows_unknown_state_and_the_sudo_commands` both force a permission failure that root bypasses. Both pass in CI, which runs non-root.

New tests, and what they pin:

- `a_counted_event_needs_recovery_and_measured_dimming`: the four shapes this rule separates, each with numbers already measured in the repo: the NexiGo fault, a non-recovery, the **dark-login** shape, and the healthy/settling shape.
- `the_dimming_test_is_the_probes_own_rule_at_both_boundaries`: both floors from both sides, so a mutant relaxing `<` to `<=` dies, and the constants stay the camera crate's rather than drifting copies.
- `the_switch_lands_on_the_third_consecutive_event_and_not_before`: asserts the constant is 3 *and* walks 0..=4 through the due-check, so changing 3 to 2 or 4, or moving `>=`, both fail. It also pins the premise that makes the self-heal reachable (an unmeasured pair defaults to sequential), so a future flip of that default surfaces here.
- `a_clean_concurrent_capture_clears_the_streak`: the three-way signal split, including that `Inconclusive` is neutral in both directions.
- `an_operator_forced_mode_is_never_written_back`: at every streak length, and binds the guard to the one place that produces the source string so a rename breaks the test instead of silently disabling the guard.
- `the_switch_announces_itself_and_names_the_override`, `a_verdict_that_changed_under_us_is_left_alone`, `a_failed_write_is_reported_and_never_claims_the_mode`: the journal line by return value (the debug-log switch freezes in a `OnceLock`, so capturing stderr is unreliable); the failed-write case pins that an error is reported and cannot claim the mode changed.
- `the_authentication_path_never_runs_the_capture_mode_probe`: a rule-pin test in the idiom of `camera_authority.rs`. **Proven to have teeth**: I injected `measure_contention(...)` into `maybe_switch_capture_mode` and it failed naming the exact file and line, then passed again on restore.
- Camera crate: origin-grammar parsing (lenient about the time, silent about anything it does not recognise) and that the origin key is a sidecar that never touches the mode's value.
- Doctor: an auto-switched mode is `Info`, names the mode and `camera-tune`, renders the age only when known, and never says "measured".

**Not verified on hardware**, and the reason this is left for you:

- No real Hello pair has been auto-switched end to end. `device_identity` reads `/sys/class/video4linux`, which CI and the loopback lane do not have, so the write path itself is exercised only by its guards and its pure parts.
- One assumption behind the dark-login exclusion is untested: that a solo RGB recapture of a genuinely dark room cannot read a mean ≥ 100 when the overlapped frame read far less. The repo has no measurement of solo-vs-concurrent exposure behaviour in the dark. If that is wrong, dark logins re-enter the count. Worth one check on the NexiGo.

- [x] `cargo test --no-fail-fast` across all five affected crates, every result line read; only the two pre-existing root-only failures
- [x] `cargo clippy --all-targets` clean
- [ ] Docs/CHANGELOG updated if user-facing (COMMANDS, DEBUGGING, SETUP and MACHINE-API updated; CHANGELOG is release-curated)
- [ ] Hardware-validated: **needed before merge**, per the two gaps above


archledger:
Holding this as a draft rather than merging it tonight, on the project's own merge policy.

That policy requires real-hardware validation on the devices a change touches, and says to fall back to asking when any part of the bar is missing. For this PR it is missing in a way that matters:

- it writes `/etc/irlume/cameras.conf` from the authentication path, and no real Hello pair has been auto-switched end to end;
- the switch threshold of three is a decision, not a measurement, as its own doc comment says;
- the dark-login exclusion rests on an assumption nobody has tested, that a solo RGB recapture of a genuinely dark room cannot read a mean at or above 100 when the overlapped frame read far less.

Everything else is done: the review round's findings are fixed and proven, CI is green on the final sha, and the origin-stamp defect that made a wrong inference permanent is closed, so `camera-tune` can now correct it.

The rest of tonight's backlog has merged. This one wants a camera session: one NexiGo run through three contention logins to see the switch fire and the journal line appear, and one dark-room capture to check the exclusion holds. Ready to merge the moment that exists, or now if you would rather take it on the code review alone.

Its base, #378, is unaffected and merges on its own.

archledger:
Understood, and the three gaps are the right ones — they are the same three I listed, so nothing here surprises me. Leaving it as a draft and not touching it.

One thing that may shorten the camera session, on the third gap. The dark-room reading you would be taking already exists in the tree, for the exact quantity the exclusion tests and on the camera in question:

> the same NexiGo N930W measured 0.42-0.56 retention with the sequential arm at 117-143, then 0.91 an hour later **in a dark room with the sequential arm at 62**
> — `CONCLUSIVE_SCENE_BRIGHTNESS`, measured 2026-07-25

`conclusive()` tests `sequential.rgb_mean >= CONCLUSIVE_SCENE_BRIGHTNESS`, and `self_heal_is_dimming` tests `solo_mean >= CONCLUSIVE_SCENE_BRIGHTNESS` — the same number, the solo RGB arm. In that dark room it read 62, well under the 100 floor, so the exclusion holds on the one dark-room measurement this repo has.

Three honest limits on that, none of which I would want glossed over:

- one room, one camera, one session;
- it comes from the `camera-tune` probe's sequential arm rather than the self-heal's solo recapture — same one-shot capture path, but not the same moment or exposure history;
- it shows the dark solo arm reading 62, which is not the same as showing it can never reach 100. A room lit brightly enough to push the solo arm over 100 is no longer the dark-login case the clause exists to exclude, but that is an argument, not a measurement.

So it narrows the gap rather than closing it. If it is worth anything, it is that the dark-room half may need confirming rather than establishing, and the NexiGo contention run is the part that has no substitute.

No action wanted from me on this PR unless you ask.

---
_Generated by [Claude Code](https://claude.ai/code)_

archledger:
Merged main into this branch as `cee7814`. Leaving it here rather than waiting, because it was not just behind: it had gone un-mergeable, and it was also carrying conflict markers that an earlier merge committed without finishing.

`docs/MACHINE-API.md` had this sitting in the `capture-mode` row on the pushed branch:

```
<<<<<<< HEAD
| `capture-mode` | ... forced by `IRLUME_SEQUENTIAL_CAPTURE` ...
=======
| `capture-mode` | ... switched to sequential automatically ...
>>>>>>> 427b328 (Stop asking a camera to capture concurrently once it keeps failing at it (#100))
```

The row now names all four origins in one sentence: measured by `camera-tune`, switched automatically after repeated concurrent-capture RGB losses, forced by `IRLUME_SEQUENTIAL_CAPTURE`, or the unmeasured sequential default. It keeps main's account of when the check reports `unknown` and this branch's rule that an automatic switch is never reported as a measurement.

The conflicts against main came from #378, which landed the doctor capture-mode report while this branch was parked. Five in `crates/irlume-cli/src/main.rs`, all the same shape: main added the `Overridden` and `Unreadable` cases, this branch added `AutoSwitched` and `age_days_since_unix`, and neither removes anything the other added. The enum now carries both sets.

Full workspace suite with `--no-fail-fast`: the only failures are the two that fail on unmodified main in this container as well, both root-only (`ir_emitter::tests::a_visible_config_after_a_failed_save_keeps_the_record`, `tui::tests::models_screen_unprivileged_shows_unknown_state_and_the_sudo_commands`). fmt and clippy clean.

Still a draft, and I have not touched the switching behaviour itself. This is only so the branch builds and merges when you come back to it on hardware.

---
_Generated by [Claude Code](https://claude.ai/code)_

archledger:
Hardware session on the NexiGo, 2026-08-08. The switch never fired, and the reason is worth more than the test passing would have been.

## Setup

NexiGo HelloCam N930W (`3443:c803`) moved onto the Zenbook, this branch built and running as the daemon (`rgb=/dev/video4 ir=/dev/video6`, confirmed in its own journal), `IRLUME_LOG=1`, a fresh enrollment bound to that camera, and the pair forced to `concurrent` in `cameras.conf` so the self-heal path is reachable at all. Everything restored afterwards, byte-exact.

The contention this feature targets is real on this module. `camera-tune` measured it directly:

> capture mode sequential for this camera: concurrent capture keeps **50% of RGB** and 98% of IR brightness and saves 3505ms (6 rounds)

That is squarely in the 42-56% band recorded on #100.

## The switch did not fire, in any lighting

Ten or more authentications through `irlume auth test`, which reaches `authenticate_for` and therefore the counter, in three conditions:

| lighting | RGB under concurrent capture | counted events |
|---|---|---|
| normal room light | `rgb 640x480 in 2447ms, faces=1 top-det=0.95` | 0 |
| dimmed | `faces=1 top-det=0.95` | 0 |
| dark, laptop screen only | `faces=1 top-det=0.93`, `ir_bright=111-130` | 0 |

`capture_mode` stayed `concurrent` and `capture_mode_origin` stayed empty throughout. Every attempt granted; one went through the dark path on IR (`rgb+ir fusion p=1.00; rgb 0.52/ir 0.95`).

The counter needs the RGB frame to LOSE the face. On the authentication path it never did, even with the room dark.

## Where the contention actually shows up

It shows up under HELD SESSIONS. During enrolment on the same camera, same room, minutes earlier:

```
assess: rgb 640x480 in 173ms, faces=0 top-det=0.00
assess: ir  640x360 in 472ms, faces=1 top-det=0.94
```

Four consecutive scans, identical shape: RGB back in 128-173ms with nothing, IR finding the face at 0.94. That is the degradation signature this feature is built on, and it is severe.

Enrolment holds both streams open for its whole capture loop. Authentication opens per capture. The NexiGo starves its RGB interface in the first case and survives the second.

## What that means for this PR

The counter is deliberately scoped to authentications, and the comment explaining why is sound: enrolment coaches the user through partial poses and would poison the streak. But on this hardware the authentication path does not produce the fault, so the trigger is aimed at the path where the problem is absent and excluded from the path where it is present.

I am not proposing a fix in this PR. Two things follow and both want deciding rather than patching:

1. Whether the streak should count enrolment-shaped captures at all, given that is where the contention bites, and how to do that without the partial-pose problem the current exclusion exists for.
2. Whether a threshold of three is meaningful when the event it counts did not occur once in a session designed to provoke it.

Kept as a draft. This session is the argument for holding it, not for abandoning it: the feature targets a real, measured fault, and the evidence now says the trigger does not reach it.

## Not established

One camera, one subject, one room, one evening. A different Hello module, or this one on a busier USB bus, could well lose the RGB face on the authentication path. What is established is that a module measured at 50% RGB retention did not, in any light I could produce, including none.

archledger:
## The path this counter watches now has a measured sibling, and the sibling cannot feed it

#414 merged tonight (`524028f`) and added `CaptureShape { held_sessions, attempts, ir_only_attempts }` to the enrolment capture loop. That tally is exactly the signature this PR is built on, on the path the 2026-08-08 hardware session found the fault actually occurs. It looks like the missing trigger. It is not, and the reason is structural rather than a matter of wiring.

A counted event here requires `self_heal_is_dimming(recovered, concurrent_mean, solo_mean)`, and all three inputs come from the solo RGB recapture. On a held session that recapture never happens: `self_heal_may_recapture` returns false on its `held_sessions` clause, because #381 forbids reopening a node this process is already streaming, because of #187. So the enrolment path produces the SHAPE with none of the evidence that discriminates it.

That matters because the shape alone is ambiguous, and this repository has now written down why twice. A dark room produces an identical reading: no RGB face, IR face present, which `uncertain_short_circuits` calls the dark-login shape and records observed live at rgb faces=0, ir faces=1 with a 0.92 detection. #389's analysis reaches the same wall from the other side, which is why the message that landed there offers two readings instead of naming one.

So "count enrolment-shaped captures" imports precisely the ambiguity this PR's counter was carefully designed to exclude. Question 1 from the hardware session has an answer, and the answer is no, not with the evidence available on that path.

## Why I am not landing the counter without the write

The body above calls that the strictly better order, and it was right when written. The hardware session changed it: the counter watches authentications, and on a module measured at 50% RGB retention it produced zero events in more than ten attempts across normal, dimmed and dark rooms. Landing an observer collects nothing where the event does not occur, and it would put code in the authentication path to gather data that the one available module says will not arrive. Two Hello modules exist here and both have been measured.

## The route that would make an enrolment-side trigger possible

Give the held path the arm it lacks, without reopening a streaming node: after a failed enrolment has RELEASED its sessions, take one solo RGB frame and compare it against the concurrent means already observed in that loop.

That is legal where the self-heal is not, because the sessions are gone by then, so #187 is not in play. It supplies `recovered` and `solo_mean`, which is the whole discriminator. And it would pay twice, since it is also what #389 needs to name a cause rather than offer the user two readings.

What it needs before any code: the enrolment loop does not currently retain an RGB mean per attempt, and taking a capture after a failure is a behaviour change on a path a user is waiting on. Both want deciding, and validating it wants the NexiGo again.

## State

Kept as a draft. It is 21 commits behind main and carries a merge history that already had to be repaired once. If it is picked up, rebasing onto current main comes first, and the counter's target path is the open question rather than its implementation, which the review round and the test suite already settled.

archledger:
## Measured on the NexiGo, 2026-08-10: the solo frame after release IS the discriminator

The question this PR has been stuck on is whether the held path can ever supply the evidence its counter needs. It can, if the frame is taken after the sessions are released. Two runs, minutes apart, same seat:

| | run 1 | run 2 |
|---|---|---|
| held frames finding a face | 0 of 6 | 0 of 6 |
| held RGB mean | 51.7 | 51.1 |
| solo RGB mean, after release | 146.9 | 146.6 |
| solo faces | 1 at 0.95 | 1 at 0.95 |
| retention | 35.2% | 34.9% |
| `recovered` / `lit` / `dimming` | true / true / true | true / true / true |
| `self_heal_is_dimming` would count it | **true** | **true** |

NexiGo HelloCam N930W (`3443:c803`), `rgb=/dev/video4 ir=/dev/video6`, both sessions armed for the held phase so the RGB reads happen while the IR stream runs, which is the enrolment shape. The sessions are then dropped and one solo RGB frame is taken through `capture_rgb`, so no node is reopened while this process streams it and #187 is not in play.

Retention of 35% against a `CONCURRENT_SIGNAL_FLOOR` of 0.80 is not a marginal call. The earlier `camera-tune` on this module measured 50%; under a held pair with a detector running it is worse.

### What this changes

The blocker recorded above was that the enrolment path yields the SHAPE with none of the discriminator, because `self_heal_may_recapture` suppresses the solo recapture while sessions are held. Taking the frame one step later, after the release, supplies `recovered` and `solo_mean` without touching that rule. The shipped `self_heal_is_dimming` then answers true on real frames from the module this feature was built for.

It also answers #389. The reason that message offers two readings is that darkness and contention are identical while both streams run. They are not identical after the release: a dark room stays dark in the solo frame, and this camera went from 51 to 147.

### What it does NOT establish

The IR arm was dark in both runs, mean 17 and 32, with the emitter warning firing. The #379 hardware session saw enrolment fail with IR FINDING the face at 0.94. So this measured the RGB-starvation half of the signature and not the IR-kept-the-face half, and a counted event in production would want both.

One module, one subject, one room, two runs. Nothing here says the authentication path produces this on any hardware; the 2026-08-08 session says it does not on this one.

### What it would cost to build

The enrolment loop does not retain an RGB mean per attempt, and taking a capture after a failed enrolment is a behaviour change on a path a user is waiting on: one extra RGB open, measured at 146ms to 173ms on this module in the session above. That is paid only on a capture loop that already failed.

archledger:
## Both exclusions measured, 2026-08-10. Ten runs, three cells.

The positive result above needed its two false-positive paths closed before it could justify code. Both are now measured on the same probe, same seat.

| cell | runs | held faces | held mean | solo faces | solo mean | `self_heal_is_dimming` |
|---|---|---|---|---|---|---|
| NexiGo, lit room, the fault | 2 | 0 of 6 | 51.7, 51.1 | 1 at 0.95 | 146.9, 146.6 | **true** |
| NexiGo, dark room | 4 | 0 of 6 | 46.5 to 47.2 | 0 | 18.0 every run | false |
| ASUS built-in, lit room | 4 | 6 of 6 | 160.1 to 163.0 | 1 at 0.94 | 156.0 to 157.9 | false |

### The dark room is excluded twice, not once

`lit` fails as expected, 18.0 against a floor of 100. The unplanned part: the relationship INVERTS. Held reads 46.6 while solo reads 18.0, because the IR emitter fires during the held phase and leaks into the RGB sensor. So `dimming` is false too, on its own, without leaning on the brightness floor at all.

That matters for the design's robustness. The exclusion does not rest on one constant being right for every room; a dark scene fails both clauses for independent physical reasons.

### The healthy camera is excluded by recovery

The ASUS built-in finds the face in 6 of 6 held frames, so `recovered` is false and nothing is counted. Its held mean of 161.0 against a solo 157.7 is 98% retention: this module does not starve its RGB interface under a held pair at all, which is what makes it the right negative control.

Worth noting in passing, and not acted on here: the ASUS pair is stored `sequential` in `cameras.conf` while measuring 98% retention under exactly the condition sequential exists to avoid.

### What the three cells together support

Fires on the fault, excluded on darkness, excluded on a healthy module. That is the basis this feature has been missing since #100 was filed, and it is now measured rather than argued.

Still not established: one subject, one room, one evening, two modules. The IR arm was dark in the positive runs, so the RGB half of the signature is measured and the IR-kept-the-face half is not. Nothing here says the authentication path produces this on any hardware.

archledger:
## The config write has its evidence, and it is not the design that shipped in #417

Sixteen runs, 2026-08-10, one subject, two modules. The question was whether a solo frame after release can carry a durable config write, and the answer is that it cannot on its own, but held / solo / held-again can.

| cell | camera | runs | held A | solo B | held A' | A/B | A/B/A |
|---|---|---|---|---|---|---|---|
| lit, starving | NexiGo | 4 | 50.2 to 50.6, 0 of 5 faces | 139.7 to 140.5, face | 50.2 to 50.8, 0 of 5 | true | true |
| lit, healthy | ASUS | 4 | 153.2 to 153.7, 5 of 5 | 149.3 to 151.2, face | 153.0 to 154.3, 5 of 5 | false | false |
| dark, light switched on between B and A' | ASUS | 4 | 28.9 to 31.8, 0 of 5 | 162.7 to 163.7, face | 168.4 to 174.4, **5 of 5** | **true** | **false** |
| dark, starving | NexiGo | 4 | 46.5 to 47.2, 0 of 5 | 18.0, no face | (earlier session) | false | n/a |

### The A/B is wrong 4 times out of 4 in the adversarial cell

A healthy camera, a dark room, and a light coming on between the two captures. That is not a contrived scenario, it is walking up to a sleeping machine with a motion-controlled lamp. The single solo frame confirms contention every time, and there is no camera fault.

The A/B/A rejects all four, and the reason is physical rather than statistical: a healthy camera's held phase TRACKS the light, so A' returns at 168 to 174 with a face in 5 of 5 frames and the reversal fails.

### Why it still confirms on the module that is genuinely starving

On the NexiGo the held frame reads 46 to 51 whether the room is lit or dark, while the solo frame moves from 18.0 to 140. The held reading is not measuring the scene at all there; it is pinned by the starved stream. So A' stays dark with no face and the reversal holds.

That is the discriminator stated properly: not "is the solo frame brighter", which a lamp satisfies, but "does the held phase track the scene or is it pinned". Two held phases either side of a solo one is the cheapest way to ask.

### What this means for each of the three

**#417, the message, stands.** It uses the A/B, so in the adversarial cell it will say the wrong thing, and this measurement puts a number on how often: 4 of 4 in that cell. It is acceptable there only because the message names that exact confound out loud and because the remedy it prescribes, `camera-tune`, measures the thing directly and would find nothing on a healthy camera. The cost of being wrong is one wasted measurement.

**#379, the config write, needs the A/B/A.** A wrong write is durable, overwrites a possibly-measured verdict, and taxes every later capture inside fixed retry budgets. A design with a 4-of-4 false-positive rate in a realistic scenario cannot carry that, and now it does not have to.

**#100's item 4 becomes cheaper to justify either way.** A re-measure makes a wrong inference cost one slow session rather than forever, which lowers the evidence bar rather than raising the evidence.

### Not established

One subject, one room, one evening, two modules. The adversarial cell used a deliberate light change, which is the strongest form of that confound rather than the average one. A camera that starves AND sits in changing light would still confirm, and correctly so, since it is starving. Nothing here measures a second starving module, and the IR arm was dark in the NexiGo positives, so the RGB half of the signature is what these numbers describe.

archledger:
Closing this draft. #100 was resolved by #419, which was hardware-validated on the NexiGo HelloCam N930W. The approach in this PR was measured wrong: the trigger watches a path that produced zero counted events on the NexiGo in 10+ attempts across three lighting conditions, and the A/B discriminator was wrong 4 of 4 in the adversarial cell. #419 fixed both with an enrolment-only trigger and an A/B/A check.

Issue #374, opened 2026-08-08
The camera pin is published four times and read without the lock, so a reader can see a pair that never existed


archledger:
Split out of #365, where taking the file's lock around the four writes fixed a real defect (a locked writer racing unlocked ones lost the keys it had just written) but did NOT fix the one below. Filing it rather than leaving the fix's comment implying more than it does.

`config::write_camera_pin` writes four keys. Each `write_kv` reads the whole file, rewrites it, and renames a new version over the old (`write_0600_atomic`). So the pin is published **four times**, not once, and `flock` is advisory while the readers in `irlume-camera` (`select_pair` and friends) never take it.

Two consequences:

**A reader can observe a pin that was never committed.** `select_pair` opens the file more than once, so a rename landing between those reads combines an RGB path from the old pin with an IR path or identity from the new one. `resolve_saved_pair` then evaluates a pair that never existed as a unit. That pair is the anti-injection binding, so this is not cosmetic: the point of pinning `vid:pid:serial` is that the RGB and IR nodes are known to belong together.

**A partial failure is permanently published.** The old pin is camera A; `SetCameras` selects camera B; the RGB write succeeds and the IR write fails, say on a full filesystem. `cameras.conf` then contains B's RGB path with A's IR path and identities, and nothing rolls it back.

The fix is a different write path, not another lock: build the updated text in memory and publish all four keys in ONE rename, and give readers a single-read tuple accessor so the pin is read as one value. `write_kv`'s single-key parse should be factored so both share it.

Worth being explicit about what a test for this needs, because the one in #365 does not do it: asserting the four keys are present afterwards proves nothing about atomicity, and asserting the `.lock` sidecar exists shows only that `OpenOptions::open` ran, not that `flock` succeeded. The observation that matters is a concurrent reader seeing only the complete old tuple or the complete new one, which needs a harness this repo does not have today.

Not urgent by likelihood: inside the daemon both writers run on the single worker and are already serialized, and the CLI writes only `settings.conf`. The exposure is an administrator or a configuration manager editing `/etc/irlume/cameras.conf` while the daemon writes it, plus the partial-failure case, which needs no concurrency at all.

PR #380, opened 2026-08-08
A lock only its writers take is no exclusion at all (daemon tests)

archledger:
## What & why

The ASAN lane has been going red intermittently: a SEGV on a READ of `0xffffffff00000000` inside `getenv`, on a test thread, aborting the whole daemon test binary. It failed on two branches, **passed on a re-run of the identical commit**, and passes on `main`. That is the signature of a race, not a defect in any one diff. No issue filed for it yet; happy to open one if you'd rather track it separately.

The stack names it exactly:

```
getenv <- _nss_systemd_getpwnam_r <- getpwnam_r <- users::uid_for_name
      <- uid_of <- authorized_for <- pregate
      <- tests::pregate_enforces_the_privilege_every_variant_declares
```

A passwd lookup reads the environment **inside glibc**. On a systemd host it resolves through `libnss_systemd`, whose `getenv` walks the same `environ` array `setenv` reallocates. Meanwhile forty-odd `set_var`/`remove_var` calls run in other tests on other threads.

This crate already had `env_lock` for exactly this hazard, and every one of its **writers** took it. What that cannot protect against is a test that only **reads** the environment, and reads it without ever mentioning an environment variable. Nine tests resolve a username and took no lock, so they raced every writer, and it was the reader that died. Same shape as the advisory `flock` in #374: exclusion needs both sides to take it.

All nine now take `env_lock`. I found the set by searching for the calls that reach `getpwnam_r`/`getpwuid_r` rather than just the one in the crash: `pregate`, `authorized_for`, `uid_of`, `uid_for_name`, `name_for_uid`, `identify_scope`. Searching only the crashing path would have fixed four of the nine and left the lane still flaky.

A test pins the **rule**, in the idiom of `camera_authority.rs`, because the failure mode is a new test added later that resolves a username and takes no lock. No behavioural test can catch that: the race needs a concurrent writer and enough load to lose, which is why it arrived as a randomly red pipeline instead of a failing test.

## Two things my first attempt got wrong

Both caught by running the suite locally, which is the reason this PR is worth reading:

**`enrollment_summary_test_lock()` RETURNS `env_lock()`.** Its own doc says so, and says *"No test acquires both helpers, so this cannot recurse."* Adding a second acquisition to the two tests that hold it deadlocked the entire suite on a non-reentrant mutex. Those two need no guard; they already hold it. The rule test now enforces that invariant too, which nothing did before.

**Holding a suite-wide lock across slow work serializes everything behind it.** The wedged-queue test waits on socket timeouts, and guarding its whole body took the daemon suite from 17s to 131s. Its guard now covers the passwd lookup alone, which is all correctness ever required.

## Testing

- Daemon suite: **105 passed, 0 failed, 4 ignored, 9.9s**, faster than the 16.7s baseline, since the added test replaces no work and nothing serializes now. Zero tests report as slow.
- `cargo fmt --check`, `cargo clippy -p irlume-daemon --all-targets`, and `RUSTDOCFLAGS="-D warnings" cargo doc -p irlume-daemon --no-deps` all clean.
- The rule test is **proven to have teeth**: removing the guard from `pregate_enforces_the_privilege_every_variant_declares` (the test that actually crashed in CI) makes it fail and name that test. Its own sanity assertion earned its keep too: my first walker parsed only 21 of 109 tests and the assertion caught that rather than passing vacuously.

What this cannot prove: the race is load-dependent, so a green ASAN run here is consistent with the fix and also with getting lucky. The argument for it is the mechanism, which is exact: the crash is a read of `environ` from a lookup that now holds the same lock every writer holds. Worth watching the ASAN lane over the next several runs.

- [x] `cargo test -p irlume-daemon` passes (105, 0 failed)
- [x] `cargo clippy -p irlume-daemon --all-targets` clean
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc` clean
- [ ] Docs/CHANGELOG updated if user-facing (test-only change; nothing user-visible)
- [ ] Hardware-validated (not applicable: test-suite locking only, no runtime code touched)

### DIFF
--- a/crates/irlume-daemon/src/main.rs
+++ b/crates/irlume-daemon/src/main.rs
@@ -15,6 +15,41 @@ use std::io::{BufRead, BufReader, Read, Write};
 use std::os::unix::net::{UnixListener, UnixStream};
 use zeroize::Zeroize;
 
+/// The one lock that serialises `environ` access across this binary's tests.
+///
+/// It lives at crate scope, not inside `main.rs`'s `mod tests`, because
+/// `irlume-daemon` is a single bin target: `users.rs`'s own `#[cfg(test)] mod
+/// tests` compiles into the SAME test binary and libtest runs both across one
+/// thread pool. A lock private to one module cannot be taken by the other, so
+/// two `users.rs` tests resolved usernames against every environment writer
+/// with nothing between them (#380 review).
+///
+/// A `RwLock`, not a `Mutex`, because the hazard is asymmetric. `getpwnam_r`
+/// READS `environ` inside glibc; `set_var` REWRITES it. Concurrent readers do
+/// not race each other, only a writer. Shared read guards let the passwd
+/// lookups overlap, which is what keeps a suite-wide guard from serialising
+/// every socket timeout behind it: an earlier attempt at this took the daemon
+/// suite from 17s to 131s.
+///
+/// Never taken inside `users::uid_for_name`/`name_for_uid` themselves. A test
+/// holding the write guard reaches those functions through production code,
+/// and a nested read acquisition would deadlock.
+#[cfg(test)]
+pub(crate) mod test_support {
+    static ENV_LOCK: std::sync::RwLock<()> = std::sync::RwLock::new(());
+
+    /// Shared: this test only READS the environment, which is what a passwd
+    /// lookup does inside glibc.
+    pub(crate) fn env_read() -> std::sync::RwLockReadGuard<'static, ()> {
+        ENV_LOCK.read().unwrap_or_else(|e| e.into_inner())
+    }
+
+    /// Exclusive: this test calls `set_var`/`remove_var`.
+    pub(crate) fn env_write() -> std::sync::RwLockWriteGuard<'static, ()> {
+        ENV_LOCK.write().unwrap_or_else(|e| e.into_inner())
+    }
+}
+
 mod arbiter;
 mod users;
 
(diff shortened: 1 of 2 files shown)


archledger:
ASAN went red on `6e4f453`, and it is not the `getenv` SEGV — no SEGV at all:

```
tests::a_request_needing_the_engine_is_refused_while_it_loads
a refusal within the deadline, not a wait for the engine:
Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }
```

Four tests take the shared guard **inside** the spawned server thread, while the client on the other end is already counting down a read deadline. `std::sync::RwLock` is write-preferring, so a queued writer blocks a new reader; with sixty tests taking the exclusive guard, that thread can wait past the deadline and the client reports a timeout instead of the refusal the test is asserting.

Fixed in `7ad2fd5` by taking the guard in the test body, before the spawn. The comment that placed it inside the thread reasoned that a test-body guard could not cover a lookup on another thread — but the guard does not have to be held *by* that thread, only *while* it runs, and all four tests join their server thread four lines later, for the reason written there. So a test-body guard outlives the lookup and covers it exactly, and nothing blocks inside the timed region.

Also moved two reader-only tests off the exclusive guard: `root_and_self_authorized_others_denied` and `identify_scope_confines_non_root_peers_to_their_own_account` resolve a username and mutate nothing, so they take `passwd_lock()` now. Leftovers from before the lock was split.

`irlume-daemon`: 105 passed, 0 failed, including the test that failed here and the scanner that pins the rule.

One caveat on reading the ASAN lane: it had been green on four consecutive runs before this, and it also passed once on the *unfixed* code earlier, so a green run is not evidence either way. The mechanism is the argument.

---
_Generated by [Claude Code](https://claude.ai/code)_

archledger:
Heads-up: the guard-hoist fix I pushed as `7ad2fd5` was dropped when the branch was rewritten — all four `env_read()` acquisitions were back inside the spawned server threads. Re-applied in `ac3c8e9`.

Flagging it because the failure it fixes is intermittent, so its absence does not show up as a red run. It only surfaces when a writer happens to be queued while a server thread is acquiring the shared guard and the client is counting down its read deadline — which is how it arrived the first time.

Re-applied: the four guards move into the test body before the spawn (each of those tests joins its server thread, so a test-body guard covers the lookup), and the two reader-only tests go back on `passwd_lock()`.

`irlume-daemon`: 105 passed, 0 failed, clippy clean.

---
_Generated by [Claude Code](https://claude.ai/code)_

archledger:
ASAN red again, same test and same cause. Re-applied in `8707202` — this is the third application; `7ad2fd5` and `ac3c8e9` were both discarded when the branch was rewritten, and each time the four `env_read()` calls returned to the inside of the spawned threads.

Worth knowing when rebasing this branch: **the failure is intermittent, so losing this fix does not show up as a red run.** It only fires when a writer happens to be queued while a server thread acquires the shared guard and the client is counting down its read deadline. A green pipeline is not evidence the fix is still there; `grep -c` for `env_read()` inside a `thread::spawn` is.

If the in-thread placement is deliberate and I am missing a reason for it, say so and I will stop re-applying it. My reading is that the comment there — "a guard in the test body cannot cover it" — does not hold for these four tests, because a guard does not have to be held *by* the thread doing the lookup, only *while* it runs, and all four join their server thread a few lines later, as the surrounding code itself notes.

`irlume-daemon`: 105 passed, 0 failed on the current merge head.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #222, opened 2026-08-02
The glint cue measures a peak, so it cannot tell a corneal highlight from a retinal return, and only one of those is hard to fake


archledger:
Research finding, not a measurement on irlume hardware. Filed so the reasoning survives.

## Two different physical effects land in the same cue

`eye_glint` takes the PEAK grey level inside a fixed 8-pixel-radius square at each eye landmark (`crates/irlume-auth/src/lib.rs`, `GLINT_SEARCH_RADIUS_PX = 8`), and `GLINT_MIN = 180` gates on that peak. A peak carries no information about the extent or shape of what produced it, and two very different things produce a bright spot at the eye under active IR:

- **Corneal specular reflection.** A first-surface mirror reflection of the emitter off the tear film. Small, a few pixels. A glossy print or a screen reproduces this trivially, which is consistent with the documented result that a glossy IR print cleared the centre/edge cue in 69 of 70 trials (`docs/pad-results/2026-06-30-ir-liveness-selftest.md`).
- **Retinal retroreflection, the "bright pupil" effect.** With the source close to the lens axis, light entering the pupil is reflected back off the retina and the whole pupil reads as a bright disc. The gaze-tracking literature uses exactly this to separate retro-reflective from non-retro-reflective objects, and eye trackers implement it with coaxial and non-coaxial LED rings alternating to produce bright-pupil and dark-pupil frames ([Nguyen et al., pupil and glint detection with NIR LED array](https://pmc.ncbi.nlm.nih.gov/articles/PMC4721713/)). A print has no retina and is not retro-reflective, so it cannot produce this; a patent for face anti-spoofing builds on the same principle, adding that light near 942 nm is absorbed by the eye's water content and so does not reach the retina, which distinguishes a live eye from a reproduction ([US 11,216,647](https://image-ppubs.uspto.gov/dirsearch-public/print/downloadPdf/11216647)).

So one of the two signals feeding this cue is fakeable by gloss and the other is not, and irlume currently cannot tell them apart because it only records the maximum.

## Why this is worth testing rather than assuming

- **Extent separates them.** A corneal glint is a point; a bright pupil is a disc the size of the pupil, which at typical login framing is tens of pixels across. Measuring the extent of the bright region, not just its peak, would distinguish them with no model and no new hardware.
[...]

PR #382, opened 2026-08-08
A glint reading that railed measured nothing (#222)

archledger:
## What & why

Part of #222, the remainder you named after settling the retroreflection half: *"the actionable remainder is the peak semantics (saturation-aware reading plus the least-clipped-frame selection on #221)"*. The #221 half landed; this is the peak semantics.

`eye_glint` returned the eye-window maximum with no notion of a ceiling, so a peak that reached the negotiated format's white level was recorded as the **strongest possible reading** rather than as a reading that established nothing. A clipped sample says the true value was at least the ceiling and never what it was.

### The corpora already in this repo show how total the conflation is

I computed these from the committed files rather than quoting them:

| file | records | railed at 255 | `glint_present` | present **and** unpegged |
|---|---|---|---|---|
| `2026-08-04-occluder-gate.jsonl` | 26 | 8 | 8 | **0** |
| `2026-08-02-center-edge-corpus.jsonl` | 104 | 16 | 34 | 20 |

In the occluder-gate file `glint_present` is a perfect synonym for "the peak railed": every record carrying a glint is pegged at exactly 255, and the highest genuine reading is 126, far under `GLINT_MIN` (180). In the older center-edge corpus a real sub-ceiling population does exist (20 records, max 253), so marking the railed ones absent **separates** the populations rather than erasing them.

That matches what you measured before settling the issue: with glasses on the peak pinned at 255 in all 30 frames, reading the lens specular rather than the cornea.

### #237's exposure gate structurally cannot catch this

On those railed records the **face** region clipped only 0.42%–1.57%, far below `IR_SATURATED_FRAC_MAX` (0.10), so `ir_exposure_ok` is true and the capture is judged normally. A specular blob inside a 17×17 eye window is a rounding error against a whole face region. Distinct defect, not a duplicate.

### The change

`Signals::ir_eye_glint` becomes `Option<f32>`, returning `None` for a railed peak, a missing IR face, or the RGB-only path. The corpus writes `ir_glint: null`, and a new `cues.glint_readable` keeps "read and dim" distinct from "not readable" so the record does not simply trade one conflation for another. The three display sites print `n/a`, the idiom `ir_clipped` already uses beside them.

**No number is invented.** `GLINT_MIN` stays 180, `IR_SATURATED_FRAC_MAX` stays 0.10. The ceiling is `clipping_white_level`, which already derives it from the negotiated format and was **already passed to `saturated_frac_of` at both call sites**, so the value was sitting in scope. In-band versus at-the-rail is a property of the format, not a threshold anyone has to measure. A format that cannot name its ceiling (Grey16, NV12, YUYV) passes the peak through unchanged, which is #237's own settled precedent.

The ceiling test reads the **raw** frame, for the reason stated beside `ir_saturated_frac`: ambient subtraction moves a railed 255 to 254, so a subtracted frame would quietly stop reading as railed (#238 review).

## What this is NOT

**It moves no verdict, and it is not a security fix.** `glint_present` reaches the PAD corpus and a dev probe print; both assignments sit after every `Spoof` return in their evaluators, so the cue is structurally incapable of changing an outcome. A regression test locks that on both evaluators, so making the cue honest cannot quietly promote it.

The value is narrower and worth stating plainly: any future re-derivation of `GLINT_MIN` would be fitted to this corpus, and two populations cannot be observed in a variable pinned at the sensor's limit.

## A finding this surfaced that I did NOT act on

The same ceiling-unaware peak is read by `eye_open_at` (`irlume-auth/src/lib.rs:3816`), which returns `peak >= EYE_OPEN_PEAK_MIN` (200.0), feeding `Assessment::eyes_open`, which **is** enforced as a real deny at `:2516` (`if enr.require_eyes_open && !a.eyes_open`).

A railed 255 satisfies `255 >= 200` unconditionally, so on a frame whose eye windows clip, that opt-in gate can be satisfied by **saturation rather than by an open eye**, which is a fail-open.

I deliberately left it alone, because the fix is a posture decision that is yours: your own measurement records the peak pinned at 255 in **all 30** frames with glasses on, so treating railed as "not open" would deny effectively every glasses wearer who has `require_eyes_open` enabled. That is a lockout-shaped change and it needs your call, not mine. Worth its own issue.

## Testing

Container-run, full workspace, every result line read.

- **1,280 passed** against a **1,277 baseline measured on this same base**, so the +3 are the new tests. The only two failures are the container's pre-existing root-only artifacts (`ir_emitter::…failed_save_keeps_the_record`, `tui::…models_screen_unprivileged…`), identical before and after, and both pass in CI which runs non-root.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets`, and `RUSTDOCFLAGS="-D warnings" cargo doc` all clean.
- New tests: the ceiling arm at full range and limited range (235), the unmeasurable-format arm pinning #237's precedent, the no-face arm, the three-way readable/present separation on **both** evaluators, and a regression lock that an unreadable glint changes no verdict.
- The ceiling test is **proven to have teeth**: a mutant that ignores the white level makes it fail (`left: Some(255.0), right: None`).

- [x] `cargo test --workspace` passes (1280; 2 pre-existing root-only failures)
- [x] `cargo clippy --workspace --all-targets` clean
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc` clean
- [x] Docs/CHANGELOG updated (PAD_SELFTEST cue table, CHANGELOG entry recording the corpus counts and that pre-change records are railed-contaminated)
- [ ] Hardware-validated (not required: no verdict moves, and the ceiling comes from the negotiated format rather than a measurement)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,41 @@ All notable changes to irlume are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- **The IR eye-glint cue recorded the sensor's ceiling as the strongest possible
+  reading.** `eye_glint` returned the window maximum with no notion of a
+  ceiling, so a peak that railed at the negotiated format's white level was
+  stored as a measurement rather than as a reading that established nothing. A
+  clipped sample says the true value was AT LEAST the ceiling and never what it
+  was, and a maximum is exactly the statistic that destroys.
+
+  Measured in the corpora already committed here. In
+  `docs/pad-results/2026-08-04-occluder-gate.jsonl` all 8 records that recorded
+  a glint are railed at exactly 255, and no record has `glint_present` with an
+  unpegged peak, so "glint present" and "the peak railed" named the same set;
+  the highest genuine reading in that file is 126, far under `GLINT_MIN` (180).
+  In `docs/pad-results/2026-08-02-center-edge-corpus.jsonl` 16 of 104 are railed
+  while 20 records carry a genuine sub-ceiling glint, so a real population does
+  exist and marking the railed ones absent separates it rather than erasing it.
+  This is the shape #222 named after measuring all three supported modules.
+
+  #237's exposure gate cannot catch it: on those railed records the FACE region
+  clipped 0.42% to 1.57%, far below `IR_SATURATED_FRAC_MAX` (0.10), because a
+  specular blob inside a 17x17 eye window is a rounding error against a whole
+  face. `Signals::ir_eye_glint` is now `Option<f32>`, `None` for a railed peak,
+  a missing IR face, or the RGB-only path; the corpus writes `ir_glint: null`
+  and a new `cues.glint_readable` separates "read and dim" from "not readable".
+  No threshold moved: `GLINT_MIN` stays 180, and the ceiling comes from
+  `clipping_white_level`, which was already plumbed to both call sites. A format
+  that cannot name its ceiling (Grey16, NV12, YUYV) passes the peak through
+  unchanged, which is #237's own settled precedent.
+
+  No authentication verdict changes: the cue reaches the PAD corpus and a dev
+  probe print, and nothing else. Records written BEFORE this change keep the old
+  semantics, so any future re-derivation of `GLINT_MIN` must treat the existing
+  corpora as railed-contaminated.
+
 ### Security
 
 - **One socket request reached the keyring with an unscreened username.**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,10 +24,14 @@ All notable changes to irlume are documented here. This project adheres to
   exist and marking the railed ones absent separates it rather than erasing it.
   This is the shape #222 named after measuring all three supported modules.
 
-  #237's exposure gate cannot catch it: on those railed records the FACE region
-  clipped 0.42% to 1.57%, far below `IR_SATURATED_FRAC_MAX` (0.10), because a
-  specular blob inside a 17x17 eye window is a rounding error against a whole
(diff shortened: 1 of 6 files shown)


archledger:
Hardware validation, after merge. Zenbook, ASUS FHD IR module, `/dev/video2` GREY 640x400, merged main at `325d7ae`.

Three consecutive `liveness` probes exercise all three states this change introduced:

```
run 1  eye-glint 145  glint=false  glint_readable=true    read, and under GLINT_MIN
run 2  eye-glint 186  glint=true   glint_readable=true    read, and over it
run 3  eye-glint n/a  glint=false  glint_readable=false   not readable (no IR face)
```

That is the separation the change exists for, on real captures. Before it, run 1 and run 3 both recorded `glint_present: false` and were indistinguishable in the corpus: one was a genuine dim reading at 145, the other was no reading at all. A re-derivation of `GLINT_MIN` fitted to that corpus was fitting the two together.

The railed case did not occur in these three runs, which is itself worth recording: the peaks came back at 145 and 186 against a format ceiling of 255, so this module reads sub-ceiling glints in ordinary room light. The all-255 population documented on #222 was measured with glasses on, reading the lens specular rather than the cornea.

The frames were not clean of clipping generally: the raw IR frame reported `max 255` with a face-region clipping fraction of 1.7% in an earlier run, well under `IR_SATURATED_FRAC_MAX` (0.10). So the eye window and the face region really do rail independently, which is the premise of "#237's exposure gate structurally cannot catch this".

Verdicts were `Live`, `Live`, and `Spoof: no face in IR`, all unchanged from what the cue values imply, consistent with the claim that this moves no verdict.

Issue #227, opened 2026-08-02
doctor reports a camera it cannot open as a camera that does not exist


archledger:
## What happens

`classify()` returns `Role::Other` both when a node advertises neither colour nor grey formats and when the node could not be opened or its formats could not be enumerated:

```rust
pub fn classify(device: &str) -> Role {
    let Ok(dev) = Device::with_path(device) else {
        return Role::Other;
    };
    let Ok(formats) = Capture::enum_formats(&dev) else {
        return Role::Other;
    };
```

`discover_nodes()` then drops every `Other`, so a node that could not be read disappears from the list rather than appearing with a reason. `irlume doctor` prints what is left, and a machine whose cameras are all unreadable gets `(none found under /dev/video0..9)`.

Demonstrated on the archhost runner, which has eleven video nodes including two the same binary classifies as `Ir` when run by a user in the `video` group:

```
$ ls /dev/video*
/dev/video0 /dev/video1 /dev/video10 /dev/video2 ... /dev/video9

$ sudo -u nobody irlume doctor
[doctor] camera nodes (classified by pixel format):
  (none found under /dev/video0..9)
```

The user is told the hardware is missing. The actual condition is that this account cannot open the nodes.

## Why it matters

`EACCES` (not in the `video` group) and `EBUSY` (another process is streaming) are the two most common camera complaints, and both render as absent hardware. Someone whose camera is held by a video call is sent looking for a driver problem they do not have.

The codebase already commits to the opposite rule a few lines below, in `control_read_failure_means_absent`, added by the #193 review: a failure to observe a control must never be reported as the control being absent. The same distinction is missing one function up.

This also produced a scheduled Hardware suite failure on 2026-08-01 (run 30722301904), where the step reported `doctor classified 2 camera node(s) but none as Ir on a runner that has one` while the IR node had been present since 20:24 that evening.

## Fix

Keep the three outcomes apart: a classified node, a node that answered and is neither, and a node that could not be read. Report the third with its errno so doctor can say which of `EACCES`, `EBUSY`, `ENODEV` it hit, and map the common ones to the action that resolves them (`video` group membership, another process holding the device).

The `camera-nodes` check state should distinguish "no nodes on this machine" from "nodes present, none readable" instead of reporting `Fail` with `(none found)` for both.

## Second defect in the same function

[...]

PR #383, opened 2026-08-08
The nightly camera stage names its IR node from a parse it can test

archledger:
## What & why

The nightly hardware suite has failed **every scheduled run since 2026-08-01**: runs `30722301904`, `30771201517`, `30860760581`, `30958682250`, `31054740453`, `31138203146`, `31224343018`, `31281803049`. Eight nights, one error:

```
::error::doctor classified 5 camera node(s) but none as Ir on a runner that has one
```

printed directly above the doctor output it was reading:

```
[doctor] camera nodes (classified by pixel format):
  /dev/video0: Rgb (uvcvideo, USB)
  /dev/video2: Ir (uvcvideo, USB)
```

`42c03c6` (2026-07-31, #195) appended the backend to every node line. The workflow matched `/: Ir$/`, anchored at end of line, and the role token is no longer the last thing on it. The camera is fine; the parser was reading a format that stopped existing the day before the first red run.

What that cost: the strobe-burst capture is the only check anywhere in this repo that drives the real IR emitter, and `hardware-suite.yml` is the only workflow that contains it. It has not run for eight nights.

## The parse was half the defect

The other half is that the inline version had no way to say "I could not read this". Its only two outcomes were "no camera" and "no IR camera", so a format change had to arrive dressed as a hardware fact. That is why this sat for eight days: the error names a plausible condition, and the runner really does sometimes have its camera borrowed.

`scripts/ir-node-from-doctor.sh` gives the caller three:

| exit | meaning | workflow does |
|---|---|---|
| 0 | IR node found, path on stdout | run the burst |
| 10 | doctor listed no camera nodes | clean skip (the borrowed-camera case) |
| 11 | nodes listed, none classified `Ir` | fail loudly |
| 12 | output did not parse | fail loudly, saying in words that this is **not** a missing camera |

Exit 12 fires on three shapes: the section header gone, a role token this script does not know (so a new `Role` variant surfaces here rather than being silently skipped), and node lines present with none parsing, which is exactly what happened.

## Why a file, and how it is tested

Moving the parse out of YAML is what makes it testable at all. `--self-test` runs in the `units` job on every PR, fixtures only, no camera and no Rust build, in the idiom `run-tests-guarded.sh --self-test` and `verify-ppa-publish.py --self-test` already use in the same job.

Eleven cases: the current line, the pre-#195 bare line (so the parse stays a superset rather than trading one exact format for another), the privacy-switch and backend-unknown annotations, a deeper-indented continuation line that must not count as a node, an IR node after the section closed that must not be picked up, and each of the three empty outcomes.

Proven to have teeth rather than assumed:

- Restoring the anchored `/: Ir$/` match fails **6 of the 11** cases.
- That same mutant, run over the doctor output captured from run `31281803049`, returns 12 and says `doctor listed 5 node line(s) and none parsed`. The honest diagnosis was available all along; the old code had no exit code for it.
- The fixed script over that same real output returns `/dev/video2`.

The self-test also earned its keep during development: it caught two bugs in my own first draft. `read` with a tab `IFS` strips the leading empty field, because tab is IFS whitespace, so "no IR node" shifted into "the IR node is `2`", which is the same class of defect this script exists to stop. Separator is now `|`. A `trap ... RETURN` set inside a function stays installed and fires again for the next one, dereferencing a `local` that is already gone.

Also simulated the four workflow branches end to end against fixtures, checking the exit code and the message each produces.

## Testing

- `./scripts/ir-node-from-doctor.sh --self-test`: 11 passed, 0 failed.
- Against the real captured nightly doctor output: `/dev/video2`, exit 0.
- Mutation as described above.
- `shellcheck` clean on the script and on the rewritten workflow step.
- Both workflow files parse.

Not covered: nothing hosted has a camera, so the fixtures are captured doctor output, not live output. A doctor change the fixtures do not model would still reach the nightly. That is the reason an unparseable result is loud instead of skipped.

I would like to run the suite once by `workflow_dispatch` after this merges, to see the camera stage green on archhost rather than argue it from fixtures.

- [x] `shellcheck` clean
- [x] self-test passes and fails on the reverted parse
- [x] verified against the doctor output from the failing run
- [x] `scripts/README.md` updated
- [ ] Hardware-validated: pending a `workflow_dispatch` run on archhost after merge

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -700,6 +700,15 @@ jobs:
       - name: PPA publish verifier (verdict logic)
         run: python3 scripts/verify-ppa-publish.py --self-test
 
+      # The nightly hardware suite finds its IR node by parsing `irlume doctor`,
+      # and that parse is why the camera stage did not run for the eight nights
+      # after 42c03c6 changed the node line. Nothing hosted has a camera, so the
+      # parse itself is what gets pinned: fixtures only, seconds, every PR.
+      # Here and not also in `coverage`, which carries copies of the two steps
+      # above; one job proving a fixture parse is enough.
+      - name: doctor IR-node parser (the nightly hardware suite depends on it)
+        run: ./scripts/ir-node-from-doctor.sh --self-test
+
       - name: systemd-analyze verify (unit syntax)
         run: |
           set -euo pipefail
--- a/.github/workflows/hardware-suite.yml
+++ b/.github/workflows/hardware-suite.yml
@@ -97,26 +97,38 @@ jobs:
             echo "::error::irlume doctor failed — not a clean camera-absent skip"
             cat "$doc"; exit 1
           fi
-          # An empty result has two very different causes, and reading both as
-          # "no camera" turned a classifier or doctor-format regression into a
-          # clean skip on the ONLY lane with a real camera. Tell them apart by
-          # asking whether doctor classified ANY node: none at all is a genuine
-          # absent-camera skip, while nodes present with no Ir among them is a
-          # regression on a runner that is supposed to have one.
-          if ! grep -q "camera nodes (classified by pixel format)" "$doc"; then
-            echo "::error::doctor produced no camera-nodes section — its output format changed"
-            cat "$doc"; exit 1
-          fi
-          nodes=$(awk '/camera nodes \(classified by pixel format\)/{f=1;next} f&&/^  \/dev\//{n++} f&&!/^  /{f=0} END{print n+0}' "$doc")
-          ir=$(awk -F: '/: Ir$/{print $1}' "$doc" | tr -d " " | head -1)
-          if [ -z "$ir" ]; then
-            if [ "$nodes" -eq 0 ]; then
+          # An empty result has three very different causes, and the parser
+          # that used to live inline here could only express two of them. It
+          # read a changed doctor format as "no IR node", so from 2026-08-01
+          # every scheduled run failed with "classified 5 camera node(s) but
+          # none as Ir" while the doctor output printed directly beneath it
+          # said "/dev/video2: Ir (uvcvideo, USB)". Eight nights of the only
+          # check that drives the real IR emitter, not run.
+          #
+          # The parse now lives in a script with a self-test that hosted CI
+          # runs on every PR, so the next doctor format change fails in
+          # seconds instead of parking this stage until someone reads a log.
+          ir=""; rc=0
+          ir=$(scripts/ir-node-from-doctor.sh "$doc") || rc=$?
+          case "$rc" in
+            0) ;;
+            10)
               echo "::notice::no camera nodes at all — camera stage skipped (borrowed for the OS bench?)"
               exit 0
-            fi
-            echo "::error::doctor classified $nodes camera node(s) but none as Ir on a runner that has one"
-            cat "$doc"; exit 1
-          fi
+              ;;
+            11)
+              echo "::error::doctor classified camera node(s) but none as Ir on a runner that has one"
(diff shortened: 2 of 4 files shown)


archledger:
Validated against live doctor output on a second machine, since the fixtures in the self-test were captured from archhost's job log and a fixture cannot prove the format it was copied from is still what the binary emits.

Zenbook, ASUS FHD IR module, the packaged `irlume 0.9.0` rather than a build from this branch:

```
[doctor] camera nodes (classified by pixel format):
  /dev/video0: Rgb (uvcvideo, USB)
  /dev/video2: Ir (uvcvideo, USB)
```

- Old CI parser (`awk -F: '/: Ir$/{print $1}'`): empty. It would have failed here too, on a machine whose IR camera is sitting right there.
- `scripts/ir-node-from-doctor.sh`: `/dev/video2`, exit 0.

So the break reproduces on both machines that have an IR camera, and the fix resolves both. Two different modules (ASUS FHD IR at 640x400 GREY here, NexiGo N930W at 340x340 GREY on archhost), same result.

Still outstanding, and the reason the hardware box above is unticked: this proves the parse, not the burst. The strobe capture it feeds has not run since 2026-07-31, so nothing here says the emitter still passes the mean-spread check. I would like to `workflow_dispatch` the suite on archhost once this merges and report what the camera stage actually does.

archledger:
Codex round: DO NOT SHIP on one real finding, now fixed in a6ebac2.

**The finding.** My script inferred absence from a node count of zero. doctor keeps three states, not two: classified nodes, unreadable nodes, and a listing error. It prints `(no /dev/video* nodes on this machine)` under exactly one condition (no listing error AND no nodes AND no unreadable nodes) and renders the other two as warning lines my parser counted as nothing.

So a camera that is present but denied by `EACCES`, held by another process, or lost to a failed `/dev` walk returned exit 10, the workflow took exit 0, and the nightly went green having never driven the emitter. Reproduced before fixing:

```
[doctor] camera nodes (classified by pixel format):
  ⚠ /dev/video0, /dev/video2: could not be opened: permission denied
```
```
doctor listed no camera nodes
exit=10
```

That is a worse version of the bug this PR exists to fix. The old parser at least went red; mine would have gone green.

**Where it came from.** It is #227 one layer up, and doctor states the rule directly above the code I was reading, at `crates/irlume-cli/src/main.rs:3822`:

> "Could not look" is not "nothing there". Reporting a /dev that would not list as a machine with no cameras is the same mistake at one level up.

I read that comment and wrote the inference underneath it anyway.

**The fix.** Absence is taken only from doctor's own marker, never inferred. A section with no node lines and no marker is a new outcome, exit 13, and the workflow fails on it saying the emitter test did not run and this is not an absent camera. Separate from 12 on purpose: 12 means fix this parser, 13 means go look at the machine.

Three fixtures now cover what must never reach the skip: unreadable nodes, an unwalkable `/dev`, and an empty section. 14 cases, 14 pass.

Proven load-bearing rather than asserted: replacing the marker test with an unconditional `true` fails exactly those three and leaves the other eleven green. Both real doctor captures (archhost, Zenbook) still resolve to `/dev/video2`, and genuine absence still skips at exit 0.

**What I take from it.** A positive assertion of absence is evidence; a count of zero is not. Only the first distinguishes "nothing is there" from "I could not see". The self-test has now caught three defects of mine on this one script, which is the argument for the file existing.

Codex cleared the rest, including that the eleven original cases can all fail, that `KNOWN_ROLES` matches the current `Role` enum, and that the workflow preserves the exit status rather than losing it through a pipeline. Its two open items are unchanged: no live `EACCES` reproduction on hardware, and the `workflow_dispatch` run still to come after merge.

PR #387, opened 2026-08-09
The strobe check drives the emitter it claims to test (#384)

### LINKED ISSUE #384: The nightly strobe-burst check runs with irlume's emitter control disabled by a lock permission error
Found while validating #383 on archhost: the nightly's strobe-burst step runs with irlume's emitter control disabled, and passes anyway.

## What the run printed

Hardware suite run `31284722953`, 2026-08-08, the first successful camera stage since 2026-07-31:

```
IR node: /dev/video2
irlume: could not check whether an interrupted emitter setup left this camera changed
(lock the camera: open /run/lock/irlume-emitter-aea4046c…lock: Permission denied (os error 13)).
irlume will not write to this camera's emitter from this process until the lock error is
resolved and the journal can be checked, so IR face authentication will not light here.
captured 6 frames
mean spread: 17.9
```

The step passed: 6 frames, and a mean spread of 17.9 against its `< 5` failure floor.

## Why the permission error happens

On archhost:

```
drwxr-xr-x 2 root root  /run/lock
-rw-r----- 1 root root  /run/lock/irlume-emitter-aea4046c….lock   (2026-08-07)
-rw-r----- 1 root root  /run/lock/irlume-emitter-c7c0b0ed….lock   (2026-08-06)
```

`/run/lock` is 0755 root-owned and the lock files are 0640 root-owned. The runner account (`archledger`, groups include `video` and `tss`, not `root`) cannot open them, so any non-root invocation takes the refusal branch. The daemon runs as root and is unaffected; the CLI and the examples do not.

## Why this matters beyond the runner

Two separate things, and I am more confident about the first.

**The strobe check proves less than it appears to.** Its stated purpose is that the emitter must actually strobe, and a dead emitter shows as a flat dark burst. But on this run irlume never wrote an emitter control, so the 17.9 spread came from whatever state the emitter was already in, not from irlume driving it. A regression in irlume's own emitter path would not move that number. The check is currently a camera-delivers-frames test wearing an emitter test's name.

**Any non-root user meets this too.** The lock files are created root-owned by the daemon, so a user running a capture tool or a diagnostic hits the same refusal. The refusal direction is right (it will not write the emitter when it cannot check the undo journal), but the outcome is that the documented behaviour depends on who is asking, and nothing tells the user that up front.

## What I have not established

Whether the lock files are meant to be root-only. The undo journal they guard (#181) is root-written, so 0640 may be deliberate and the correct fix may be for the non-root path to degrade explicitly rather than for the permissions to change. I have not read the emitter journal design closely enough to say which.

Nor have I checked whether the spread would still be 17.9 with the emitter genuinely off; that needs a run with the emitter deliberately disabled, which I did not want to do to the runner unattended.

Filed rather than fixed because both halves are design calls.

archledger:
## What & why

Closes #384.

The nightly's emitter check passed on 2026-08-08 with irlume's emitter control disabled for the entire run, and nothing in the output said so. This is the check the whole hardware lane exists for: it is the only thing in the repo that drives the real IR emitter.

irlume guards its emitter bookkeeping with a lock in `/run/lock`. The daemon creates that file under `UMask=0027`, so it lands `0640 root:root`:

```
drwxr-xr-x. 6 root root  /run/lock
-rw-r-----. 1 root root  /run/lock/irlume-emitter-209e9051….lock
```

The runner account cannot open it, so irlume takes its refusal branch and declines to write the emitter for the rest of the process. The step never read stderr, and the spread it measures is whatever state the emitter was left in by the last root process. A regression in irlume's own emitter path would not move it.

## Measured, with a control

Zenbook, ASUS FHD IR module, same binary, back to back:

| run | stderr | mean spread |
|---|---|---|
| unprivileged | `will not write to this camera's emitter` | **49.6** |
| under `sudo -n` | clean | **57.9** |

The existing check passes at 49.6 having proven nothing. Both clear its `< 5` floor, which is exactly the problem: the number cannot distinguish the two runs, so the floor was never testing what its comment claims.

## The change

The burst runs under `sudo -n` when available. That is also the production context: `irlumed` drives the emitter as root, so this exercises the path users actually get rather than one no shipped component uses. archhost's runner has passwordless sudo and is in `video`; without it the step warns and continues unprivileged.

The step then reads stderr. If irlume reports it could not take the lock, the step fails and says the spread below it is the emitter's leftover state rather than evidence. Same rule as the node parse above it and as #227: a check that could not run is not a check that passed.

Verified against real output rather than a guessed pattern: the guard fires on the actual unprivileged stderr and does not fire on the privileged run.

## Scope of the permission gap, narrowed

While checking this I established the user-facing path is unaffected. `ir-setup` goes through the daemon, which runs as root and gates on it. What cannot take the lock is the `IRLUME_DEV=1` tooling that opens the camera directly, which is dev tools and this CI step.

So the permission question the issue raised is a tooling question, not a user-facing one, and this makes the tooling honest without touching the lock's mode or ownership. I would rather not widen a lock on a security daemon to solve a CI problem.

## Testing

- Guard fires on the real unprivileged stderr; does not fire under sudo.
- Both burst runs captured 6 frames at 640x400.
- Workflow parses.

Not covered: whether the spread collapses with the emitter genuinely off. That needs a run with it deliberately disabled, which I did not want to do to the runner unattended. So the check now proves "irlume drove the emitter and the scene strobed" rather than "irlume is what made it strobe".

- [x] Measured before and after, with a control
- [x] Guard verified against real output
- [ ] Hardware-validated on archhost: pending a `workflow_dispatch` after merge

### DIFF
--- a/.github/workflows/hardware-suite.yml
+++ b/.github/workflows/hardware-suite.yml
@@ -160,18 +160,31 @@ jobs:
             exit 1
           fi
           err="$(mktemp)"
-          if sudo -n true 2>/dev/null; then
-            sudo -n "$burst" "$out" "$ir" 6 2>"$err"
-          else
-            echo "::warning::no passwordless sudo; running the burst unprivileged"
-            "$burst" "$out" "$ir" 6 2>"$err"
+          # Privilege is required, not preferred. The emitter lock lives in
+          # /run/lock and the daemon creates it 0640 root:root under UMask=0027,
+          # so an unprivileged run cannot take it and irlume declines to write.
+          # Root is also the production context: irlumed drives the emitter as
+          # root.
+          if ! sudo -n true 2>/dev/null; then
+            echo "::error::passwordless sudo is required to drive the emitter here (#384)"
+            exit 1
           fi
+          sudo -n env IRLUME_LOG_EMITTER_WRITES=1 "$burst" "$out" "$ir" 6 2>"$err"
           cat "$err"
-          # "Could not manage the emitter" must not read as "the emitter is
-          # fine". Same rule as the node parse above and as #227: a check that
-          # could not run is not a check that passed.
-          if grep -q "will not write to this camera.s emitter" "$err"; then
-            echo "::error::irlume could not take the emitter lock, so it never drove the emitter. Any spread below is the emitter's leftover state, not evidence that irlume can light it (#384)"
+          # A POSITIVE marker, not the absence of one refusal string. The first
+          # version of this guard grepped for the lock refusal, which covers one
+          # of at least six inert paths through `ir_emitter::enable`: an
+          # unreadable USB identity, a recovery reporting Busy or
+          # OwnerStillRunning (both silent by design), no applicable control, and
+          # a failed apply_device_default or apply_known_payload. Any of those
+          # captures frames that look exactly like a successful run.
+          #
+          # `AlreadyHeld` is deliberately NOT accepted: it shows the control
+          # holds the wanted value, which is not the same as this invocation
+          # having driven it, and driving it is what this step claims to test
+          # (#384 review).
+          if ! grep -Fxq "irlume: capture emitter write completed" "$err"; then
+            echo "::error::irlume did not complete the emitter write, so the spread below is the emitter's leftover state rather than evidence irlume can light it (#384)"
             exit 1
           fi
           frames=$(find "$out" -name "frame*.pgm" | wc -l)
--- a/.github/workflows/hardware-suite.yml
+++ b/.github/workflows/hardware-suite.yml
@@ -140,7 +140,40 @@ jobs:
           echo "IR node: $ir"
           out=$(mktemp -d)
           trap 'rm -rf "$out"' EXIT
-          cargo run -q -p irlume-camera --example burst_dump -- "$out" "$ir" 6
+          # Run the burst as ROOT when we can. irlume guards its emitter
+          # bookkeeping with a lock in /run/lock, and the daemon creates that
+          # file under `UMask=0027`, so it lands 0640 root:root. The runner
+          # account cannot open it, and irlume then REFUSES to write the
+          # emitter for the rest of the process, saying so on stderr.
+          #
+          # That refusal is why this step needs to be explicit about privilege.
+          # The check below asserts the emitter strobes, and an unprivileged
+          # run measures whatever state the emitter was left in by the last
+          # root process rather than anything irlume did. It passed at a spread
+          # of 17.9 on 2026-08-08 with irlume's emitter control disabled
+          # throughout (#384). Root is also the production context: irlumed
+          # runs as root, so this exercises the path users actually get.
+          burst="${CARGO_TARGET_DIR:-target}/debug/examples/burst_dump"
+          cargo build -q -p irlume-camera --example burst_dump
+          if [ ! -x "$burst" ]; then
+            echo "::error::burst_dump not built at $burst"
+            exit 1
+          fi
+          err="$(mktemp)"
+          if sudo -n true 2>/dev/null; then
+            sudo -n "$burst" "$out" "$ir" 6 2>"$err"
+          else
+            echo "::warning::no passwordless sudo; running the burst unprivileged"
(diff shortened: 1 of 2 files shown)


archledger:
Correcting my own justification for this PR, from a control run I should have done before opening it.

I claimed running the burst privileged makes the check prove irlume drove the emitter. Measured on the Zenbook's ASUS FHD IR module, that is not what the spread shows, and the change of privilege makes no difference to it at all.

Three runs each, same binary, back to back:

```
irlume manages the emitter        spread 51.5, 53.1, 52.4
IRLUME_IR_EMITTER=off             spread 53.0, 56.0, 52.6
```

The per-frame means say why:

```
managed        46.0  0.0  52.2  0.0  51.4  0.0
override off   56.8  0.0  55.1  0.0  53.9  0.0
```

Identical alternation, lit frame then a frame at exactly 0.0, in both cases.

**The instrument checks out**, which is the part that makes this a finding rather than a mistake. `IRLUME_IR_EMITTER=off` resolves to `OverrideSetting::Disabled` in `ir_emitter::enable`, which returns `StreamMode::inert()` before any write, so irlume really did drive nothing on the second set. `capture_raw_burst_timed` does call `ir_emitter::enable`, so the path is exercised. And the burst still strobed.

So on this module the alternation is produced by the camera, not by irlume's extension-unit write. The `mean spread >= 5` assertion measures that the camera is delivering alternating frames. It cannot detect a regression in irlume's emitter control, whether the run is privileged or not.

**What this does and does not change here.** The two changes in this PR are still right for their own reasons: irlume silently declining to manage the emitter should fail rather than pass, and root is the production context because irlumed runs as root. What is wrong is my sentence claiming the result proves irlume drove the emitter. It proves irlume was ABLE to, and that the camera strobed.

**Not yet established.** This is the ASUS module. The nightly runs on archhost's NexiGo N930W, which is a different module and could well behave differently; a self-strobing Hello camera and one that needs the host to toggle the emitter would both produce a spread. I have not been able to run the same control there because the runner has been busy with this PR's own CI. I will post the NexiGo numbers before asking for a merge, since they decide whether the check has any value on the machine that actually runs it.

If the NexiGo behaves the same way, the honest follow-up is that this check needs a real discriminator or a narrower claim in its comment, and I will file that rather than leave the comment overstating what the number means.

archledger:
NexiGo numbers, as promised, from archhost. They confirm the finding on the module the nightly actually uses, and they corrected an intermediate reading of mine along the way.

## First attempt, and why it was wrong

Alternating the two conditions gave what looked like a clean result: with irlume managing the emitter the brightest frame read 141.0, with `IRLUME_IR_EMITTER=off` it read 27.8. That reads as "irlume's write makes the emitter five times brighter".

It is an artifact. The runs alternated conditions, and emitter state persists between processes, so each run could be inheriting what the previous one left.

## Redone within a condition

Five consecutive runs of each, no switching:

```
five OFF runs, brightest frame       134.5  31.6  139.9  27.8  134.5
five MANAGED runs, brightest frame    31.1 139.5   29.1 148.0   25.1
```

Both conditions alternate between roughly 30 and roughly 135. The camera cycles its own state between capture sessions, and which end a run lands on has nothing to do with whether irlume drove the emitter. My 141-versus-27.8 comparison was two samples from the same alternation, one from each end.

## What this establishes

The `mean spread >= 5` assertion cannot detect a regression in irlume's emitter control, on the NexiGo any more than on the ASUS. Every one of these twenty runs clears that floor by a wide margin, in both conditions. The check measures that the camera is delivering alternating frames, which it does on its own.

This is the same self-strobing behaviour recorded on #264, seen from the other side.

## What I am doing about it

The two changes in this PR stand on their own reasons and I am leaving them: irlume silently declining to manage the emitter should fail rather than pass, and root is the production context because irlumed runs as root. Neither claim depends on the spread meaning what its comment says.

What I am changing is the comment, so it stops asserting something these numbers refute, and I am filing the real problem separately: this check needs a discriminator that actually depends on irlume's emitter write, or it should say plainly that it tests frame delivery. Writing that as a follow-up rather than inventing a threshold here, because picking one needs a measurement campaign across both modules and this PR should not carry it.

PR #388, opened 2026-08-09
Take the environment guard before the spawn, not on the thread being timed

archledger:
Follow-up to #380, which is merged. The guard it added is right; where four of the tests take it is not.

## What is wrong

#380 gave every daemon test that reaches a passwd lookup a shared read guard. Four of them take it inside the spawned `serve` closure, with a comment arguing it has to live there because the lookup runs there:

```rust
let server = std::thread::spawn(move || {
    // The passwd lookup this reaches runs HERE, on a thread that
    // outlives the test body, so a guard in the test body cannot
    // cover it.
    let _passwd = crate::test_support::env_read();
    serve(theirs, &a, &ready).unwrap();
});
```

The lookup does run there. The premise underneath does not hold: the thread does not outlive the test body, because each of these tests joins it before returning, and `every_test_that_resolves_a_user_holds_the_env_lock` already refuses to let a spawned `serve` go unjoined. A binding on the test thread covers the whole thing.

Taking it inside the closure is what made the suite flaky. `std::sync::RwLock` is write-preferring, so a waiting writer stops a new reader from entering. The server thread queues behind the suite's forty-odd `env_lock()` writers while the client that spawned it is already counting down `set_read_timeout`. When the queue outlasts the deadline, `read_line` returns `WouldBlock`:

```
thread 'tests::a_request_needing_the_engine_is_refused_while_it_loads' panicked at
crates/irlume-daemon/src/main.rs:5343:14:
a refusal within the deadline, not a wait for the engine:
Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }
```

Nothing is broken at the failure site. The test is reporting the wait for a lock as a missing answer from the daemon.

## Measurement

Daemon suite, twenty runs of each form, same container, four cores, `ORT_DYLIB_PATH` set so the engine-loading tests do not drop out:

| guard acquired | runs failed | median wall time |
| --- | --- | --- |
| inside the spawned thread (merged form) | 7 of 20 | 18.4s |
| on the test thread, before the spawn | 0 of 20 | 17.8s |

The seven failures were `a_request_needing_the_engine_is_refused_while_it_loads` five times and `a_status_request_answers_while_the_queue_is_wedged_and_workerless` twice. Those are the two spawn sites whose clients wait on an explicit timeout (10s and 5s) rather than on an unbounded read, which is why the other two never surfaced it and why this shows up as an unrelated lane going red rather than as a failing test.

Holding the guard across the socket work costs nothing measurable, and the earlier 17s-to-131s result does not apply here: that was an exclusive guard. Read guards are shared, so the other reading tests still overlap this one and only the writers wait.

## What changed

Four spawn sites now take `passwd_lock()` on the test thread ahead of `thread::spawn`. The comments that asserted the guard had to live on the server thread are replaced with the reason it does not.

`a_status_request_answers_while_the_queue_is_wedged_and_workerless` needed a little more than a move. It was taking one narrow guard for the lookup that builds its request line and another inside the closure, and one guard for the whole body replaces both. That is not tidying: `RwLock` reads do not nest, so a second `read()` on a thread that already holds one deadlocks as soon as a writer queues between them.

`every_test_that_resolves_a_user_holds_the_env_lock` gains the rule, next to the ones it already enforces. The brace matcher it uses runs over `structural()` output so the JSON request literals these tests write down the socket cannot throw the closure extent off.

## Checks

- Rule pin has teeth: moving a single guard back inside its closure fails the check, naming the test and its line. Restored afterwards.
- Full workspace suite with `--no-fail-fast`, every result line read: the only failures are the two that fail on unmodified main in this container as well, both root-only (`ir_emitter::tests::a_visible_config_after_a_failed_save_keeps_the_record`, `tui::tests::models_screen_unprivileged_shows_unknown_state_and_the_sudo_commands`).
- `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -D warnings` clean.

## What this does not prove

Twenty runs on one four-core container is enough to separate 35% from 0% and not enough to call the fixed form incapable of failing. The argument that it cannot is the mechanism: the wait that used to happen with the deadline running now happens before it starts. If the lane goes red here again it will be a different mechanism, not this one.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LpXmHLXnFxNsptNQs6s1yj)_

### DIFF
--- a/crates/irlume-daemon/src/main.rs
+++ b/crates/irlume-daemon/src/main.rs
@@ -34,6 +34,13 @@ use zeroize::Zeroize;
 /// Never taken inside `users::uid_for_name`/`name_for_uid` themselves. A test
 /// holding the write guard reaches those functions through production code,
 /// and a nested read acquisition would deadlock.
+///
+/// Nor inside a spawned `serve` thread, even though that is where the lookup
+/// runs. A waiting writer stops a new reader from entering, so a thread that
+/// asks for its own guard waits there while the client that spawned it is
+/// already counting down a socket deadline. The test thread takes the guard
+/// before the spawn and joins afterwards, which covers the same work without
+/// spending the deadline on it.
 #[cfg(test)]
 pub(crate) mod test_support {
     static ENV_LOCK: std::sync::RwLock<()> = std::sync::RwLock::new(());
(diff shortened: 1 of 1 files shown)


archledger:
Closing as superseded by #390, which is merged at `cd3e41e`.

This PR and #390 diagnosed the same failure independently and reached the same first answer: move the guard out of the spawned closure and into the parent. That diagnosis is correct as far as it goes, and this PR states it well.

It is not sufficient, and the reason is worth recording here rather than losing with the branch.

Holding the guard in the parent covers the child only while the parent is alive. Every one of these tests does something fallible before its join, and on that path Rust unwinds, DROPS the `JoinHandle`, which detaches the thread, and only then drops the guard. The detached server can still reach `pregate` and `getpwnam_r` afterwards with a writer free to run, which is the race #380 exists to prevent, reappearing on the failure path of the test that reports it. `read_line(..).expect(..)` is exactly that path, and it is the assertion that was failing.

#390 wraps the socket pair, the spawn and the join in a `std::thread::scope`, which joins its threads even while unwinding, so the caller's guard outlives the server on every exit. It also fixes a fifth site this PR does not touch, `an_unpublished_listing_reaches_the_worker_instead_of_erroring`, which discarded its handle outright and passed the rule test only because an unrelated `worker.join()` satisfied a check looking for any `.join()`.

Nothing here is wasted: the parent-guard move is the first half of what landed, and the ASan lane has been green since.

Issue #386, opened 2026-08-09
require_eyes_open can be satisfied by a saturated eye window rather than an open eye


archledger:
Surfaced while doing #222's peak semantics (PR #382) and deliberately left alone there, because the fix is a posture decision.

## The path

`eye_open_at` (`irlume-auth/src/lib.rs`) reads the same ceiling-unaware eye-window peak that #222 just made honest for the glint cue, and returns `peak >= EYE_OPEN_PEAK_MIN` (200.0). That feeds `Assessment::eyes_open`, which unlike the glint cue **is** enforced as a real deny:

```rust
if enr.require_eyes_open && !a.eyes_open { … }
```

A railed 255 satisfies `255 >= 200` unconditionally. So on a frame whose eye windows clip, the opt-in gate can be satisfied by saturation rather than by an open eye. It is a fail-open, on an opt-in gate whose whole purpose is to refuse a sleeping or unconscious user.

## Why PR #382 did not fix it

The obvious fix, treating a railed peak as "not open", is lockout-shaped. The measurement recorded on #222 has the peak pinned at 255 in **all 30** frames captured with glasses on, reading the lens specular rather than the cornea. Treating railed as not-open would deny effectively every glasses wearer who has `require_eyes_open` enabled.

So the two candidate postures both have a real cost:

- **Railed means not open**: closes the fail-open, denies glasses wearers with the gate on.
- **Railed means unreadable, and unreadable refuses**: same lockout, but honest about why, and it matches the direction #358 settled for the exposure gate (a gate that could not run is not a gate that passed).
- **Railed means unreadable, and the gate falls back to another cue**: no lockout, but needs a cue that survives clipping. EAR from the mesh is the obvious candidate and is already computed for the closure gesture.

The third is the only one that closes the hole without a lockout, and it is a design change rather than a threshold tweak.

## What is measured and what is not

Measured: the peak pins at 255 in all 30 glasses-on frames (#222). Measured: `EYE_OPEN_PEAK_MIN` is 200.0 and the comparison is `>=`.

Not measured: how often eye windows rail for a user WITHOUT glasses, which decides whether this is a glasses-only problem or a general one. That is a corpus question and the committed corpora do not carry a per-eye-window peak, only the glint peak, so it needs a capture campaign rather than a re-read of what exists.

Not established: whether anyone has `require_eyes_open` enabled in the field. It is opt-in per enrollment.

archledger:
[...]

PR #397, opened 2026-08-09
A railed eye window cannot report an open eye (#386)

archledger:
Fixes the fail-open half of #386. The issue stays open for the other half; see the last section.

## The defect

`both_eyes_open` took the window maximum around each eye landmark and compared it against a fixed 200, with no notion of the sensor's ceiling. A maximum is exactly the statistic clipping destroys: a railed window says the true value was at least the ceiling and never what it was, so nothing about the eyelid can be read out of it.

Its sibling `eye_glint_of` was given this treatment for #222, and its doc already states the reason:

> The peak reached `white`, the negotiated format's ceiling. A clipped sample tells you the true value was AT LEAST that, never what it was, and a maximum is exactly the statistic that destroys. This is the #222 reading: the repo's own measurements have the peak pinned at 255 in all 30 frames with glasses on, where it is reading the lens specular rather than the cornea.

`both_eyes_open` sampled the same statistic in the same eye window and was left alone.

Measured on hardware 2026-08-08, ASUS FHD IR module, one variable at a time:

| condition | gate | result |
|---|---|---|
| bare-eyed, eyes OPEN | on | denied 5/5 |
| glasses, eyes CLOSED | on | **granted 3/3** |

Eye-window peaks: 170 to 226 bare-eyed with 0 of 10 railed, 255 with glasses with 10 of 10 railed. The gate whose only purpose is refusing a sleeping or unconscious user was admitting exactly that case.

## The change

A railed window establishes nothing, and because this gate is deny-only and answers a bool, nothing collapses to `false`. The ceiling test runs **before** the threshold, since a railed peak clears 200 by construction and that is the path that granted.

The call also moves to the RAW frame and the negotiated ceiling, the pair `eye_glint_of` and `saturated_frac_of` already take at their call sites. Ambient subtraction moves a railed 255 to 254, so on a subtracted frame the refusal would silently never fire. The old call passed `&ir.data` with no ceiling, which is both halves of that mistake at once.

A format that names no ceiling (`Grey16`, `Nv12Luma`, `YuyvLuma`) passes the peak through unchanged, which is #237's settled precedent and the same choice `eye_glint_of` makes.

## Why the suite was green through all of this

`ir_frame_with_glints` models a closed eye as **no specular at all**. A real closed lid behind a spectacle lens still returns the emitter, railed. "Closed" in the old tests meant a condition that does not occur in the failing case, so nothing was ever red.

`a_railed_eye_window_cannot_report_an_open_eye` builds a railed window explicitly and was checked by mutation, not by reading it: deleting the ceiling arm fails it on "a window at the sensor ceiling establishes nothing about the eyelid", and restoring passes. Two further assertions stop a lazy fix passing anyway, since the refusal must come from the ceiling rather than from a lowered threshold, and a genuine sub-ceiling corneal glint must still read open.

## The review round found the half that test could not see

Codex returned DO NOT SHIP on `0c427a1` with one finding, and it was right. The fix has two independently necessary halves, and rejecting a railed peak buys nothing if the gate is handed the subtracted frame: subtraction moves every railed 255 to 254, which is under the ceiling and over `EYE_OPEN_PEAK_MIN`, so both eyes report open again. A test calling `both_eyes_open` with an already-railed buffer cannot observe which buffer production supplied, so reverting that line would have landed green.

`eyes_open_from_capture` makes the selection observable, and `the_eyes_open_gate_measures_the_raw_frame_not_the_subtracted_one` asserts it. Mutating the helper to ignore `saturation_frame` fails the new test on "the ceiling test must run against the preserved raw frame" while the ceiling test above still passes, which is the gap demonstrated rather than argued.

## The peak cannot be tuned into working, and the corpus says so

Recomputed here from `~/irlume-research/2026-08-07-blink-corpus`, whose 288 frames all verify against the committed `docs/pad-results/2026-08-07-blink-corpus.sha256`. Counting pixels at exactly 255 in each lit IR frame, glasses on:

| segment | pixels at 255, per frame |
|---|---|
| open-frontal-glasses | 370, 357, 330, 300, 292, 266, 259, 246, 228, 227, 227, 208 |
| held-closure-glasses | 371, 367, 322, 289, 281, 260, 249, 244, 244, 233, 224, 216 |

Open spans 208 to 370, closed spans 216 to 371. The two distributions are indistinguishable frame for frame, and their medians differ by about 8 pixels out of roughly 250. No threshold orders them.

That is the case for refusing rather than retuning. The lens sits in front of the eyelid, so the specular the window measures does not depend on whether the eye behind it is open. `EYE_OPEN_PEAK_MIN` is not mistuned; the statistic is measuring the wrong object, and the maximum over that window rails at 255 in both states.

The exposure gate is not a safety net here either. In the 8 records of `docs/pad-results/2026-08-04-occluder-gate.jsonl` where the eye peak railed, `ir_saturated_frac` reads 0.0042 to 0.0057 against an `IR_SATURATED_FRAC_MAX` of 0.10, and `ir_exposure_ok` is true in all 26 records. A point highlight is invisible to an area statistic.

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --locked`
- [x] `cargo test --workspace`
- [x] New test verified by mutation

Every command above is the exact line from `ci.yml` with its exit status captured.

- [x] **Hardware-validated 2026-08-10 on the Zenbook.** Glasses on, eyes fully closed: the pre-fix daemon granted 9 of 9, this branch denied 3 of 3, and the same branch with the gate switched OFF granted 3 of 3. That third run is what attributes the denial to this gate rather than to the `flir` PAD cue or to the sixteen commits merged into main since the pre-fix build. Full numbers and limits in a comment below.
- [ ] **The bare-eyed open case was NOT re-run**, so this PR carries no new evidence there. It is expected to keep denying, because peaks of 170 to 226 never reach the ceiling and still fail `EYE_OPEN_PEAK_MIN`, and that half of #386 is untouched here.

## This denies more, and only for opt-in profiles

`require_eyes_open` defaults to false, so the affected population is users who turned it on. For them, a glasses-on capture whose eye window rails will now deny and fall back to the password. That is the fail-safe direction for this gate, and it was previously granting those same users with their eyes shut.

**Measured, rather than described.** Replaying the committed blink corpus through the shipped `both_eyes_open`, twelve detected frames per cell:

| cell | open before | open after |
|---|---|---|
| glasses, eyes closed | 12 of 12 | 0 of 12 |
| glasses, eyes open | 12 of 12 | 0 of 12 |
| bare-eyed, eyes open | 1 of 12 | 1 of 12 |

So this does not make the gate stricter for a glasses wearer, it stops the gate working for them: after this change it refuses them whether their eyes are open or shut. Nothing bare-eyed is newly denied, and the bare-eyed row is the other half of #386 quantified, since eleven of twelve genuinely open frames already read as not-open before this PR touched anything.

Across all six cells the gate goes from admitting 37 of 72 genuine detected frames to admitting 1. Every frame it stops admitting is one where a lens, not an eyelid, was being measured. That is the argument for landing it, and equally the argument that nobody should enable `require_eyes_open` until the per-user `calibrate-closure` route exists.

## What stays open on #386

The bare-eyed half. A peak of 170 to 226 still fails a threshold of 200, so the gate still denies users it should admit, and this PR does not touch that.

I looked at replacing the statistic with the IR mesh EAR and did not, because the evidence does not support a global constant. The committed blink corpus separates the states cleanly on IR (computed from `docs/pad-results/2026-08-07-blendshapes-blink.csv`: worst closed min-EAR 0.1072, worst open 0.1625, glasses included), but `ClosureCalibration` records the same subject's median open EAR at 0.109 under an ambient of 22-42 and 0.166 under an ambient of 1, a 52% shift, and states plainly that no single calibration spans both sessions. A 0.109 open reading sits at the top of that corpus's closed band. A global EAR threshold would reproduce this defect in the other direction, for a different reason.

The certifiable route already exists and is per-user: `closure_calibration` on the enrollment, captured by `sudo irlume calibrate-closure`, which already accepts `--pose glasses-on-open`. Wiring the gate to it is a camera flow and an owner decision, not something to settle here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,39 @@ All notable changes to irlume are documented here. This project adheres to
 
 ### Security
 
+- **The require-eyes-open gate granted with the eyes closed, behind glasses.**
+  `both_eyes_open` took the maximum grey level in a window around each eye
+  landmark and compared it against a fixed 200, with no notion of the sensor's
+  ceiling. Its sibling `eye_glint_of` was made ceiling-aware for #222 and its
+  doc already records why: the measurements committed here pin the eye peak at
+  255 in all 30 frames with glasses on, where the window reads the lens specular
+  rather than the cornea. This gate sampled the same statistic and never got the
+  same treatment.
+
+  Demonstrated on hardware 2026-08-08 on the ASUS FHD IR module, one variable at
+  a time. Bare-eyed with the eyes OPEN it denied 5 of 5. With glasses on and the
+  eyes CLOSED it granted 3 of 3. Eye-window peaks read 170 to 226 bare-eyed, 0 of
+  10 railed, against 255 with glasses, 10 of 10 railed. The gate whose only
+  purpose is refusing a sleeping or unconscious user was admitting exactly that
+  case (#386).
+
+  A railed eye window now establishes nothing, and since this gate is deny-only
+  and answers a bool, nothing collapses to `false`. The call also moves to the
+  RAW frame and the negotiated ceiling, the pair its two siblings already take:
+  ambient subtraction moves a railed 255 to 254, so on a subtracted frame the
+  refusal would never fire. A format that names no ceiling passes the peak
+  through unchanged, which is #237's settled precedent.
+
+  This DENIES more than before, for profiles that opted in; the gate defaults
+  off. The other half of #386 is not fixed here. A peak of 170 to 226 still fails
+  a threshold of 200, so bare-eyed users are still denied, and no replacement
+  threshold can be certified from what the repository holds:
+  `ClosureCalibration` records the same subject's median open EAR at 0.109 under
+  an ambient of 22-42 and 0.166 under an ambient of 1, a 52% shift, and states
+  that no single calibration spans both sessions. A global constant on any of
+  these statistics reproduces this defect in the other direction. That half stays
+  open and wants the per-user `calibrate-closure` route.
+
 - **The IR exposure gate was off, not lenient, on every IR camera that is not
   8-bit grey.** `exposure_refusal` read "could not measure" as "clean", so on a
   node negotiating `Y16`/`Y10`/`Y12`/`NV12`/`YUYV` the #237 gate never ran and
@@ -954,7 +987,7 @@ Fixes a camera-firmware hazard present in 0.7.1. Please upgrade.
 
 - **Six checks reported success having checked nothing.** One theme in six places:
   a zero-match, zero-iteration or zero-sample case indistinguishable from a clean
-  pass — a conformance run that attempted nothing and exited 0, a set comparison
+  pass: a conformance run that attempted nothing and exited 0, a set comparison
   where no line carried the field, a `zip()` over an empty side. Each was proved
   by reproducing the failure first. Closes [#161], [#162], [#163], [#164], [#165]
   and [#166].
--- a/crates/irlume-auth/examples/landmark_failure_probe.rs
+++ b/crates/irlume-auth/examples/landmark_failure_probe.rs
@@ -182,7 +182,10 @@ fn main() {
             "| {name} | {} | {} | {} | {} | {} | {align} |",
             fmt(eye_glint(&ir, W, H, &lm)),
             fmt(eye_glint_contrast(&ir, W, H, &lm)),
-            both_eyes_open(&ir, W, H, &lm),
+            // `None`: these are synthetic pathology frames with no negotiated
+            // format behind them, so there is no ceiling to honour. The probe
+            // measures landmark failure, not clipping (#386).
+            both_eyes_open(&ir, W, H, &lm, None),
             fmt(pose.yaw_asym),
             fmt(pose.pitch_frac),
         );
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -1905,9 +1905,24 @@ impl Engine {
         // Eyes-open (IR corneal-glint heuristic), for the opt-in require-eyes-open
         // gate. Needs an IR face (the emitter lights the cornea); conservative:
         // false when it can't be verified.
+        //
+        // Same RAW-frame and ceiling pair its two siblings above take, and for
+        // the same reasons: subtraction moves a railed 255 to 254 so the
+        // clipping test must see `saturation_frame`, and a railed eye window
+        // reads the lens rather than the cornea (#386, #238 review). This call
+        // passed `&ir.data` with no ceiling until then, which is both halves of
+        // that mistake at once.
         let eyes_open = ir_top
             .as_ref()
-            .map(|f| both_eyes_open(&ir.data, ir.width, ir.height, &f.landmarks))
+            .map(|f| {
+                both_eyes_open(
+                    ir_stats.saturation_frame.as_deref().unwrap_or(&ir.data),
+                    ir.width,
+                    ir.height,
+                    &f.landmarks,
+                    ir_stats.white_level,
+                )
+            })
             .unwrap_or(false);
         Ok(Assessment {
             verdict,
@@ -3910,7 +3925,38 @@ const EYE_OPEN_PEAK_MIN: f32 = 200.0;
 /// eyelid does not. Conservative: requires the glint, so an unverifiable eye
 /// reads closed (auth falls back to password). Heuristic; used only when a
 /// profile opts into the require-eyes-open gate.
-pub fn both_eyes_open(grey: &[u8], w: u32, h: u32, lm: &irlume_vision::Landmarks5) -> bool {
+///
+/// `white` is the negotiated format's ceiling, and passing it is what stops
+/// this gate reading a lens instead of an eye. [`eye_glint_of`] next door
+/// already refuses a railed peak, and its doc records why: the repo's own
+/// measurements pin the peak at 255 in all 30 frames with glasses on, where it
+/// reads the lens specular rather than the cornea. This function sampled the
+/// same statistic and never got the same treatment, so on 2026-08-08 it
+/// GRANTED 3/3 with the eyes CLOSED behind glasses while denying 5/5 bare-eyed
+/// with them open (#386). A maximum is exactly the statistic clipping
+/// destroys: a railed window says the true value was at least the ceiling and
+/// never what it was, so no eyelid state can be read out of it.
+///
+/// Unlike `eye_glint_of`, which answers `None` for "not established", this gate
+/// is deny-only and returns a bool, so unreadable collapses to `false`. That is
+/// the fail-safe direction for a gate whose whole purpose is refusing a
+/// sleeping or unconscious user.
+///
+/// `white` of `None` means the format named no ceiling (`Grey16`, `Nv12Luma`,
+/// `YuyvLuma`) and the peak passes through unchanged, which is #237's settled
+/// precedent and the same choice `eye_glint_of` makes.
+///
+/// Pass the RAW frame. Ambient subtraction moves a railed 255 to 254, so a
+/// subtracted frame stops reading as railed and this refusal would not fire;
+/// the callers of `eye_glint_of` and `saturated_frac_of` already pass
+/// `saturation_frame` for that reason (#238 review) and this one now does too.
+pub fn both_eyes_open(
+    grey: &[u8],
+    w: u32,
+    h: u32,
+    lm: &irlume_vision::Landmarks5,
+    white: Option<u8>,
+) -> bool {
     // Same w*h invariant guard as eye_glint/mean_in_bbox: eye_open_at's in-bounds
     // test is against the logical w/h, so a truncated IR frame (buffer < w*h)
     // would index past the slice and panic the root daemon. A short frame reads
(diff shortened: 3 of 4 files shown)


archledger:
## The hardware A/B, written out so it can be run without rederiving it

This PR's one unticked box. Nothing below has been run; it is the procedure, and
every command in it was checked against the CLI and the daemon source rather
than recalled.

Current state on the Zenbook, read tonight: `irlume profiles list` reports
`require-eyes-open: off`, which is where the 2026-08-08 session left it. The
installed daemon is the packaged 0.9.0 and does not contain this change, so the
gate has to be swapped before any of it means anything.

### 1. Build the branch, merged up

The branch is behind main.

```
cd ~/irlume
git fetch origin
git checkout fix/eyes-open-railed-window-386
git merge origin/main --no-edit      # merge, not rebase: this repo squash-merges
cargo build --release -p irlume-daemon
```

### 2. Put that daemon under systemd

SELinux denies init_t exec of anything in `~/irlume/target`, which is labelled
`container_file_t`, and the symptom is a restart loop plus alert popups rather
than a clear error.

```
sudo install target/release/irlumed /usr/local/bin/irlumed-eyesopen-test
sudo restorecon -v /usr/local/bin/irlumed-eyesopen-test
sudo install -d /etc/systemd/system/irlumed.service.d
printf '[Service]\nExecStart=\nExecStart=/usr/local/bin/irlumed-eyesopen-test\n' \
  | sudo tee /etc/systemd/system/irlumed.service.d/eyesopen-test.conf
sudo systemctl daemon-reload && sudo systemctl restart irlumed
systemctl show irlumed -p ExecStart      # must name the test binary
```

The drop-in keeps the packaged unit's socket activation and watchdog, which
`scripts/install-host.sh` would replace with a plain unit.

### 3. Turn the gate on

`SetRequireEyesOpen` is `RootOrTarget`, so the owning account can set it without
sudo.

```
irlume profiles eyes-open on
irlume profiles list                     # must read require-eyes-open: on
```

### 4. The two conditions, one variable at a time

Three attempts each, and read the decision out of the event stream rather than
off the screen:

```
irlume auth test --events=jsonl
```

**Condition A, the one this PR is about.** Glasses on, eyes closed, same seat
and light as 2026-08-08. Every attempt must DENY. This granted 3 of 3 before
the change, which is the fail-open the PR closes.

**Condition B, the regression check.** Bare-eyed, eyes open. The expected result
is that it still DENIES, because the bare-eyed half of #386 is untouched here:
peaks of 170 to 226 against `EYE_OPEN_PEAK_MIN` of 200, both eyes required. A
grant in condition B would mean something other than this change moved, and it
needs explaining before anyone treats it as good news.

Space the attempts. Five consecutive face failures throttle face auth for 30
seconds, and the 2026-08-08 run was contaminated by counting throttled
non-attempts as denials. Count only attempts that reached a capture.

Optional, and it needs no daemon swap because the developer probes open the
camera directly: `IRLUME_DEV=1 sudo ./target/release/irlume liveness` reports the
eye-window peak, which should read 255 in condition A and 170 to 226 in
condition B. It does not surface `Assessment::eyes_open`, which is why the
decision above comes from `auth test`.

### 5. Put the machine back

```
irlume profiles eyes-open off
irlume profiles list                     # confirm it reads off again
sudo rm /etc/systemd/system/irlumed.service.d/eyesopen-test.conf
sudo systemctl daemon-reload && sudo systemctl restart irlumed
systemctl show irlumed -p ExecStart      # must name the packaged binary again
```

### What the result decides

Condition A denying 3 of 3 is what this PR claims and is the merge condition.
Condition B is not a merge condition either way; it is the number that says
whether the bare-eyed half of #386 has moved, and it belongs on that issue.

archledger:
## Hardware-validated 2026-08-10. The fail-open is closed, and the denial is attributable to this gate.

Zenbook, built-in ASUS FHD IR module, `rgb=/dev/video0 ir=/dev/video2`, secure tier. Same subject, same seat, same room light, all three runs inside fifteen minutes. Glasses on, eyes fully closed throughout every capture.

| daemon | `require_eyes_open` | result |
|---|---|---|
| built from `e59f935` (no fix) | ON | **granted 9 of 9** |
| built from this branch (`51c0857`) | ON | **denied 3 of 3**, `reason: not-live` |
| built from this branch (`51c0857`) | OFF | **granted 3 of 3** |

Rows one and two are the claim: the gate that exists to refuse a sleeping or unconscious user was admitting exactly that case, and now refuses it.

Row three is why the claim is attributable. `not-live` is a shared reason string and the daemon logs `third-party PAD cue 'flir' loaded (deny-only)` at startup, so rows one and two on their own would only establish that the newer binary denies. The newer binary carries this PR plus sixteen commits merged into main since `e59f935`. Toggling only the gate, on the identical binary, moves the verdict from deny to grant, which places the denial in the eyes-open gate rather than in the PAD cue or anywhere else in that range.

One thing was checked before the run rather than assumed: `both_eyes_open` is byte-identical between `e59f935` and today's `origin/main`, so between the two daemons that function differs by this PR alone.

Nine attempts in row one rather than three because the run was offered as three batches and every one granted. No throttle line appeared in any run, so no attempt in any row was a throttled non-attempt; that is the contamination the 2026-08-08 measurement recorded and this one avoided by keeping each batch under five.

### What was not measured

The bare-eyed open case. It is expected to keep denying, since peaks of 170 to 226 do not reach the ceiling and still fail `EYE_OPEN_PEAK_MIN` at 200, and that half of #386 is untouched by this PR. It was not run tonight, so this PR carries no new evidence about it either way.

Per-stage tracing was off, so no journal line names the gate. The attribution above rests on the toggle, not on a log.

One subject, one module, one room, one evening.

### Machine state

Restored. The systemd drop-in used for the swap is removed, the daemon is back on the binary it was running before, the test binary is deleted, and `require_eyes_open` is off again. The overlay that was already in place from 2026-08-07 is untouched.

archledger:
## Holding this PR. An adversarial pass found a deny population the body does not state, and two corrections to my own comment above.

Three agents were asked to argue against last night's A/B. Two of their central objections are answered by the gate-off row, which they were told had not been run. What survives is worth more than the A/B.

### The body understates who this denies

`docs/pad-results/2026-08-02-center-edge-corpus.jsonl`, 60 genuine frames from this same ASUS module, counting `ir_glint` at exactly 255:

| species | centre/edge range | railed |
|---|---|---|
| live_normal_glasses | 1.40-1.47 | **12 of 12** |
| live_close | 1.26-1.37 | **4 of 12** |
| live_normal_noglasses | 1.40-1.43 | 0 of 12 |
| live_far | 1.36-1.41 | 0 of 12 |
| live_offangle | 1.43-1.49 | 0 of 12 |

The blocks carry no species label in the `.jsonl`; they are matched to the table in the companion `.md` by their centre/edge ranges, which are disjoint and identify each block uniquely.

Under this change a railed eye window establishes nothing and the gate answers false, so with `require_eyes_open` on, every one of those 16 captures is denied. Twelve of them are the glasses population the body already names. **Four are `live_close`, and the corpus does not record whether that species wore glasses**; it labels eyewear only at the normal distance. So this may deny close-range captures that have nothing to do with lenses, and the body's account of the affected population would then be incomplete.

That is not a reason to abandon the change. A deny-only gate failing closed is the safe direction, and the fail-open it closes is demonstrated. It is a reason not to merge while the body says something narrower than the evidence.

### Two corrections to my own hardware comment above

**The confound is 57 commits, not sixteen.** `git log --oneline e59f935..524028f | wc -l` returns 57, and the two crates that decide this verdict changed by 2102 insertions against 207 deletions. I wrote sixteen from the PR numbers I remembered rather than counting.

**One rival I asserted was ruled out only by argument, and is now ruled out by measurement.** #340 flipped the unmeasured capture-mode default from concurrent to sequential (`None => (false, "default")` at `e59f935` against `None => (true, "default")` today), which would have changed the IR burst itself between the two arms. It did not fire here: `/etc/irlume/cameras.conf` holds `capture_mode.3277:0059:200901010001=sequential`, both `/dev/video0` and `/dev/video2` resolve to USB device `3-5` so `physical_device_id` matches and the legacy single-identity key is honored by the new pair-keyed lookup, and the new default is sequential in any case. Both arms captured sequentially.

### What still has no hardware evidence

The frame-selection half. `IRLUME_IR_AMBIENT_SUBTRACT` is opt-in and was not set, so `saturation_frame` was `None` and the gate read the raw frame either way. That is the half the review round returned DO NOT SHIP over, and last night exercised only the ceiling half.

### Cheapest way to settle the deny population

Replay the committed frames through the shipped gate offline. No camera, no daemon: the raw GREY8 PGMs are on disk from this module, and `liveness_replay` already walks stored IR frames emitting every shipped cue per frame. That answers what fraction of each genuine species this gate now refuses, which is the number the body should carry.

archledger:
## Replayed offline against the committed blink corpus. The hold is lifted, and the body needs two numbers it does not have.

No camera and no daemon. `both_eyes_open` called twice on the same pixels and the same landmarks, `Some(255)` being this PR and `None` being byte-for-byte the old behaviour, over `~/irlume-research/2026-08-07-blink-corpus`. All 288 files verified against `docs/pad-results/2026-08-07-blink-corpus.sha256` first. Detection is the shipped path: `grey_to_rgb`, `Detector::detect`, highest-scoring finite detection, exactly as `liveness_replay` does it.

Twelve of the 24 frames in each cell are the emitter-off half of the burst and detect no face, so the denominator is 12.

| cell | detected | open, OLD | open, NEW | newly denied |
|---|---|---|---|---|
| held-closure-glasses | 12 | 12 | 0 | 12 |
| natural-blink-glasses | 12 | 12 | 0 | 12 |
| open-frontal-glasses | 12 | 12 | 0 | 12 |
| held-closure | 12 | 0 | 0 | 0 |
| natural-blink | 12 | 0 | 0 | 0 |
| open-frontal | 12 | 1 | 1 | 0 |

### What it settles

**The fail-open is closed on stored evidence too.** Glasses with eyes held closed: 12 of 12 read open before, 0 of 12 after. That is the same result as last night's hardware run, arrived at without a camera, so the two are independent.

**Nothing bare-eyed is newly denied.** Zero in all three bare cells. The concern I raised from the centre/edge corpus, that close-range captures without lenses might be swept in, does not reproduce here. The newly-denied population is the glasses population the body already names.

### The two numbers the body is missing

**Glasses with eyes genuinely OPEN also goes 12 of 12 to 0 of 12.** The gate does not become stricter for a glasses wearer, it stops functioning for them: after this change it refuses them whether their eyes are open or shut. The body says a railed capture "will now deny and fall back to the password", which is true and understates it.

**The bare-eyed half was already broken, and this measures it: 1 of 12.** Eleven of twelve genuinely open bare-eyed frames were read as not-open by the OLD code, and the new code agrees, so `newly_denied` is zero. That is #386's other half quantified on this module, and it is the strongest argument that `require_eyes_open` should not be enabled by anyone until the per-user `calibrate-closure` route exists.

Taken together: with the gate on, this change moves the fraction of genuine detected frames admitted from 37 of 72 to 1 of 72. Every one of the 36 it removes is a frame where a lens, not an eyelid, was being measured.

### Recommendation

Merge, with the body carrying these numbers. A deny-only gate that defaults off, failing closed on a statistic it cannot read, is the correct direction, and leaving a demonstrated fail-open in place to protect a gate that already admits 1 genuine frame in 12 is the worse trade. What must not happen is this landing while the body implies the cost is narrower than it is.

The probe was a throwaway and is deleted. To reproduce: call the shipped `both_eyes_open` with `Some(255)` and `None` over the corpus, which is fifty lines around the `liveness_replay` scaffolding.

Issue #406, opened 2026-08-09
The strict-model acceptance test reports success when its fixture is absent


archledger:
`strict_verify_refuses_a_tampered_model_and_accepts_a_shipped_one` runs its tampered-model half, then checks for `models/blaze_face_short_range.onnx`. If the fixture is absent, it prints a skip line and returns from the test successfully.

That means a build where strict mode rejects every release model can still report this test as passed. The failure appears in source-package or worktree environments that do not carry the fetched model files.

The fix needs to preserve those environments rather than replacing the return with an unconditional path assertion. Either supply the pinned fixture wherever this test runs, or move the acceptance half into a guarded integration lane whose harness fails when its required fixture is absent. The tampered-model refusal can remain a fixture-free unit test.

Found by the single Codex review round on #405. The issue is follow-up work because the defect predates that message-only PR.

PR #410, opened 2026-08-10
Pin the recognizer's embedding so a runtime bump cannot move it quietly (#407)

### LINKED ISSUE #407: No CI gate pins the recognizer's embeddings, so every runtime bump is held by hand
Every bump of `ort` or of the system ONNX Runtime raises the same question:
did the embedding move? If it did by any amount, stored templates stop matching
and face login breaks for everyone. Nothing in CI answers it, so the question
gets answered by hand or not at all.

#400 sat held across two sessions on exactly this. The proposed check was one
authentication on the Zenbook against the existing enrollment, which is weaker
than it looks: it shows the embedding did not move *enough to fail a single
match at the configured threshold*, not that it did not move.

The measurement that settles it took one command and no hardware, because the
embedding is a pure function of the aligned chip bytes. Written up on #400.

## What to build

A test that embeds a fixed input and compares the result against a vector
committed beside it.

The input should be a synthetic 112x112x3 chip from a seeded LCG written into
the test source, not a corpus frame and not a camera capture. That keeps the
fixture at zero bytes on disk, needs no research corpus, and makes the input
reproducible from the source alone.

The recognizer itself is already fetched by `scripts/fetch-models.sh` in the
`check` and `test (stable)` jobs, so the model side needs no new machinery.

## Compare by cosine, not by sha256

The measurement on #400 came back bit-identical, and it is tempting to pin the
sha256 of the output bytes. Do not.

`ci.yml` says the `ci` label can run on either machine, so the same job runs on
a hosted ubuntu-24.04 runner and on the self-hosted Arch runner. Those are
different CPUs, and ONNX Runtime does not guarantee the last bits of a result
agree across instruction sets. A sha256 gate would go red on a runner swap and
teach everyone to ignore it.

Cosine against the stored vector, with the floor high enough that a real shift
fails and CPU noise does not, is the gate that means something. Pick the floor
from a measurement across both runners rather than from a guess.

## The fixture trap this must not repeat

The recognizer is gitignored and fetched. A test that reads it must not follow
the shape in #406, where an absent fixture makes the test return successfully
and report a pass. Whatever guard this test uses for a missing model has to make
the job RED, or the gate reports green in exactly the environments where it did
not run.

## Not in scope

Detector, mesh, Blaze and the IR adapter outputs. The detector matters to
matching because its landmarks move the aligned chip, and it was measured on
#400, but pinning it is a separate decision with its own tolerance question.

archledger:
Closes #407.

Every bump of `ort` or of the system libonnxruntime asks the same question: did
the embedding move? If it moves, stored templates stop matching and face login
breaks for everyone. Nothing in CI answered it, so #400 sat held across two
sessions while the question was answered by hand.

`the_recognizer_embedding_of_a_fixed_input_has_not_moved` embeds `chip(407)`,
the module's existing deterministic chip, through `Embedder::embed` and compares
the result against 512 values committed beside it. The input and the reference
are both source, so there is no fixture that can be missing.

This is not a model-identity check. Substituted weights are already caught by
digest, and the recognizer's sha256 is the `embed:<sha256>` space tag that
filters stored scans. What had no check is the same weights producing different
numbers because the code around them changed.

The floor was measured, not chosen. Perturbing the pipeline in the ways two
machines differ moved the embedding by at most 1.5e-12 in `1 - cosine`: graph
optimization Level3 against Level1, Level2 and off, with intra-op thread counts
1, 2, 4 and 8 bit-identical to each other. Embedding a different input scored
2.7e-1. The limit of 1e-6 sits six orders above the largest perturbation that
could be produced here and five orders below a change that would mean anything.
In per-element terms it trips when the average element moves more than about
6e-5, which is six times the 1e-5 element-wise band ONNX Runtime maintainers
state as expected variation.

It compares by cosine rather than by sha256 of the output bytes, and the issue's
stated reason for that is wrong. The issue says the `ci` label can resolve to
either a hosted or a self-hosted runner. It cannot: `runs-on: [self-hosted, ci]`
requires every label to match, and all seven checks required for a merge run on
hosted ubuntu-24.04. The conclusion survives for better reasons. ONNX Runtime
publishes no reproducibility guarantee even on identical hardware, MLAS selects
a different GEMM kernel per CPUID, Level3 enables layout optimization whose
NCHWc block size depends on AVX-512, and `load-dynamic` means the libonnxruntime
under this crate is whatever the host installed.

The gate was mutated four ways, each by running it:

- Every element perturbed by 1e-4, predicted drift 2.6e-6: fails at 2.5599e-6,
  which is the prediction the floor was derived from.
- Every element perturbed by 3.2e-5, predicted drift 2.6e-7: passes.
- Reference zeroed so the cosine is NaN: fails. The assertion is spelled
  `drift <= limit` rather than `!(drift > limit)` for exactly this, since NaN
  compares false to everything and the negated form would pass on it.
- `models/glintr100.onnx` moved out of the tree: fails rather than skipping,
  which is the shape #406 was filed for and this test must not repeat.

It sits in `model_tests`, where the other model-backed tests live, so it runs in
`check`, `test (stable)`, `coverage`, asan and the hardware suite with no
workflow edit and shares the already-loaded 260MB session.

## What the review corrected

The first revision claimed asan and the hardware suite gave this self-hosted
coverage. They do not: `asan.yml` runs on hosted ubuntu-24.04,
`hardware-checks.yml` selects two named test sets rather than the workspace, and
`hardware-suite.yml` is schedule triggered on the default branch. No self-hosted
machine runs this on a PR.

It also said `verify_models` refuses an unmanifested file. That holds only under
`IRLUME_MODELS_STRICT=1`; without it the mismatch warns and the daemon carries
on. The comment now claims the property that holds unconditionally, which is the
digest becoming the `embed:<sha256>` space tag.

The review's merge-blocking finding is real and is filed as #411: this gate runs
against the runtime `ci.yml:118` downloads, while Nix, Fedora, Debian and the
PPA each pin their own and Arch takes the rolling system package, with nothing
checking that the seven agree. Unifying them touches five packaging lanes and
carries release risk this test does not, so the test's own documentation now
states the limit instead of implying coverage it lacks.

Clippy 1.88 is the version the required check runs, and it fires
`uninlined_format_args` where 1.96 does not. Local clippy on stable was not the
gate; the fix is verified on both toolchains.

Not covered: no cross-machine number exists yet. Every measurement above is from
one CPU, an Intel Core Ultra 7 258V with no AVX-512, against libonnxruntime
1.24.4. The first CI run is the first cross-machine evidence, and if it fails
the limit is not what to change until the cause is known. The detector, mesh,
Blaze and IR adapter outputs stay unpinned, which #407 puts out of scope.

### DIFF
--- a/crates/irlume-vision/src/lib.rs
+++ b/crates/irlume-vision/src/lib.rs
@@ -2083,12 +2083,23 @@ mod model_tests {
     /// matching and face login breaks for everyone. #400 sat held across two
     /// sessions on exactly that question with nothing in CI able to answer it.
     ///
-    /// This is NOT a model-identity check. Substituted weights are already
-    /// caught by digest: `verify_models` refuses an unmanifested file, and the
-    /// recognizer's sha256 is the `embed:<sha256>` space tag that
-    /// `recognizer_space_matches` filters stored scans by. What has no check is
-    /// the same weights producing different numbers because the code around
-    /// them changed.
+    /// This is NOT a model-identity check. `verify_models` hashes the
+    /// recognizer at startup: under `IRLUME_MODELS_STRICT=1` an unmanifested
+    /// digest refuses the start, and without it the mismatch warns and carries
+    /// on. Either way that digest becomes the `embed:<sha256>` space tag
+    /// `recognizer_space_matches` filters stored scans by, so substituted
+    /// weights cannot be matched against templates from the old ones. What has
+    /// no check is the same weights producing different numbers because the
+    /// code around them changed.
+    ///
+    /// What it does NOT cover is the runtime it runs against. The required
+    /// lanes download the version hardcoded at `ci.yml:118`, while Nix, Fedora,
+    /// Debian and the PPA each pin their own copy and Arch takes the rolling
+    /// system package. Nothing makes those agree (#411), so a packaging lane
+    /// moving to a newer runtime does not turn this test red. It also runs on
+    /// hosted runners only: `hardware-checks.yml` selects two named test sets
+    /// rather than the workspace, and the nightly hardware suite runs on the
+    /// default branch, so no self-hosted machine executes this on a PR.
     ///
     /// The input is generated from source, so there is no fixture to be
     /// missing. The reference below was produced through this exact call,
@@ -2287,12 +2298,12 @@ mod model_tests {
         // default.
         assert!(
             drift <= PINNED_EMBEDDING_MAX_COSINE_DRIFT,
-            "the recognizer's embedding moved: 1-cos = {drift:e} against a limit of {:e}. \
+            "the recognizer's embedding moved: 1-cos = {drift:e} against a limit of \
+             {PINNED_EMBEDDING_MAX_COSINE_DRIFT:e}. \
              Stored templates are matched against embeddings from this model, so a real \
              shift breaks face login for every enrolled user. Do not widen this limit to \
              make the build green; find what changed (ort version, libonnxruntime version, \
-             session options) and decide whether enrollments must be rebuilt.",
-            PINNED_EMBEDDING_MAX_COSINE_DRIFT
+             session options) and decide whether enrollments must be rebuilt."
         );
     }
 }
--- a/crates/irlume-vision/src/lib.rs
+++ b/crates/irlume-vision/src/lib.rs
@@ -2074,6 +2074,227 @@ mod model_tests {
         };
         assert!(b.detect_top(&view).expect("blaze flat").is_none());
     }
+
+    /// The recognizer's output for one fixed input, pinned so a runtime bump
+    /// cannot move it silently (#407).
+    ///
+    /// Every bump of `ort` or of the system libonnxruntime raises the same
+    /// question: did the embedding move? If it moves, stored templates stop
+    /// matching and face login breaks for everyone. #400 sat held across two
+    /// sessions on exactly that question with nothing in CI able to answer it.
+    ///
+    /// This is NOT a model-identity check. Substituted weights are already
+    /// caught by digest: `verify_models` refuses an unmanifested file, and the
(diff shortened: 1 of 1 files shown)


archledger:
Review round applied in `1e73c39`. All three findings were checked in the tree before acting on them, and all three were right.

The merge-blocking one is filed as #411 rather than fixed here. The gate runs against the runtime `ci.yml:118` downloads, and Nix, Fedora, Debian and the PPA each pin their own copy while Arch takes the rolling system package. Seven pins, no check that they agree. Unifying them touches five packaging lanes and carries release risk this test does not, so the test now documents the limit instead of implying coverage it lacks. `scripts/check-packaging-parity.sh` already exists for this class and is the natural home for the fix.

The `verify_models` wording was wrong in the way the finding said: refusal is the `IRLUME_MODELS_STRICT=1` branch, and the default warns and carries on at `main.rs:148`.

One more, found by CI rather than the review: the required check runs clippy on Rust 1.88, which fires `uninlined_format_args` where 1.96 does not. Verified on both toolchains after the fix.

archledger:
First cross-machine evidence, and it is a pass rather than a number. On this head the three jobs that fetch the weights and run the workspace suite are green: `test (stable)`, `AddressSanitizer (test suite)` and `coverage (line-coverage ratchet)`. Each is a separate hosted `ubuntu-24.04` instance, none of them the Intel Core Ultra 7 258V the reference was generated on, and the gate cleared 1e-6 on all three.

What that does not establish: the actual drift. A passing test captures its output, so the observed `1 - cos` is not in the log, and every one of those runners downloads the same onnxruntime 1.24.4 that #411 is about. An AVX-512 machine is still unmeasured, and the nightly hardware suite on archhost is the first thing that will run this against a system-installed runtime.

PR #412, opened 2026-08-10
The routing comment on the required checks was left behind by the rebuild

archledger:
`ci.yml` told readers that five of the seven required checks could only run on
`arch` and that two could run on either machine. Both halves are false. All
seven run on `ubuntu-24.04`, and `runs-on: [self-hosted, ci]` needs a runner
carrying every listed label, so `ci` never resolves to a hosted runner in the
other direction either.

The comment predates the 2026-08-07 capability routing. It was not harmless:
#407 was written on its authority and its central premise, that the same job
runs on a hosted runner and on the self-hosted Arch runner, came from these two
lines. That premise decided the issue's recommendation, and correcting it took a
reviewer plus a reading of the workflow files.

The comment also implied one place to change. There are two: classic branch
protection lists all seven contexts, the active ruleset lists five of them, and
what is enforced is the union. Read from the API rather than from the previous
comment: `repos/archledger/irlume/branches/main/protection` returns the seven,
`repos/archledger/irlume/rulesets/19429289` returns five. Editing one alone does
not drop a check, which is the trap worth naming.

Comment only. No job, trigger, runner label or step changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,9 +55,20 @@ concurrency:
 #   "systemd units (verify + exposure ratchet)"
 #   "nix flake check"
 # Renaming a job below without updating the protection rule leaves a PR waiting
-# on a check that never reports, so it can never merge. Change both together.
-# Five of the seven can only run on `arch` (archhost); the two on the shared
-# `ci` label can run on either machine.
+# on a check that never reports, so it can never merge. Change both together,
+# and note that "the protection rule" is TWO settings: classic branch protection
+# lists all seven, the active ruleset (id 19429289) lists five of them, and the
+# effective requirement is the union. Editing one alone does not drop a check.
+#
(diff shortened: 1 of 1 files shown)


archledger:
Applied in `b9e4b23`. The finding is right, and this PR had introduced a smaller version of the error it exists to remove.

Checked against this repository rather than the documentation alone before changing anything. A completed `pull_request` run on #413 logs:

> HEAD is now at 593ba54 Merge 3e20d8c into e971020

and `refs/pull/413/merge` resolves to a commit whose two parents are `e971020` (main) and `d5d9624` (the branch head). The runner evaluates a synthetic merge commit that exists in neither branch, so neither "the PR head" nor "the base branch" was accurate.

The same wording also sits in the file's opening comment and predates this PR. Correcting only the new copy would have left the file self-contradicting, so both are fixed, and the comment now names `pull_request_target` as the trigger that does read the base branch.

PR #413, opened 2026-08-10
Make the eight ONNX Runtime pins agree, or fail the build (#411)

### LINKED ISSUE #411: The ONNX Runtime version is pinned in eight places with no check that they agree
The ONNX Runtime version is hardcoded in seven places and nothing checks that they agree.

```
.github/workflows/ci.yml:118, :346, :422   ver=1.24.4
.github/workflows/asan.yml:134             ver=1.24.4
.github/workflows/install-matrix.yml:126   ver=1.24.4
flake.nix:50                               ortVersion = "1.24.4"
packaging/fedora/irlume.spec:2             %global ort_ver 1.24.4
packaging/debian/build-deb.sh:8            ORT_VER="${ORT_VER:-1.24.4}"
scripts/build-ppa-source.sh:28             ORT_VER="${ORT_VER:-1.24.4}"
```

Arch takes the rolling system package and pins nothing.

They agree today. That is the only thing keeping the divergence invisible.

## Why it matters now

#410 adds a test that pins the recognizer's embedding, so a runtime that moves the numbers turns a build red before stored templates stop matching. That test runs against the runtime `ci.yml` downloads. Bump `%global ort_ver` in the Fedora spec and every required check keeps testing 1.24.4, keeps passing, and Fedora users get the new runtime with no measurement behind it.

The same reasoning covers the reverse: bump `ci.yml` alone and the gate starts guarding a runtime nobody ships.

## Where the fix belongs

`scripts/check-packaging-parity.sh` already exists for this exact class. Its own header says every lane "must agree on the version", and it was written because the same bug shipped three times. It compares units, installed programs and the irlume version. It does not compare the ONNX Runtime version.

Two shapes, and the second is better:

1. A parity check that reads all seven and fails when they differ. Cheap, no build changes, catches divergence at the point it is introduced.
2. One canonical `packaging/onnxruntime.version` that every consumer reads, so divergence cannot be expressed. More edits, and the Fedora spec cannot read a file at macro-expansion time without a generation step, so that lane still needs a parity assertion.

Option 1 first is enough to close the hole. Option 2 is the tidier end state.

## Also worth deciding, and not the same question

The Arch lane in `install-matrix.yml` downloads 1.24.4 rather than installing the `onnxruntime` package its own PKGBUILD depends on, so the rolling version Arch users actually get is never exercised. That is a coverage gap in the opposite direction from the pins above, and it is the one lane where a version check has nothing to compare against.

## Found by

The Codex review round on #410, which called it merge-blocking for that PR. It is filed separately instead: #410's gate is worth having against CI's runtime, and unifying seven pins across five packaging lanes is a change with its own release risk. #410 now states this limit in the test's own documentation rather than implying coverage it does not have.

archledger:
Closes #411.

irlume builds `ort` with `load-dynamic`, so libonnxruntime.so is whatever the
host supplies and the version a lane bundles is the version its users run. Eight
files name it and nothing made them agree:

    .github/workflows/ci.yml (three times)
    .github/workflows/asan.yml
    .github/workflows/install-matrix.yml
    flake.nix
    nix/module.nix
    packaging/fedora/irlume.spec
    packaging/debian/build-deb.sh
    scripts/build-ppa-source.sh

They agree today, which is the only thing that kept the divergence invisible.
Bump the Fedora spec alone and every required check keeps testing the old
runtime, keeps passing, and Fedora users get a runtime nothing executed. The
embedding gate that landed in #410 would stay green while the numbers it pins
moved under the shipped package, which is the failure it exists to prevent.

`check-packaging-parity.sh` already exists for this class. Its header says every
lane "must agree on the version" and it was written because the same bug shipped
three times. It compared units, helper programs, models and the irlume version,
and not this.

The check reads each pin, refuses to proceed when a listed file has no
extraction pattern, and fails when any file yields zero or more than one
version. That last part matters: a pattern that stops matching yields an empty
string, two empty strings compare equal, and the result is a guard that reports
ok while checking nothing. That is the failure mode this script's own dest_for()
comment was written about.

`docs/DEVELOPMENT.md` is checked too, because it hands out copy-paste setup
commands carrying the version and a contributor who follows it after a bump
installs a runtime no lane ships. The dated survey under docs/cross-distro/ is
deliberately excluded: it records what was true on its date and is not
instructions.

Proved by mutation, each one run:

- flake.nix moved to 1.24.5: fails, naming both files and both versions.
- The spec's `%global ort_ver` renamed so the pattern misses: fails on the count
  rather than passing with an empty string.
- DEVELOPMENT.md left on 1.24.3: fails.
- `packaging/arch/PKGBUILD` added to the list with no pattern for it: hard error,
  exit before any comparison.
- Unmutated: passes, and shellcheck is clean.

Not covered: the tarball sha256 digests, which several lanes pin beside the
version and which have to change with it. Arch is not in the list because it
depends on the rolling system `onnxruntime` package and pins nothing, so there
is no version there to compare; that its CI lane downloads 1.24.4 instead of
installing that dependency is a separate gap, recorded on #411.

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -78,14 +78,20 @@ let
       };
     };
 
-  # The pinned onnxruntime 1.24.4 build. NixOS stable ships 1.22.2, which is
-  # below irlume's 1.24 API floor and deadlocks the `ort` loader at startup, so
-  # bundle the exact upstream release the RPM and .deb carry.
+  # The pinned onnxruntime build. NixOS stable ships 1.22.2, which is below
+  # irlume's 1.24 API floor and deadlocks the `ort` loader at startup, so bundle
+  # the exact upstream release the RPM and .deb carry.
+  #
+  # One binding, interpolated into the label and the URL, the way flake.nix
+  # already does it. Written out three times, a bump could change the derivation
+  # label while still fetching the old archive, and the parity check in
+  # scripts/check-packaging-parity.sh reads the label (#411).
+  ortVersion = "1.24.4";
   onnxruntime-bin = pkgs.stdenv.mkDerivation {
     pname = "onnxruntime-linux-x64";
-    version = "1.24.4";
+    version = ortVersion;
     src = pkgs.fetchurl {
-      url = "https://github.com/microsoft/onnxruntime/releases/download/v1.24.4/onnxruntime-linux-x64-1.24.4.tgz";
+      url = "https://github.com/microsoft/onnxruntime/releases/download/v${ortVersion}/onnxruntime-linux-x64-${ortVersion}.tgz";
       hash = "sha256-OiEfvqJSweZikGWPG3NbdyBWFJ8oMh5xwwiULNtUt0c=";
     };
     nativeBuildInputs = [ pkgs.autoPatchelfHook ];
--- a/scripts/check-packaging-parity.sh
+++ b/scripts/check-packaging-parity.sh
@@ -177,6 +177,86 @@ for derived in packaging/ppa/build-ppa-container.sh nix/package.nix; do
 done
 
 echo
+echo "== every lane pins the same ONNX Runtime =="
+# irlume builds `ort` with `load-dynamic`, so libonnxruntime.so is whatever the
+# host supplies at runtime and the version each lane bundles IS the version its
+# users run. Eight files name it and nothing made them agree (#411): a lane
+# bumped on its own ships a runtime no CI job ever executed, and the embedding
+# gate in irlume-vision (#407) would stay green while the numbers it pins moved
+# under the shipped package.
+#
+# Each lane spells the pin its own way, so extraction is per file and an
+# unlisted file is a hard error rather than a skip. Same reason as dest_for()
+# above: a pattern that silently matches nothing yields an empty string, and two
+# empty strings compare equal, which is a guard reporting ok while checking
+# nothing.
+ort_pin_of() {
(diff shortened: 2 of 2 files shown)


archledger:
All three findings applied in `5b3dda3`, each reproduced before it was fixed.

**The Nix bypass was the important one.** `nix/module.nix` wrote the version three times, once as the derivation label and twice inside the fetch URL, and the check read the label. A bump could move every extracted pin while Nix kept fetching the old archive, and with the hash untouched that old artifact still satisfies the fixed-output check. The file now holds one `ortVersion` binding interpolated into both, which is what `flake.nix` already did, so the duplication is gone rather than checked indirectly. Moving the binding alone now leaves zero occurrences of the old version in the file, and `nix-instantiate --parse` accepts it.

**Counting distinct values is not counting pins.** Deleting one of ci.yml's three fetch steps left the other two agreeing and the guard reported ok. Each file now declares its expected occurrence count, a file with no declared count is a hard error, and uniqueness is tested after the count. The mutation now reads `expected 3 pin(s), found 2`.

**The doc predicate matched one filename anywhere.** All five places the recipe names the version are checked, so a document that downloads one version and puts a different one on `ORT_DYLIB_PATH` fails. Proved by leaving the tarball name correct and staling only the export path.

That last one is worth naming: this is a guard I wrote in the same session as #410, and all three findings are the same defect, which is reading one spelling of a value while several live spellings exist.

Issue #389, opened 2026-08-09
Enrolment cannot complete on a concurrent-stored pair whose camera starves RGB, and the error blames the lighting


archledger:
Measured on hardware 2026-08-08 while validating #379. Filing separately because #381 is merged and the tradeoff wants a decision rather than a comment on a closed PR.

## Symptom

On a camera pair stored as `concurrent`, whose camera starves its RGB interface when both sensors stream, `irlume enroll` fails:

```
enroll failed: protocol: no live scan captured; check lighting and framing
```

The message names lighting and framing. Both were fine; the face was plainly visible on the RGB node checked independently in a browser.

## Mechanism

NexiGo HelloCam N930W (`3443:c803`), pair forced to `concurrent`, four consecutive enrolment scans, identical:

```
assess: rgb 640x480 in 173ms, faces=0 top-det=0.00
assess: ir  640x360 in 472ms, faces=1 top-det=0.94
liveness(cross-spectrum): Uncertain (no face in RGB; present your face)
```

`camera-tune` on the same camera, same session: concurrent capture keeps **50% of RGB** and 98% of IR brightness.

RGB-lost/IR-kept is the shape the cross-spectrum self-heal rescues by recapturing RGB alone. #381 stops that recapture when the caller holds the streams, and enrolment holds both open for its whole capture loop, so the rescue no longer happens and every scan in the loop fails the same way.

#381 is right to forbid it: that recapture is a standalone reopen of a node this same process is already streaming, which is #187, and a module answering EBUSY kills the enrolment anyway. The gap is in its rationale, which says skipping the recovery costs little because "the next scan gets a fresh pair of frames anyway". That holds for transient contention and not for structural: at 50% retention every scan fails identically and the loop runs out.

## Reachability

Only when the pair is stored `concurrent`. The shipped default after #340 is sequential, and `camera-tune` stores sequential on this very camera because it measures the dimming, so a fresh install does not meet this.

What meets it is a verdict that was measured concurrent once and stopped being true: a different USB port, a hub, a busier bus, a second camera added. The stored verdict is sticky and nothing re-measures it.

## Options, none free

1. **Let the enrolment loop fall back to sequential capture** when it sees the RGB-lost/IR-kept shape. Fixes the population that cannot enrol, but changes capture behaviour on a path that was not measured for it, and sequential capture widens the interval the cross-spectrum co-location cue binds.
[...]

PR #414, opened 2026-08-10
Offer the second reading when enrolment captures nothing (#389)

archledger:
Addresses #389 option 3. The issue stays OPEN for options 1 and 4, which change capture behaviour and re-measure a stale verdict; this changes neither.

On a pair stored `concurrent` whose camera starves RGB while both sensors
stream, every enrolment attempt returns an IR face and no RGB face, and the
error read "no live scan captured; check lighting and framing". The room was
lit and the face was plainly visible on the RGB node checked independently.

The message now keeps that advice and adds a second reading where the evidence
allows one: the infrared sensor found a face on every attempt and the colour
sensor found none, so if the room was lit this camera may be dimming its colour
stream while both sensors run, and `camera-tune` should be re-run in a lit room.

Three conditions, each of which the issue's own analysis established:

The held path only. On the fallback path `self_heal_may_recapture` returns true,
RGB is recaptured standalone and `rgb_top` is reassigned, so an IR-only attempt
there means RGB found no face WITH THE SENSOR TO ITSELF. Naming starvation there
would assert a cause the code disproved one branch earlier.

Every attempt, not any. One attempt in ten is a user who turned away and the
loop recovers; ten in ten is the structural case, which is why it exhausts the
budget instead of retrying out of it.

The lighting clause is never replaced, only appended to. An unlit room produces
the identical shape, and this repository records it observed live at rgb faces=0
and ir faces=1 with a 0.92 detection (`uncertain_short_circuits`). Nothing at
this point in the code separates the two causes.

The remedy says "in a lit room" deliberately. `camera-tune` stores its verdict
unconditionally under `ProbeStore::Always` without consulting `conclusive()`,
and `conclusive()`'s own doc records retention reading 121%, 122% and 126% at an
RGB mean of 17, which is arithmetic on noise. Advising a re-measure without that
qualifier would talk a user in the dark into persisting the Concurrent verdict
that option 4 of the issue exists to fix.

All three enrolment failure sites share one advice tail, because a message that
learns something while its two siblings do not is how one of them keeps blaming
the room.

Proved by mutation, each one run:

- Trigger without `held_sessions`: fails.
- Trigger on any ir-only attempt rather than every one: fails.
- Hint replacing the lighting clause instead of appending: fails.
- Remedy without "in a lit room": fails.
- Call-site arguments swapped: this one SURVIVED as two bools, so
  `observe_attempt` now takes the two embeddings, whose types differ. The swap
  is a compile error, which is the stronger kill.

No test asserts that any message omits the word lighting. Such an assertion
would pin the regression in place by making the removal of correct dark-room
advice a requirement to stay green.

Not covered: no camera. The predicate and the tally are unit-tested, the
plumbing that fills the tally inside `capture_scan_loop` is not, because that
loop needs a device. What the issue reproduced on hardware was the old message,
and confirming the new one needs the same NexiGo pair forced to `concurrent`.
This changes no capture behaviour: options 1 and 4 of the issue, which would,
are untouched and the issue stays open for them.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -3084,7 +3084,7 @@ impl Engine {
         &mut self,
         want: usize,
         pitch_neutral: Option<f32>,
-    ) -> irlume_common::Result<Vec<CapturedScan>> {
+    ) -> irlume_common::Result<(Vec<CapturedScan>, CaptureShape)> {
         // Hold the cameras open for the whole loop. This is the heaviest repeated
         // capture in the codebase (the budget below is ten assessments per wanted
         // scan), and every one of them otherwise re-opened, re-negotiated,
@@ -3184,8 +3184,14 @@ impl Engine {
             &mut irlume_camera::IrSession<'_>,
         )>,
         capture_mode: Option<(bool, &'static str)>,
-    ) -> irlume_common::Result<Vec<CapturedScan>> {
+    ) -> irlume_common::Result<(Vec<CapturedScan>, CaptureShape)> {
         let mut out = Vec::new();
+        // Read once, before the loop: `sessions` is borrowed per iteration but
+        // never taken, so this is the whole loop's answer (#389).
+        let mut shape = CaptureShape {
+            held_sessions: sessions.is_some(),
+            ..CaptureShape::default()
+        };
         // Budget (was ×4) absorbs the added frontality gate (a frame grabbed the
         // instant the user drifts off-angle is rejected, not saved) with enough
         // retries that a brief drift near the capture moment doesn't abort enroll.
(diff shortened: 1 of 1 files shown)


archledger:
Applied in `34dbc3c`. The finding is right, and both of its premises were checked in the tree before I touched anything: the capture gate at `capture_scan_loop` keeps a scan only under `if let Some(e) = a.embedding`, so reaching the top-up proves RGB succeeded once, and the next line then replaced that tally with the top-up's.

**The first fix was wrong in an instructive way.** I folded at the call site with a `CaptureShape::include`, and then a mutation reverting that single line back to an assignment passed every test, because the tests exercised the folding helper rather than the site that calls it. That is the same defect as the argument swap you found one round earlier, so I changed the shape instead of adding another test: `capture_scans` now takes the enrolment's tally by reference and folds into it, both loops receive the same one, and there is no value left for a caller to replace.

`include` adopts on the first fold rather than intersecting, because every enrolment starts from a default that has never seen a loop and ANDing `held_sessions` against it would zero the flag immediately.

Seven mutations now, each run and each killed: the two trigger conditions, the two message properties, and three on the fold itself.

One correction while running them. My mutation harness split its cases on `|`, which is also the character in `|=`, so the AND-to-OR case was mangled into a syntax error and its "kill" was a compile failure rather than a test failure. Re-run on its own it fails the intended assertion, `held_sessions must AND across the loops`. A mutation that dies for the wrong reason is not a kill, and I would have reported it as one.

PR #416, opened 2026-08-10
Refuse to enable a gate that refuses the user (#386)

archledger:
Addresses #386. The issue stays OPEN: this stops the harm, it does not make the gate work.

`require_eyes_open` cannot admit the user it exists for. Replaying the committed
blink corpus through the shipped `both_eyes_open`, twelve detected frames per
state: bare-eyed with eyes genuinely open reads open 1 time in 12, and with
glasses 0 times in 12. Across all six cells the gate admits 1 of 72 genuine
detected frames. There is no eyewear or lighting state in the record where
turning it on leaves face authentication usable.

So turning it on is refused, and turning it off is always allowed. Refused
rather than warned because a warning implies a tradeoff the measurement does not
offer: there is no configuration that works, so there is nothing to trade.

The refusal sits at the daemon's dispatch arm because that is the single choke
point. `irlume profiles eyes-open on` and the TUI toggle both send
`SetRequireEyesOpen`, so a check in either would leave the other open.

An enrollment that already carries the flag is not trapped. The OFF direction
never reaches the refusal, and the message says so, because a refusal that named
only the ON direction would read as a dead end to the one person who most needs
the escape.

What would lift this is a cue that survives a change of light or eyewear, and
#386 now records that none is available: the peak cannot be retuned because the
two distributions overlap completely behind glasses, and a per-user EAR
threshold fails across sessions, the same subject's median OPEN reading landing
inside the open/closed separation window measured in another session.

Proved by mutation, each run and each naming the assertion that caught it:

- The refusal arm removed: fails.
- The message stops naming the issue: fails.
- The message drops the hint that OFF still works: fails.
- The guard widened to swallow OFF as well: fails.

That last one is the reason this test is split in two. The first version carried
a `tpm_available()` skip copied from its neighbour, which meant it returned early
on any developer machine with a TPM and survived all four mutations while
reporting a pass. The refusal touches no storage, so it needs no TPM-free host
and is now asserted unguarded; only the write check and the OFF round trip, which
end in `storage::save`, keep the guard. The OFF direction is additionally
asserted without storage, by dispatching for a user with no enrollment: a correct
guard lets that through to the missing-enrollment error, and only a widened one
answers with the refusal text.

Not covered: no camera. This changes no capture, no verdict and no matching. It
changes which requests the daemon accepts, and the numbers behind the decision
come from stored frames verified against `docs/pad-results/2026-08-07-blink-corpus.sha256`.

### DIFF
--- a/crates/irlume-cli/src/main.rs
+++ b/crates/irlume-cli/src/main.rs
@@ -1129,7 +1129,8 @@ fn usage_profiles() -> std::process::ExitCode {
         delete --profile P [--scan S]           delete a profile or a scan\n  \
         forget-model <model>                    remove one recognizer's scans from every\n  \
                                                 profile (shipped | a catalog name | embed:<sha256>)\n  \
-        eyes-open <on|off>                      require eyes open to unlock\n  \
+        eyes-open off                           turn OFF the eyes-open check\n  \
+        \x20                                       (it cannot be turned on; see issue #386)\n  \
         challenge <on|off>                      opt-in passive blink liveness"
     );
     std::process::ExitCode::from(2)
--- a/crates/irlume-cli/src/tui.rs
+++ b/crates/irlume-cli/src/tui.rs
@@ -3367,6 +3367,17 @@ impl App {
             // Settings.
             (SC_SETTINGS, KeyCode::Enter) | (SC_SETTINGS, KeyCode::Char(' ')) => {
                 let on = !self.eyes_open;
+                // Turning it ON is refused by the daemon (#386), so do not fire
+                // a request known in advance to fail. The refusal still lives
+                // there, because that is the one choke point both this and the
+                // CLI go through; this only avoids offering the user an action
+                // whose only outcome is an error modal.
+                if on {
+                    self.set_error(
+                        "require-eyes-open cannot be enabled: it refuses the user it exists                          to admit (measured 1 of 12 bare-eyed frames with eyes open, 0 of 12                          with glasses). See issue #386.",
+                    );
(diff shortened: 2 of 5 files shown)


archledger:
Both findings applied in `d324409`, each checked in the tree before I acted on it.

**The contract break is the one that mattered.** I changed what the daemon accepts and left `docs/COMMANDS.md`, the CLI usage text and the TUI all still offering the enable. For a security setting that is worse than stale wording: the interface kept presenting protection that could not be selected, and the TUI actively fired a request known in advance to fail.

Fixed by removing the advertisement rather than duplicating the policy. The docs and usage name only the OFF direction and say where the refusal is recorded; the TUI raises the reason instead of sending; the refusal itself stays in the daemon, which is the choke point both producers share.

**The cache eviction was a genuine invariant violation**, and the invariant is stated at the line itself: a request about to be refused may not change state, because an unprivileged peer could otherwise evict root's summary and charge root's next listing a storage load and its TPM work (#349). My refusal ran after it. It now sits above, with the other refusals, and a test publishes a summary, dispatches a refused enable, and asserts the summary survives. Moving the refusal back below the invalidation fails that test.

Worth recording separately: the first version of the main test carried a `tpm_available()` skip copied from its neighbour. This machine has `/dev/tpm0`, so it returned early and **survived all four mutations while reporting a pass**. The refusal touches no storage and needed no such guard. It is now split, with the refusal asserted unguarded and only the storage round trip keeping the skip, and the OFF direction additionally asserted without storage by dispatching for a user with no enrollment.

PR #417, opened 2026-08-10
Name the cause when a solo frame proves it (#389)

archledger:
Addresses #389. The issue stays OPEN for options 1 and 4, which change capture behaviour and re-measure stale verdicts; this changes neither.

#414 taught the enrolment failure to offer a second reading, and would not let
it name one, because while both streams run an unlit room and a starved RGB
interface are the same reading: no RGB face, IR face present. That was the right
call on the evidence then. It is no longer the evidence.

They stop being the same once the sessions are released. Measured on a NexiGo
HelloCam N930W, 2026-08-10, ten runs across three conditions, `frame_mean`
throughout so the numbers are compared against the statistic the constants were
fitted to:

| condition | held | solo, after release | rule |
|---|---|---|---|
| lit room, starved | 0 of 6 faces, mean 51.7 / 51.1 | face at 0.95, mean 146.9 / 146.6 | confirms |
| dark room | 0 of 6 faces, mean 46.5 to 47.2 | no face, mean 18.0 in all four runs | refuses |
| ASUS built-in, lit | face in 6 of 6, mean 160 to 163 | face, mean 156 to 158 | refuses |

So on a failed enrolment whose attempts all had the IR-only shape, the daemon
now takes ONE solo RGB frame and says which it was. Confirmed, the message names
the camera and drops the lighting clause, because the room being lit is part of
the confirmation and sending that user after their lamps would be wrong.
Refuted, the speculative second reading is dropped instead, because a solo frame
that also found nothing is evidence against the camera. No probe, and the
message is exactly what #414 shipped.

Safe where the cross-spectrum self-heal is not. That recapture is forbidden
while sessions are held, because reopening a node this process streams answers
EBUSY on some modules (#187, #381). This runs after `capture_scans` has returned
and both sessions are dropped, so no node is reopened while streamed. It costs
one RGB open, 146ms to 173ms on that module, and only on a loop that already
failed. Every error path answers `None`, so a failed probe can never turn a
failed enrolment into a different error.

The dark room is refused TWICE, and that is worth stating because it is why this
does not rest on one constant being right for every room. With the emitter
firing during the held phase its light leaks into the RGB sensor, so the held
frames read BRIGHTER than the solo one, 46.6 against 18.0, and the dimming
clause fails on its own before the brightness floor is consulted.

`irlume_camera::frame_mean` is now public. `signals.rgb_face_brightness` could
not serve here: it measures the face region, so it is 0.0 by construction in
exactly the case this probe exists for, and a private reimplementation would
have compared one brightness definition against thresholds fitted to another.

Proved by mutation, each run and each naming the assertion that caught it:

- The face requirement dropped: fails.
- The dimming clause dropped: fails.
- The brightness floor dropped: fails, but only after a cell was added to
  isolate it. The measured dark room is refused twice over, so with only the
  hardware cells the floor could be deleted and every test still passed. The
  new cell is a dim scene where the solo frame IS brighter, so the dimming
  clause passes and the floor is the only thing left to refuse.
- The confirmed message keeping the lighting clause: fails.

Not covered: no camera in CI. The decision function and the message rule are
unit-tested against the measured cells; the capture itself is not, because it
needs a device. The IR arm was dark in the positive runs, so the RGB-starvation
half of the signature is measured and the IR-kept-the-face half is not. One
subject, one room, one evening, two modules.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -4104,12 +4104,20 @@ fn capture_advice(shape: CaptureShape, solo_probe: Option<bool>) -> String {
         // `solo_mean >= CONCLUSIVE_SCENE_BRIGHTNESS`, so the room being lit is
         // measured rather than assumed, and telling this user to check their
         // lighting would send them after the wrong thing.
+        // The light comes FIRST, and that ordering is measured rather than
+        // stylistic. On a healthy camera in a dark room with a lamp coming on
+        // between the two captures, this branch fires wrongly: 4 runs of 4 on
+        // 2026-08-10, held 28.9 to 31.8 with no face, solo 163 with one. Naming
+        // the camera first would put the wrong cause at the front of the
+        // sentence in every one of them. The second held phase that WOULD
+        // separate the two is what #379's config write needs; a message does
+        // not earn a second pair of session opens on a failed enrolment.
         Some(_) => String::from(
             "the colour frame was dark on every attempt while both sensors were streaming, and \
              a capture taken straight afterwards with only the colour sensor running found a \
-             face. That is the shape of a camera that cannot sustain both streams; run `sudo \
-             irlume camera-tune` in a lit room to measure it directly. Light that changed \
-             between those two moments would look the same from here",
+             face. If the light changed between those two moments, that is the explanation. If \
+             it did not, this is the shape of a camera that cannot sustain both streams, and \
+             `sudo irlume camera-tune` in a lit room measures that directly",
         ),
         // Not confirmed, whether the probe refuted it or never ran. Unchanged
         // from #414: offer both readings, assert neither.
@@ -5528,9 +5536,19 @@ mod tests {
             "{confirmed}"
         );
         assert!(confirmed.contains("camera-tune"), "{confirmed}");
+        // The confound is named FIRST, because on a healthy camera in a dark
+        // room with a lamp switching on this branch fires wrongly in 4 runs of
+        // 4. Leading with the camera would put the wrong cause at the front of
+        // the sentence every one of those times.
+        let light = confirmed
+            .find("If the light changed")
+            .expect("names the light");
+        let camera = confirmed
+            .find("cannot sustain both streams")
+            .expect("names the camera");
         assert!(
-            confirmed.contains("Light that changed"),
-            "a confirmation must name the confound it cannot rule out: {confirmed}"
+            light < camera,
+            "the explanation that cannot be ruled out must come first: {confirmed}"
         );
         assert!(
             !confirmed.contains("so it is dimming"),
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -155,6 +155,13 @@ pub struct Assessment {
     /// the scene did, and such scans have never proven they work without
     /// that scene light.
     pub ir_ambient_share: Option<f32>,
+    /// Mean of every byte in the RGB frame, whole-frame rather than the face
+    /// region: the enrolment starvation probe needs a reading from a frame where
+    /// no face was found, which is exactly when `signals.rgb_face_brightness` is
+    /// 0.0 by construction. Computed with `irlume_camera::frame_mean`, the same
+    /// statistic `CONCLUSIVE_SCENE_BRIGHTNESS` and `CONCURRENT_SIGNAL_FLOOR` were
+    /// measured against (#389).
+    pub rgb_frame_mean: f32,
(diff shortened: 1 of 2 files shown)


archledger:
Both findings applied in `73ae621`, and the first one names a defect class that is already on my own list, which makes it the more useful of the two.

**The message asserted a cause it had not established.** It said the camera "is dimming its colour stream under concurrent capture". Nothing in the code records that the light, the framing or the person stayed the same between the held attempts and a solo frame taken seconds later, so a lamp switching on or the subject stepping back into frame produces the identical reading with no camera fault. The message now reports what was observed, names `camera-tune` as the thing that measures it directly, and says out loud that light which changed between those two moments would look the same from here. The test forbids the old wording rather than merely preferring the new.

**The refutation direction was worse than useless**, and I had not seen it. `Some(false)` deleted the second reading, so a user who turned away before the solo frame would suppress a hint about a camera that really was starving, on the strength of an observation that tested a different scene. A refutation is now treated as no probe at all, and the test asserts the refuted and unprobed messages are equal rather than checking each separately.

**A detector error was collapsed into "no face was there"**, which let a broken detector read as a refutation. It returns `None` now.

Two further mutations, each run and each naming its assertion: restoring the refutation-deletes-hint behaviour fails, and dropping the confound sentence from the message fails.

On the generalisation point in your closing section, I agree and have not tried to argue past it: one subject, one room, two modules, and the IR arm was dark in the positive runs. The message is scoped to that, and #379 wants this same evidence for a config write, which is a higher bar than a sentence.

archledger:
## The config write has its evidence, and it is not the design that shipped in #417

Sixteen runs, 2026-08-10, one subject, two modules. The question was whether a solo frame after release can carry a durable config write, and the answer is that it cannot on its own, but held / solo / held-again can.

| cell | camera | runs | held A | solo B | held A' | A/B | A/B/A |
|---|---|---|---|---|---|---|---|
| lit, starving | NexiGo | 4 | 50.2 to 50.6, 0 of 5 faces | 139.7 to 140.5, face | 50.2 to 50.8, 0 of 5 | true | true |
| lit, healthy | ASUS | 4 | 153.2 to 153.7, 5 of 5 | 149.3 to 151.2, face | 153.0 to 154.3, 5 of 5 | false | false |
| dark, light switched on between B and A' | ASUS | 4 | 28.9 to 31.8, 0 of 5 | 162.7 to 163.7, face | 168.4 to 174.4, **5 of 5** | **true** | **false** |
| dark, starving | NexiGo | 4 | 46.5 to 47.2, 0 of 5 | 18.0, no face | (earlier session) | false | n/a |

### The A/B is wrong 4 times out of 4 in the adversarial cell

A healthy camera, a dark room, and a light coming on between the two captures. That is not a contrived scenario, it is walking up to a sleeping machine with a motion-controlled lamp. The single solo frame confirms contention every time, and there is no camera fault.

The A/B/A rejects all four, and the reason is physical rather than statistical: a healthy camera's held phase TRACKS the light, so A' returns at 168 to 174 with a face in 5 of 5 frames and the reversal fails.

### Why it still confirms on the module that is genuinely starving

On the NexiGo the held frame reads 46 to 51 whether the room is lit or dark, while the solo frame moves from 18.0 to 140. The held reading is not measuring the scene at all there; it is pinned by the starved stream. So A' stays dark with no face and the reversal holds.

That is the discriminator stated properly: not "is the solo frame brighter", which a lamp satisfies, but "does the held phase track the scene or is it pinned". Two held phases either side of a solo one is the cheapest way to ask.

### What this means for each of the three

**#417, the message, stands.** It uses the A/B, so in the adversarial cell it will say the wrong thing, and this measurement puts a number on how often: 4 of 4 in that cell. It is acceptable there only because the message names that exact confound out loud and because the remedy it prescribes, `camera-tune`, measures the thing directly and would find nothing on a healthy camera. The cost of being wrong is one wasted measurement.

**#379, the config write, needs the A/B/A.** A wrong write is durable, overwrites a possibly-measured verdict, and taxes every later capture inside fixed retry budgets. A design with a 4-of-4 false-positive rate in a realistic scenario cannot carry that, and now it does not have to.

**#100's item 4 becomes cheaper to justify either way.** A re-measure makes a wrong inference cost one slow session rather than forever, which lowers the evidence bar rather than raising the evidence.

### Not established

One subject, one room, one evening, two modules. The adversarial cell used a deliberate light change, which is the strongest form of that confound rather than the average one. A camera that starves AND sits in changing light would still confirm, and correctly so, since it is starving. Nothing here measures a second starving module, and the IR arm was dark in the NexiGo positives, so the RGB half of the signature is what these numbers describe.

Issue #428, opened 2026-08-12
Node classification can stop opening nodes: udev, media topology and USB descriptors answer without an open


archledger:
`classify_node` opens every `/dev/video*` node and runs ENUM_FMT to tell RGB from IR from metadata nodes, and `scan_nodes` / `discover_nodes` inherit that. Two facts from the sources change how this should work.

First, the #187-era belief that a probe open contends is wrong. The uAPI is explicit ("Merely opening a V4L2 device does not grant exclusive access", `userspace-api/media/v4l/open.rst`) and exclusivity begins at buffer allocation; `uvc_v4l2_open` on mainline is a pure allocation whose only error is ENOMEM. The EBUSY measured in the #187 hardware session came from S_FMT against an allocated queue (`vb2_is_busy`), which matches the strace in `RgbSession::recover`'s comment. The probe open did have a different cost: up to kernel 6.15 it powered the USB interface up (and flashed privacy LEDs on cameras that tie the LED to interface power); the 6.16 granular-power series moved power-up into the ioctl dispatcher, and neither QUERYCAP nor ENUM_FMT is on the power-up list. So on current kernels the existing probe is already contention-free and power-free; on the kernels Debian stable ships it still blips the camera once per scan.

Second, a classification path that opens nothing exists, with documented footing:

- Capture vs metadata: udev's `60-persistent-v4l.rules` runs `v4l_id` once at plug time; it prints `ID_V4L_CAPABILITIES=:capture:` for capture nodes and an empty list for metadata nodes (v4l_id has no branch for `V4L2_CAP_META_CAPTURE`). Reading udev properties opens no device.
- Grouping nodes per physical camera and finding the default capture node: `/dev/media*` topology. The media docs guarantee the open "has no side effects; the device configuration remain unchanged", `media_device_open` is a bare `return 0`, and uvcvideo flags the default capture node `MEDIA_ENT_FL_DEFAULT`.
- IR vs RGB without ENUM_FMT: the sysfs USB `descriptors` blob (stable kernel ABI since 2.6.26) carries the VideoStreaming format GUIDs, and the driver's own table in `drivers/media/common/uvc.c` maps them (`UVC_GUID_FORMAT_Y8` to GREY and so on); uvcvideo builds its format list from those same bytes at probe and never asks the device. `uvc_descriptor.rs` already parses this blob for the emitter XU, so the parser exists and needs the VS format descriptors added.

[...]

Issue #427, opened 2026-08-12
Another process can retarget the format between S_FMT and buffer allocation; read it back


archledger:
In uvcvideo the negotiated format is per-device state, not per-file-handle: `uvc_ioctl_s_fmt` writes `stream->cur_format` on the shared `uvc_streaming` struct, gated only by `vb2_is_busy` (`drivers/media/usb/uvc/uvc_v4l2.c:446`, mainline `f5bbbfe`). Buffer ownership begins at REQBUFS, not at open (`userspace-api/media/v4l/open.rst`: the filehandle that allocates buffers "becomes the owner of the device"; "Merely opening a V4L2 device does not grant exclusive access").

irlume negotiates its format in `RgbCamera::open` / `IrCamera::open` (TRY_FMT walk, then S_FMT) and allocates buffers later, in `session()` when `SafeStream::open` runs REQBUFS. Between those two calls any other process can issue its own S_FMT and silently retarget the device: nothing holds the format for us, because we own no buffers yet. The session then dequeues frames and decodes them against the width, height and fourcc recorded at open time. Wrong dimensions produce garbage decode or a short-buffer error at best; at worst a YUYV/NV12 confusion decodes into plausible-looking noise that flows into detection.

The window is small and the collision needs another camera user active at auth time, but the check that closes it is one ioctl: after buffer allocation, read G_FMT and refuse the session when it does not match what `open` negotiated, with an error naming the other consumer if the holder scan can see one. Refusing is correct rather than renegotiating: a format that moved under us means another application is actively configuring the device, and the polite behaviour (the same conclusion as the emitter's foreign-consumer stand-down) is to fail this capture toward the password prompt, not to fight over shared state.

The same read-back belongs in `recover()`, which rebuilds the stream on the already-open device after a mid-stream fault; the fault window is precisely when another process may have grabbed and reconfigured the node.

Source research with citations: [`docs/research/2026-08-12-camera-handling-audit.md`](https://github.com/archledger/irlume/blob/main/docs/research/2026-08-12-camera-handling-audit.md) (Q3).

PR #434, opened 2026-08-12
Restore what the backlight-compensation write displaced (#426)

### LINKED ISSUE #426: The backlight-compensation write persists for the next application and is never restored
`RgbCamera::session_with_progress` writes `V4L2_CID_BACKLIGHT_COMPENSATION = 2` best-effort at every RGB session start, and nothing ever restores it. The V4L2 control documentation says control values "are stored globally, they do not change when switching" and "do not change e. g. when the device is opened or closed" (`userspace-api/media/v4l/control.rst`), so the value survives irlume closing the device and lands on the next application: a video call after a face auth runs with irlume's backlight preference until something else rewrites it.

The emitter path earned a whole guard type (`StreamMode`) for exactly this class of write: capture what the write displaced, restore on session end, read back before restoring. This one predates that discipline and is the only remaining device-state write in the crate with no restore path.

Fix shape, smallest first:

1. Read the current value before writing, skip the write when it already matches, and restore the displaced value when the session drops (same displaced-value rule as the emitter guard: put back what was there, not the driver default, so another program's deliberate setting is not destroyed).
2. Or, if measurement shows the write only matters on specific cameras, keep it behind the same per-camera evidence the emitter uses. The justification on record is a NexiGo N930W measurement (face mean 49 to 124 with compensation on); no other camera has a recorded benefit.

One wrinkle from the kernel source: `uvc_ctrl_restore_values` re-applies host-modified controls after a resume with reset, and standard mapped controls (this one included) are eligible. So irlume's written value can be re-imposed by the kernel across a suspend even after irlume exits; restoring the displaced value at session end is what keeps that re-imposition harmless, because the last host-set value becomes the one the user's other software chose.

Source research with citations: [`docs/research/2026-08-12-camera-handling-audit.md`](https://github.com/archledger/irlume/blob/main/docs/research/2026-08-12-camera-handling-audit.md) (Q7).

archledger:
Closes #426.

The defect was found live on the machine that measured it: hours after an RGB session, `backlight_compensation` still read 2 against a driver default of 0, because control values persist across open and close by the V4L2 specification and the write was fire-and-forget. The issue's open question (keep-with-restore versus drop) is settled by measurement: the write earns its keep on both cameras with recorded numbers (NexiGo N930W face mean 49 to 124; this ASUS FHD center mean 138.5 to 150.6 in this session's committed measurements), so it stays, and what it displaces goes back.

Design, the emitter guard's displaced-value rules at standard-control scale:

- Read first; write only when the control holds something other than the wanted value. A control already at 2 is another writer's state: no write, no restore armed, never adopted as irlume's to undo.
- An unreadable or absent control is not a license to write (the old code wrote blind, best-effort; now it skips, same net effect on cameras without the control, no invented displaced value).
- On session drop, read back and restore the displaced value only while the control still holds what irlume wrote; a moved control carries somebody's newer choice and stays.
- The displaced VALUE goes back, never a hardcoded default: the ThinkPad's built-in camera defaults this control to 1 (read over SSH this session), so restore-to-zero would be wrong hardware that exists in this project.
- Everything stays best-effort: a failure costs picture tuning for the next app, never an authentication.

Testing:

- Both decisions are pure with every arm asserted; two mutations against the committed tree (the write adopting a foreign 2; the restore overwriting a moved control), each killed by the named test, suite green on the restored tree.
- Hardware-validated on the real ASUS camera with this branch's binary, both directions: default 0 before a session and 0 after (the merged main binary leaves 2, which is how the defect announced itself), and a pre-set 2 surviving a session untouched.
- Session measurements committed as `docs/research/2026-08-12-camera-session-measurements.md` (also carries the #428 on-hardware verification of udev properties, media topology, and the descriptors blob, including the D3DFMT_L8 GUID this laptop's IR camera advertises).
- `cargo +1.88.0 fmt --check`, clippy `--all-targets -D warnings`, the doc gate, and the guarded workspace suite (1387 tests) green locally.

Not tested:

- The Drop wiring has no hardware-free trigger (v4l2loopback exposes no backlight control), so the decision functions carry the unit coverage and the two hardware runs above are the wire-level evidence.
- The NexiGo itself was unreachable this session (minihost asleep); its effect number is the recorded one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ All notable changes to irlume are documented here. This project adheres to
 
 ### Fixed
 
+- **The backlight-compensation tuning no longer outlives the session onto
+  other applications' pictures.** RGB sessions write
+  `V4L2_CID_BACKLIGHT_COMPENSATION=2` so auto-exposure favors the face over
+  a bright window (NexiGo N930W: face mean 49→124; ASUS FHD: center mean
+  138.5→150.6), but control values persist across open and close by the
+  V4L2 specification, and the old fire-and-forget write was measured still
+  applied on a laptop hours after the session that wrote it, with the
+  driver default being 0. A session now reads the control first, writes
+  only when it holds something else, remembers what the write displaced,
+  and puts that value back when the session drops, under the emitter
+  guard's rules: a control already holding the wanted value is another
+  writer's state and is never adopted, and the restore fires only while
+  the control still holds what irlume wrote. Both directions validated on
+  hardware in the 2026-08-12 session measurements
+  (`docs/research/2026-08-12-camera-session-measurements.md`)
+  (#426).
+
 - **An emitter control that healed by power-cycle no longer stays blocked
   forever, and the advice now says what actually clears one.** When the
   restore-attempt budget for a leftover emitter control ran out, the refusal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,11 +16,15 @@ All notable changes to irlume are documented here. This project adheres to
   applied on a laptop hours after the session that wrote it, with the
   driver default being 0. A session now reads the control first, writes
   only when it holds something else, remembers what the write displaced,
-  and puts that value back when the session drops, under the emitter
-  guard's rules: a control already holding the wanted value is another
-  writer's state and is never adopted, and the restore fires only while
-  the control still holds what irlume wrote. Both directions validated on
-  hardware in the 2026-08-12 session measurements
+  and puts that value back through a guard armed before the stream opens,
+  so a failed open restores too. The emitter guard's rules apply: a
+  control already holding the wanted value is another writer's state and
+  is never adopted; the write is confirmed by read-back (drivers may
+  clamp a set to the nearest valid value, and an unconfirmed write is
+  undone on the spot); and the restore fires only while the control still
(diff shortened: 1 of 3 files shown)


archledger:
Review round applied, both merge-blocking findings and both follow-ups:

- **Finding 1 (restore had no owner on a failed open):** the restore now lives in a `BlcRestore` guard armed immediately after a confirmed write, before the stream opens, so a REQBUFS failure or the #427 format-moved refusal restores too. The session holds the guard declared after the stream; field drop order tears the stream down before the control goes back. The session's own `Drop` is gone.
- **Finding 2 (an accepted write is not proof the device holds 2):** the write is confirmed by read-back through the same predicate that authorises the eventual restore; an unconfirmed or clamped write is undone on the spot and arms nothing.
- **Follow-up 3:** the changelog and the decision doc now state the no-overwrite rule as holding at the restore's own read, since V4L2 offers no compare-and-set.
- **Follow-up 4:** the comment cites the committed measurements file.

Re-validated on the real ASUS camera with the updated branch binary: default 0 survives a session, and a pre-set 2 survives untouched. Full local gate green (fmt, MSRV clippy all-targets, doc, guarded workspace suite at 1387).

The earlier red checks on this PR were a GitHub release-asset outage (HTTP 503 on the ONNX runtime and model-weights downloads, reproduced from outside CI); assets recovered and CI was re-run.

Issue #425, opened 2026-08-12
Every Intel IPU6 capture node classifies as an RGB camera; gate classification on V4L2_CAP_IO_MC


archledger:
On any Intel IPU6 laptop (Tiger Lake through Meteor Lake MIPI cameras), every ISYS capture node classifies as an RGB camera. The `isys` driver registers up to eight capture nodes per CSI-2 port, all single-planar `V4L2_CAP_VIDEO_CAPTURE`, and its `ENUM_FMT` answers from a static table (YUYV, RGB565, GREY, Y10, Y16, every Bayer variant) whenever `mbus_code` is 0, regardless of which sensor is attached (`drivers/media/pci/intel/ipu6/ipu6-isys-video.c`, format table and `ipu6_isys_vidioc_enum_fmt`, verified on mainline `f5bbbfe`). `role_from_formats` sees YUYV and returns `Role::Rgb`.

The consequences are wrong output, not hardware damage: `scan_nodes` and `doctor` report a fleet of cameras that do not exist, and an auth attempt that selects one gets a successful S_FMT (state-only) and then EPIPE at STREAMON because the media-controller pipeline was never wired. Opens are inert on this driver, sensor links are immutable, and power is kernel/firmware owned, so nothing can be damaged; the failure is that irlume asserts facts about hardware it does not understand.

The gate is one QUERYCAP field. The V4L2 uAPI (`userspace-api/media/v4l/open.rst`) says a device reporting `V4L2_CAP_IO_MC` in `device_caps` "is MC-centric", and for MC-centric devices the video node's format list describes what the IP core can consume, not what the camera produces. So:

- Before format-based classification, read `device_caps`. If `V4L2_CAP_IO_MC` is set, classify the node as a distinct role (not `Other`) and have doctor say what it is: a MIPI camera behind a media-controller stack; the RGB side may work through libcamera/PipeWire; the IR side has no Linux support today (the Hello-era OV01A1S sensor has no mainline driver, the out-of-tree driver emits IR as SGRBG10 Bayer, and no emitter interface exists in mainline, intel/ipu6-drivers, or ipu6-camera-hal).
- Make the MPLANE ignore deliberate: `ipu3-cio2` (Surface) and `qcom-camss` (Snapdragon X) nodes are MPLANE-only, so today's single-planar ENUM_FMT probe files them under `Other` by accident of the probe shape. A driver-name match gives Surfaces an honest message too.
[...]

PR #435, opened 2026-08-12
Classify camera nodes without opening them (#428)

### LINKED ISSUE #428: Node classification can stop opening nodes: udev, media topology and USB descriptors answer without an open

archledger:
Closes #428.

Scanning opened every `/dev/video*` node to enumerate formats. On kernels before 6.16 each open powers the camera up (a privacy-LED blink per scan), and those probe opens were the premise behind the #187 workaround. The audit established the contention-free route, the READY session verified every piece on real nodes (`docs/research/2026-08-12-camera-session-measurements.md`), and this implements it.

The no-open primary answers the two questions classification needs:

- **Capture or metadata sibling:** the media graph. Opening `/dev/media*` is the one probe the kernel documents as side-effect free; the capture entity owns pads while the metadata entity registers padless. The decision keys on pads rather than `MEDIA_ENT_FL_DEFAULT` (the flags word is only documented valid from media_version 4.19). The owning media node is a plain subdirectory of the video node's USB interface directory, so nothing iterates other cameras' graphs.
- **IR or RGB:** the sysfs descriptors blob, grouped per function by Interface Association Descriptor (spec-layout contiguity when a device has none), translated to fourccs through a table copied from the kernel's own (`include/linux/usb/uvc.h` + `drivers/media/common/uvc.c`), and fed to the same `role_from_formats` the open probe uses, so the two paths cannot disagree. The table carries the whole L8 greyscale family; the reference camera advertises **KSMEDIA_L8_IR**, whose Data2 defeats tail-keyed matching, and the session doc's D3DFMT_L8 misreading is corrected in the same commit.

Everything undecidable is `None` and falls back to the open probe unchanged: loopback nodes (the CI lane keeps exercising that path), MC-centric platform stacks (the #425 QUERYCAP gate still names them), vendor-only format lists, any failed sysfs read. The fall-through is Option-shaped, so the no-open path can remove opens but never invent a classification.

Testing:

- The committed ASUS descriptor fixture classifies both functions end to end (RGB: MJPG+YUYV; IR: exactly GREY via KSMEDIA_L8_IR), the wire-order GUID presence is pinned the same way the MSXU constant is, and a synthetic no-IAD blob pins the contiguity rule.
- The media decision runs against the measured graph shape, a padless-metadata case, an absent-node case, a shadowing data link, and a future link kind sharing bit 28; the last one was added because the mask-versus-bit mutant survived the first suite and my own comment had predicted the only distinguishing case.
- Four mutations against the committed tree, all killed by named tests with clean compiles. The first mutation round produced three false kills (a self-corrupting restore made later mutants fail to compile); the harness now asserts uniqueness in both directions and the round was re-run.
- Hardware: strace open-counting on this machine's UVC pair: `doctor` video-node opens drop from 17 to 9 (the survivors are the deliberate per-node privacy/backend/format/stream checks), with byte-identical camera classification output between the main and branch binaries.
- `cargo +1.88.0 fmt --check`, clippy `--all-targets -D warnings`, the doc gate, and the guarded workspace suite (1392 tests) green locally.

Not tested:

- The three-line glue in `classify_without_open` and the live `MEDIA_IOC_G_TOPOLOGY` read have no hardware-free trigger; the decisions on both sides are pure and pinned, and the strace run above is the wire-level evidence on real hardware.
- No non-UVC media graph (IPU6-class) was available to observe the accepted asymmetry the doc comment records: a padless node on such a stack answers `Other` here without reaching the MC-centric gate that would have named it.
- Hotplug re-enumeration of the sysfs paths was not exercised.


archledger:
Review round applied by adopting the replacement wholesale: the descriptor format route is removed, not repaired.

- Findings 1 through 3 (per-function union versus per-node streams, the subset GUID table whose omissions change the enumerated set, quirk non-parity) each break the PR's own "cannot disagree" claim, and repairing them means reproducing uvcvideo's stream attribution and quirk tables in userspace, which is the wrong side of the maintenance bet. Capture nodes answer `None` in `classify_without_open` and keep the authoritative ENUM_FMT probe, with the three reasons recorded at the decision.
- Finding 4 (u8 overflow in the IAD range check) is resolved by the route's removal.
- The media-graph metadata filter survives unchanged with its tests: metadata nodes, half of a UVC camera's nodes, classify with no video-node open.
- The GUID table and function walk are deleted rather than kept dead; the KSMEDIA_L8_IR facts remain in the committed session measurements for whenever a faithful per-stream route is attempted.

Re-measured on the reference machine: `doctor` video-node opens drop from 17 on main to 13 on this branch (the four metadata-node opens), classification output byte-identical. Full gate green locally (fmt, MSRV clippy all-targets, doc, guarded workspace suite at 1389).

Issue #385, opened 2026-08-09
The IR clipping ceiling is knowable for NV12 and YUYV too, and the reason it is not read is stale


archledger:
Follow-up from #358. Now that an IR format naming no sensor ceiling is refused rather than passed, it is worth asking which formats genuinely cannot name one. The answer is narrower than the code assumes, and the reason written in the code is stale.

## The stated reason no longer holds

`IrCaptureStats::white_level` says of NV12 and YUYV:

> their ceiling depends on the negotiated quantization (limited range puts nominal white at 235, full range at 255) and irlume does not carry that field

irlume does carry it. `IrCamera::open` stores `quantization: fmt.quantization` from the negotiated format, and threads it into `IrDecoder::new` at five sites; `clipping_white_level(pix, quantization)` already takes it and already uses it for the GREY arm. The same claim appears three times: `irlume-camera/src/lib.rs` around the `white_level` doc and again in its tests, and `irlume-liveness/src/lib.rs`.

`decode_ir` also already puts all three on one scale, byte for byte: `Grey8 => buf.to_vec()`, `Nv12Luma => buf[..luma].to_vec()`, `YuyvLuma => every other byte`. Only `Grey16` rescales.

## What V4L2 says

Kernel documentation, checked against the current tree:

- The Y samples of GREY, NV12 and YUYV are all 8 bits. `v4l2_format_info` classes all three as `V4L2_PIXEL_ENC_YUV`.
- `quantization` is a field of `struct v4l2_pix_format` with no per-format exclusion. The neighbouring fields carry explicit ones (`ycbcr_enc` "is ignored for HSV pixelformats"), and `quantization` does not.
- `V4L2_QUANTIZATION_LIM_RANGE` maps [0..1] to [16..235]; `V4L2_QUANTIZATION_FULL_RANGE` maps it to [0..255].

So given the same reported quantization, the ceiling is the same number by the same rule for all three.

## The trap, and why this is not a one-line change

`V4L2_MAP_QUANTIZATION_DEFAULT` resolves `DEFAULT` to LIMITED range for Y'CbCr unless the colorspace is JPEG:

```c
#define V4L2_MAP_QUANTIZATION_DEFAULT(is_rgb_or_hsv, colsp, ycbcr_enc) \
	(((is_rgb_or_hsv) || (colsp) == V4L2_COLORSPACE_JPEG) ? \
[...]

PR #436, opened 2026-08-12
Keep the clipping-ceiling comments true after #427 (#385)

archledger:
Doc-only. The `white_level` explanation and its liveness mirror said the colorspace needed to resolve `Default` quantization "is not retained"; since #427 the camera stores the whole driver-echoed `Format`, so that sentence joined the one it had replaced in being stale. The comments now say the resolution is computable but unvalidated (no NV12/YUYV IR camera exists in this project's record), and that the `None` is load-bearing regardless: it is what makes `exposure_refusal` refuse a colour stream arriving through the role-blind paths, which is the reason #385's item 2 stays unbuilt.

With this, #385's actionable items are complete across sessions: the stale claims are corrected (item 1, twice over now), the GREY arm cites the committed corpora rather than v4l2-ctl (item 3, done earlier), Grey16 keeps answering `None` for its own reason (item 4), and item 2 carries its do-not-build analysis on the issue. Closing the issue with this merge.

### DIFF
--- a/crates/irlume-camera/src/lib.rs
+++ b/crates/irlume-camera/src/lib.rs
@@ -213,10 +213,14 @@ pub struct IrCaptureStats {
     /// correction the second draft of this comment needed.
     /// `Quantization::Default` is not itself an effective range: V4L2 resolves
     /// it with `V4L2_MAP_QUANTIZATION_DEFAULT(is_rgb_or_hsv, colsp, ycbcr_enc)`,
-    /// which needs the COLORSPACE, and `IrCamera` keeps only the quantization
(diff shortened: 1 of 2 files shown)


archledger:
Review applied. The second draft repeated the defect class it was correcting: it verified the colorspace had become available and missed that default YUV quantization is full range for XV601/XV709 encodings as well as JPEG colorspace, while the pinned v4l 0.14 `Format` drops `ycbcr_enc` on conversion (verified against the crate source: the `From<v4l2_pix_format>` impl maps colorspace, quantization and transfer, and no ycbcr field exists in the struct). Both comments now carry the encoding fact, and each names the review that forced its draft, so the trail of corrections is part of the documentation.

PR #437, opened 2026-08-12
Record the Brio emitter-cadence measurement (#264)

archledger:
Docs only: the cadence measurement #264's thread named as its next step, run on the archhost Brio today. Result: no cadence-dependent emitter depression at the minutes scale; sustained, post-idle and paced regimes are indistinguishable (table and reading in the file). Together with the thread's prior eliminations (mis-advice path pinned nonexistent, warmup eliminated 8 of 8), this leaves #264 with no actionable mechanism, and the issue closes on this merge with the hours-scale dark-room observation recorded as unexplained.

### DIFF
--- a/docs/research/2026-08-12-brio-emitter-cadence.md
+++ b/docs/research/2026-08-12-brio-emitter-cadence.md
@@ -9,30 +9,41 @@ campaign was designed to chase; the cadence CONTRAST within the campaign is
 the measurement. Lit-phase mean = mean of burst frames above the burst's
 min/max midpoint (18 of 36 frames in every burst).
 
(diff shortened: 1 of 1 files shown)


archledger:
Review applied: the table gains the per-burst paired lit-minus-ambient column the instrument's own documentation prescribes, and the reading is rewritten to what one session supports. The differential itself strengthens the observation the flawed framing gestured at: 35.8 to 36.3 across sustained, post-idle and paced regimes after the first two bursts settled, now correctly isolated from ambient. Equivalence is explicitly not claimed (single post-idle burst), synchronised compensation is named as unexcluded, and the #264 disposition rests on "no actionable mechanism identified" rather than "eliminated".

PR #439, opened 2026-08-12
Commit the per-stage third-party model policy as ADR-0006 (#276)

archledger:
Docs only. Issue #276's implementation merged weeks ago across #277, #278, #279, #280, #293, #294 and #295, but its normative content lived on the issue thread, and `THIRD-PARTY-MODELS.md` cited "recorded on issue #276" for the recognition threshold protocol. This commits the policy's canonical text as ADR-0006: the pin-is-the-safety-property constraint, the per-stage verdicts with their reasons (PAD open because deny-only; recognition open under the four-clause split-source protocol; detection closed for the measured grant-capable-rescue reason; landmarks closed with nothing to open for), and the revisit conditions. The catalog page now cites the ADR.

With this merged, #276 closes: the question it asked is answered per stage with committed evidence, and the two remaining measurement campaigns (the detection rescue corpus; the RGB PAD candidate's live attack evaluation) become their own issues so the umbrella does not stay open as a proxy for them.

### DIFF
--- a/docs/THIRD-PARTY-MODELS.md
+++ b/docs/THIRD-PARTY-MODELS.md
@@ -26,7 +26,8 @@ stage can be installed or wired. Two stages are open:
 - **Liveness (PAD)**, opened first because its wiring is deny-only: a bad
   model there can cost retries or the password, never grant.
 - **Recognition**, opened 2026-08-05 under the split-source threshold
-  protocol recorded on issue #276: the false-accept side measured on public
+  protocol of [ADR-0006](adr/0006-third-party-model-stage-policy.md), first
+  recorded on issue #276: the false-accept side measured on public
   datasets replayed through irlume's own pipeline, the false-reject side on
   this project's cameras, the threshold set from where those populations were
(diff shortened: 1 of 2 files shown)


archledger:
Review applied, both findings adopted wholesale:

- The pin bullet now states what the digest covers and what it does not (threshold, preprocessing, gating and failure behaviour are independent policy parts; a catalog-only threshold edit under an unchanged digest is named as the scenario), and the per-stage failure difference is stated as deliberate: recognition refuses startup, PAD disables the optional deny-only cue and continues, verified against the daemon's loader.
- Protocol clause 3 now describes what ships: the genuine floor is a pre-catalog measurement whose record names subjects, cameras, conditions and margin; enrollment tags templates with the digest and validates no per-user floor; per-user enrollment validation is marked as a prerequisite for any future amendment promising it.

Both first-draft sentences were the same defect class the reviewer has caught across this whole day: prose asserting more than the code does. The ADR now records the catch inline so the next reader inherits the distinction.

Issue #441, opened 2026-08-12
Decide the flrgb RGB PAD candidate with a live attack session


archledger:
Split from #276, which closes with its per-stage policy committed as ADR-0006; this issue carries the live evaluation that decides the RGB PAD candidate.

The candidate is DAMO `cv_manual_face-liveness_flrgb` (MIT model card, 1.36 MB), covering the RGB domain where `flir` is blind. The offline first contact (332 scored rows against the stored corpora; harness, per-frame CSV and model in the research store, model sha `e13b5543...`) established three things:

- The raw output is already a softmax pair and ModelScope's own pipeline softmaxes it AGAIN, squashing scores into [0.269, 0.731]; publisher numbers deserve suspicion (the anti-spoof-mn3 double-softmax trap, found a second time).
- A preprocessing fork exists (the card's 96/112 expansion against the config's 16/112 route) and the two FLIP the failure surface: pad96 fails low-light genuine frames, pad16 fails far and small faces. The fork must be settled by measurement before any entry ships.
- The decisive attack question is unanswerable offline: the stored vinyl-print bursts are IR-only, and flrgb on greyscale calls nearly all IR frames fake (genuine IR control median 0.948), so its 15/15 attack catches there are a domain artifact, not evidence.

What remains is a live RGB session: genuine captures and the vinyl print presented to the RGB camera, both preprocessing variants scored, deciding the variant, the threshold, and whether a deny-only catalog entry ships at all.

Needs: a READY camera session with the enrolled user and the print rig. Policy context: `docs/adr/0006-third-party-model-stage-policy.md`.

archledger:
Closing: the live session ran today and the answer is a decline, not a threshold.

Record: [`docs/research/2026-08-12-flrgb-live-first-contact.md`](https://github.com/archledger/irlume/blob/main/docs/research/2026-08-12-flrgb-live-first-contact.md), with per-frame scores beside it and the scorer committed to `benchmarks/pad-candidates/` with the model, detector and harness hashes pinned (re-running it regenerates the CSV byte for byte).

- **The preprocessing fork is settled.** pad16's genuine and attack populations overlap completely; pad96 is the only variant that separates anything.
[...]

PR #442, opened 2026-08-12
Record the flrgb live first contact (#441)

archledger:
Docs only: the live session #441 named as its remaining work, run today with the enrolled user and the vinyl print. Full write-up and biometric-free per-frame scores committed.

Headlines: the preprocessing fork is settled (pad96; pad16's populations overlap completely and cannot carry a threshold), and the attack surface has a shape nobody wanted: flrgb scores the flat login-distance presentation lowest and the never-chosen tilted presentations highest, so a per-frame separator exists on this corpus (0.079 against 0.179) but with margin too thin to ship, and the low-light genuine regime that decides the false-deny cost is untested. #441 stays open for the low-light genuine leg and a second attack medium; this merge gives it the committed baseline.


archledger:
Second leg added mid-review, and it settles the question: the low-light genuine session (closet, door ajar, frame luma 16-18, no auto-exposure hunting) puts the detected genuine frames at p_fake 0.166 to 0.977, overlapping the flat head-on attack's 0.179 to 0.867 completely. No threshold passes the low-light user and denies the login-shaped attack; the disposition is now a DECLINE of the flrgb candidate rather than a request for more sessions, with the reopen condition stated (a candidate that fixes the low-light collapse, evaluated against this same corpus shape). The per-frame CSV covers all three conditions.

archledger:
All three findings addressed.

**Finding 2 (reproducibility), the substantive one.** Both scripts now live in `benchmarks/pad-candidates/` beside the existing flir and mn3 harnesses; the scorer's header pins the full sha256 of the flrgb model, the YuNet detector and the harness; and the weights resolve through `FLRGB_MODEL` / `IRLUME_YUNET` instead of a home-directory path. Verified the way the reviewer asked: deleted the CSV, re-ran from the committed scripts, and it regenerates byte for byte.

Worth recording that my first attempt at that verification was a false pass. I ran the replay with stderr discarded, saw the untouched CSV still matching, and nearly reported success; the committed script could not even load its model. Deleting the file first is what exposed it, and it is why the claim above is stated as an observation rather than an inference.

**Finding 3 (attacker-behaviour claims).** The document no longer asserts what attackers universally do. It says the square-on presentation is this project's baseline print attack (#235, #237) and the easiest for a camera-facing attacker to hold, which is what makes the score ordering unfavourable.

**Finding 1 (the 0/20 arithmetic).** Already gone: that sentence died in the low-light rewrite pushed before this review landed, and the committed threshold table independently reports 7 of 20 denied at 0.5, matching the reviewer's count from the CSV. The stale buffalo-0.5 framing went with it; the table now presents thresholds as reference points with no cross-model borrowing.

Issue #440, opened 2026-08-12
Open the detection stage on an end-to-end rescue corpus, or refuse it on one


archledger:
Split from #276, which closes with its per-stage policy committed as ADR-0006; this issue carries the one measurement that decides the detection stage.

The candidate is fully built and waiting: full-range BlazeFace runs its `.tflite` unconverted on the bundled runtime (#295), the decoder holds 0.9354 mean IoU parity against Google's runtime, and the operating point is measured at 0.55 (an empty room reaches 0.5293 on near-black IR, so the rescue slot's 0.5 would admit false detections). The stage stays closed because the corpus that would justify opening it does not exist.

What the corpus must be, per the recorded reasoning: the rescue slot is grant-capable, so availability and hallucination corpora measure the wrong thing. The measurement is end-to-end on YuNet-miss frames: presentations of a print of the enrolled face, screens, other faces, and genuine users, on the frames where the rescue actually fires, followed through alignment, embedding and matching to the authentication OUTCOME at 0.55. The vinyl print from #235/#237 is the instrument. The result decides both whether the detection stage opens and whether any catalog entry may carry this model at all.

Needs: a READY camera session with the enrolled user and the print rig. The per-stage policy and its revisit conditions: `docs/adr/0006-third-party-model-stage-policy.md`.

PR #444, opened 2026-08-12
Answer the InsightFace question per stage, with a measurement

archledger:
Answers the InsightFace-stack question for all three stages. One stage needed new evidence; the other two needed the truth said where users will find it.

**Recognition: already supported.** The `buffalo` catalog entry is InsightFace ArcFace R50 (`w600k_r50.onnx`, same official pack, hash verified identical). `THIRD-PARTY-MODELS.md` now states that plainly, with the measured caution that a stronger recognizer is not a stronger defence here: through irlume's own pipeline buffalo_l beat AuraFace marginally on LFW EER (3.9% vs 4.2%), was worse on one FairFace group (5.63e-4 vs 4.50e-4), and tied on the all-group threshold (0.700 vs 0.685), while the attack that actually defeats irlume is a print the recognizer is supposed to match.

**Detection: SCRFD-10G measured.** New offline bench against the shipped YuNet on the 512-frame stage-3 corpus, manifest-verified ([record](https://github.com/archledger/irlume/blob/main/docs/pad-results/2026-08-12-scrfd-vs-yunet.md), per-frame CSV, harness in `benchmarks/pad-candidates/`). Availability is a wash: median IoU 0.904 where both fire, within one frame of each other at a matched 0.6 threshold, neither firing on 128 empty-room frames. The failure modes differ: SCRFD recovers four saturated lit-strobe IR frames YuNet loses, and at its own default 0.5 invents four background boxes in a dark pose sweep. Not enableable, same reason as full-range BlazeFace: the rescue slot feeds the grant path and no corpus covers prints or screens on rescue-firing frames (#440).

**Landmarks: rejected on the model's own output shape.** `2d106det.onnx` emits a single `fc1[1,212]` tensor: 106 points, two coords, no iris, no depth, against the 478-point FaceMesh V2 with iris that EAR closure, head pose and the consent gestures all index. That is a re-derivation of every cue and a re-measurement of every gate, not a model swap.

**No catalog entries added, deliberately.** `catalog_entries_are_well_formed` refuses a closed-stage entry, and its reasoning holds: listing a model nobody can enable is documentation pretending to be a catalog. The struct's own doc comment claimed the catalog *was* the place for such entries; that was stale and now describes the enforced rule, naming the doc section where measured-but-gated candidates live.

Verification: the bench's YuNet side is a port of irlume's own detector validated to zero score delta against the committed bench CSVs on all 320 overlapping rows; the committed harness, run after deleting the output, reproduces the committed per-frame CSV byte for byte. Full local gate green (fmt, MSRV clippy, doc, 1389 tests).

Not tested: latency of either detector, anything downstream of the box, and any camera or subject outside this corpus. No attack presentations are in this corpus at all, which is exactly why nothing here argues for opening the stage.


archledger:
All four findings applied, each verified against the code before accepting it.

**Finding 1 (high), the landmark scope.** Confirmed: `head_pose` takes `&Landmarks5`, every authentication caller passes `f.landmarks`, and the EAR indices are all below 468. So the consent gestures and head pose are untouched by a mesh swap and the iris points are not implicated. The page now scopes the migration to an eye mapping plus a re-measured closure gate, and rests the closure on the reason that survives: nothing has measured what this model's landmarks do to the cues consuming them.

**Finding 3, "matched 0.6".** Reframed as a reference point, with the circularity named: the same corpus that shows the background boxes falling below 0.6 is the only evidence they do, and the two score scales share no calibration.

**Finding 2, the harness verifying nothing.** This was the one worth the most. It printed digests and then used the files anyway, so a swapped model or one replaced frame would have produced a CSV indistinguishable from a reproduction. It now refuses before any inference unless both artifacts match the published digests and the corpus enumerates exactly 512 frames, with `--allow-unpinned` as the deliberate override, and the default output is the committed record instead of a machine-specific scratch path. Both guards were proved to fire rather than assumed: a wrong model, an empty corpus, and a one-frame corpus each produce their own refusal.

**Finding 4, CRLF.** The writer emits LF now; the committed CSV was deleted and regenerated by the fixed harness, and `git diff --check` is clean.

Issue #392, opened 2026-08-09
The nightly emitter check cannot take the emitter lock as the runner account, so it now fails


archledger:
#387 made the nightly emitter check honest, and it is now failing on archhost for a real reason that needs a machine-configuration decision.

## What happens

```
IR node: /dev/video2
::error::passwordless sudo is required to drive the emitter here (#384)
```

Run `31295697469`, on `d9af314`.

## Why

irlume guards its emitter bookkeeping with `/run/lock/irlume-emitter-<key>.lock`. The daemon creates it under `UMask=0027`, so it lands `0640 root:root`. An unprivileged process cannot open it, irlume declines to write the emitter, and the check that asserts irlume drove the emitter cannot run.

The job runs as `ghrunner`, the runner service account, which has no passwordless sudo. I verified sudo over SSH as `archledger` and assumed it applied to the job; it does not. That mistake is mine and is why this landed red rather than being caught before merge.

Failing is the correct outcome. Before #387 this step passed while irlume never drove the emitter, which is what #384 was about. A check that cannot run has not passed.

## The decision

**A. Give the runner account passwordless sudo for the burst.** Contained to one machine, no code change. The cost is that the runner's work directory is writable by that account, so a NOPASSWD rule naming a path inside it is a privilege-escalation path for anything that can write there. A fixed root-owned wrapper avoids that but has to resolve the built binary without taking an arbitrary path as an argument.

**B. Align the lock with the capability boundary.** `/dev/video*` is `root:video 0660`, so the group that can already drive the emitter is `video`. The lock guarding that capability is root-only, which is the mismatch. Creating it `0660 root:video` would let exactly those callers take it, with no root grant anywhere.

I argued against B earlier in #384 on the grounds that widening a lock on an authentication daemon to solve a CI problem is the wrong trade. I now think that framing was wrong: this is not widening it to suit CI, it is matching it to the capability it guards. A local user in `video` can already open the camera and interfere with it; what they gain is the ability to hold this lock and stop irlume managing the emitter, which is a denial of service against a capability they already hold.

**C. Leave it red.** The nightly is schedule-only and not a required check, so main is unaffected. But a permanently red lane is how the eight-night outage in #383 went unnoticed, and repeating that would be a poor lesson to draw from it.

## Recommendation

B, implemented deliberately and not at the end of a long session. It removes the need for any root grant, it is one change in `lock_camera`, and it makes the non-root CLI tooling work for the same reason.

[...]

Issue #346, opened 2026-08-07
Auth latency is capture-bound: three measured wins (session reuse across retries, one enrollment load, no double model hash)


archledger:
From a performance audit (2026-08-07, commit 92c3725, live daemon + journal + a local ort bench; no camera opened). The headline: auth latency is capture-bound, not inference-bound. One denoised RGB capture logs 1064-1176ms in the journal, and the code's own note (crates/irlume-camera/src/lib.rs:1286-1292) attributes ~700ms of a 1.11s one-shot to open, negotiate and AE warmup. Inference is milliseconds by comparison (mesh 4.5ms, blaze under 1.4ms, per docs/pad-results/2026-08-07-mp-latency-archhost.md).

Three changes worth making, each named with the measurement it needs FIRST, because nothing here changes without evidence:

1. Hold the RGB and IR sessions across the grace window's retries. `authenticate_for` (crates/irlume-auth/src/lib.rs:2417-2437) loops `authenticate_once` on the one-shot capture path, so every retry re-pays the ~700ms setup that `RgbCamera` exists to amortize. Measure first: per-attempt latency from the journal's `assess` lines plus session_bench on the Zenbook and the NexiGo box, before and after, and confirm the capture LED behavior stays acceptable while a session is held.

2. Load the enrollment once per `authenticate_for` (or overlap the template unseal with the first capture). Today every attempt re-runs storage::load, template_key::load_key and tpm::unseal (auth/lib.rs:2457, template_key.rs:92), plus one more load before the gesture watch (auth/lib.rs:2413). The project's own #249 numbers put one unseal at 2.70s quiet and 18.97s contended on a discrete TPM, so on dTPM hardware this can rival capture per retry, and TPM and camera are independent devices, so the unseal can leave the critical path entirely. Measure first: unseal wall time on the Zenbook fTPM and the archhost dTPM with tracing on, then a security pass on how long a decrypted key may live in the worker.

3. Feed the bytes `verify_models` already read and hashed into `Engine::load`. Startup reads and sha256s the 260MB recognizer twice (daemon main.rs:38-53 and :261, then auth/lib.rs:809-816, 833-836); measured waste is 165ms read plus 181ms hash warm, and a second full 260MB read cold. `load_with_recognizer_bytes` already exists and the third-party path uses it. Measure first: cold exec-to-serving from journal timestamps with caches dropped, before and after.

[...]

PR #445, opened 2026-08-13
Release-prep: fix the sixteen blockers the pre-release audit found

archledger:
Before cutting the next release, a 12-agent audit read every crate's v0.9.0..HEAD diff, the Repair tab check by check, TUI/CLI parity, the upgrade path, and the release docs. Sixteen blocker-level findings came back. This branch fixes all of them, each demonstrated against the failing state before and after, most pinned by a test whose mutant is caught.

The three worst, all verified live:

- `irlume login enable` on a packaged install with a dead daemon removed face auth from every greeter and the lock screen while reporting success. Socket activation keeps `/run/irlume.sock` present, so the capability query times out; a timeout is not proof the daemon is absent, but `caps()` guessed `{ir_pair: false, rgb: false}` and the wiring path unwired what read as unsupported. Reproduced on the released binary (`plan → face login: off  face lock: off`, exit 0). An enable now refuses an unestablished reading; a disable never needed one.
- The #346 held-session rework (unreleased) dropped session references at its seven release points while the owning bindings kept the camera's buffer queue, so the consent watch EBUSYed against its own process and a successful face match became a password prompt. The sessions are now owned Options released where the decision is made; proven on real hardware by holding, colliding, releasing, and reopening.
- `IrSession::recover` ran the fresh emitter enable while the old guard still held the per-camera stream lock; flock excludes per open file description, so the enable refused against its predecessor, and recovery reported Ok while later captures returned dark IR frames. A hardware test now holds a driven session, recovers, and asserts the emitter stays lit; with the old ordering restored it fails and prints the predicted "not driving unit14/sel6" refusal.

The rest, by area:

- Repair tab: the TFLite runtime gets a row beside ONNX (the mesh runs on it since #295 and a packaged daemon refuses to start without it; its absence had no surface anywhere), the third-party checksum check is no longer gated off exactly when the refusal kills the daemon, a starting daemon renders as starting instead of dead with a restart offer that reopened the window, EACCES renders as a permission problem, the Cameras active line distinguishes unknown from no-hardware, and the eyes-open row stops advertising a toggle the daemon refuses (#386).
- The SELinux fix now performs the relabel it reported: one shared semodule/try-restart/restorecon sequence, and the `.pp` lookup no longer prefers a file under the caller's working directory over the packaged one.
- `status` and `doctor` read the keyring-release gate the way the daemon applies it (per-service key first); doctor no longer asserts a gate the daemon does not apply, nor tells a user to nod while its own MISCONFIGURED warning two lines up says no gesture is accepted.
- `credential-release-challenge <service> status` works; four surfaces taught it while the parser exited 2. A dash-prefixed token is a usage error instead of a settings key named `service_gesture.--yes`.
- `profiles list --json` keeps `require_challenge` frozen at `false`: it was required in the published contract-1 schema, and dropping it broke every consumer validating against the pinned file.
- Daemon: a refused fd inheritance no longer reads as socket activation (the self-bound socket then skipped its chmod and every non-root client got EACCES), a panicking connection thread frees its slot, the post-panic engine rebuild feeds the watchdog, and a post-close arbiter submit refuses instead of hanging its caller for 300 seconds.
- `lock_camera` chmods the lock only when the mode is wrong and the caller can: the unconditional fchmod on the root-owned lock returned EPERM for exactly the video-group caller #392 exists for, and the emitter went dark.
- Docs: THREAT_MODEL stops offering the retired blink gate as a mitigation under its own confirmed-breach heading; ARCHITECTURE counts the five bundled models and drops the blink gate from its table and diagram; LIMITATIONS carries the two live limitations with figures from the committed corpus; the CHANGELOG gains its missing entries, including a Security note for the credential-release default flip from ON to OFF; CITATION.cff stops citing 0.7.0. About ten stale camera-crate comments corrected, including a front page describing the pre-#340 capture default.
- doctor gains a `tflite-runtime` check beside `onnxruntime`, verified in all three states.

Not in this branch: the Repair audit's suggested consent-gesture readiness checks (they need the enrollment fields the TUI currently discards) and the third-party refusal shapes beyond checksum mismatch. Held sessions aside, nothing here changes authentication behavior.

Testing: full workspace suite (1393 tests), fmt/clippy/doc gates, machine-api conformance, live demonstrations recorded per commit, and hardware runs on the Zenbook camera for the two camera fixes. Device walkthroughs and the 0.9.0 upgrade path on archhost, thinkpad and minihost run next and will be reported on this PR.


archledger:
## Fleet upgrade pass and review round

The 0.9.0-to-branch upgrade was exercised on the three test machines, each running packaged 0.9.0 with its daemon active: archhost (Arch, NexiGo IR pair, the self-hosted runner), thinkpad (Ubuntu 26.04, RGB-only convenience tier, one enrolled profile), minihost (CachyOS, IR pair, no enrollment). Method: branch binaries and PAM module installed over the packaged files, then the packaging scriptlet steps (daemon-reload, restart, `login reconcile`) run exactly as the lanes run them. The machines were restored to packaged 0.9.0 afterwards; the migrated PAM state is 0.9.0-compatible and stays.

What the pass proved, and what it caught:

- **The polkit stanza migration works on a real stale install, after two live catches.** thinkpad carried 0.9.0's plain `sufficient` polkit line, and its `login.wired` marker said polkit (and sudo) were never wired, so the stale-stanza detection alone would never have run. Reconcile now adopts wired surfaces the marker missed (the file is the ground truth), which is commit `1812259`. The first migration attempt was then refused by this branch's own unwire guard, because the scriptlet ordering lands reconcile inside the daemon's model-loading window; a wiring enable now waits out a starting daemon, commit `5c4a628`. With both in place the stanza migrated to the `abort=die` control and the marker was corrected.
- **Enrollments, template keys, keyring state and camera config survive.** thinkpad's profile and 10 scans list identically before and after; archhost's pre-existing "no template key sealed" answer is byte-identical on both binaries (test-box debris predating the upgrade, not a regression).
- **The new surfaces answer correctly on all three:** the TFLite runtime row and doctor check pass, `doctor --json` carries `tflite-runtime` and `emitter-stream-pending`, `status --json` reports `running`, the per-service `credential-release-challenge sudo status` form works everywhere (including the honest root-only answer where settings.conf exists), and `profiles list --json` carries the frozen `require_challenge: false`.
- **TUI walkthrough on all three:** Welcome, Repair, Cameras and Settings render; Repair shows the four-state daemon row, both runtime rows, and the correct tier per machine (secure on the two IR pairs, convenience on the RGB-only thinkpad); no panics.
- The daemon on thinkpad started without the mesh model present and reported the degraded state; installing the model (as the package does) brought FaceMesh up on the bundled runtime.

The Codex review round returned three merge-blocking findings, all real, all fixed in commit `8146748` and verified: the capability fallback could still establish absence from a path check on a configured install (the rule is now asymmetric and was demonstrated live in both directions), the relabel helper discarded its command statuses while both callers claimed success (statuses checked, failures propagate, the success path re-verified against a real mislabel), and `lock_camera` proceeded past an unfixable world-accessible mode (flock takes a lock on a read-only open, so that state is now a refusal, tested against a root-owned 0666 lock). The follow-up note is also in: the held-session hardware test now proves both halves of the release.

Not covered in this pass: a face grant on the fleet (nobody is enrolled in front of those cameras except thinkpad's profile, whose owner was not present), and the deb/rpm/PKGBUILD scriptlets themselves ran as their command sequences rather than through a built package.

Issue #462, opened 2026-08-15
UVC exact stream negotiation and frame provenance


archledger:
## Goal

Complete the Phase 2 UVC stream contract from `irlume-camera-architecture-research.md`: exact `(fourcc, size, frame interval)` negotiation and descriptor-bound frame provenance.

## Dependency / stack

- Depends on #455 / PR #458 inventory generations.
- Depends on #456 / PR #460 lifecycle invalidation.
- Depends on the common lease/session issue so provenance cannot be bypassed by legacy direct opens.

## Stream negotiation

- Enumerate complete supported tuples: pixel format, width, height, discrete/stepwise/continuous frame interval.
- Select one declared tuple per logical stream and record requested versus accepted values.
- Re-read the tuple after `S_FMT`, after buffer allocation/claim, and at stream start.
- Measure delivered rate; never infer it from a requested integer FPS.
- Fail closed on tuple drift, unsupported interval semantics, or one paired member changing/disappearing.
- Preserve current selection behavior unless a current path is incomplete or unsafe; no speculative bandwidth-reduction mode.

## Frame provenance

Every delivered frame carries at least:

- `CameraInstanceId` and generation;
- logical stream role;
- validated configuration/profile identity;
- V4L2 sequence;
- timestamp and timestamp-source/clock flags;
- bytes used, frame status/error flags;
- cumulative drop/gap counters;
- illumination evidence as `Unknown | Ambient | ActiveIr` (never infer optical output from a successful host write alone).

Authentication/enrollment policy must reject stale generation, configuration drift, invalid status, timestamp discontinuity, and provenance gaps outside declared profile bounds.

## Acceptance

- Fixtures cover discrete, stepwise, and continuous intervals; post-buffer format sabotage; delivered-rate mismatch; sequence wrap/gap; timestamp-source change; stale generation; short/corrupt frames; and paired-stream member loss.
- Machine-readable diagnostics report requested/accepted/delivered tuples, sequence gaps, timestamp-source flags, drops, and RGB/IR skew.
- A 60-second real dual-stream run records numeric requested/delivered rates and provenance counters.
- Existing UVC capture/emitter behavior and auth decisions remain unchanged except explicit fail-closed rejection of unproven/stale frames.

## Non-goals

- No libcamera or PipeWire backend.
- No MIPI authentication enablement.
- No unmeasured bandwidth workaround.

archledger:
[...]

PR #470, opened 2026-08-16
feat(camera): enforce delivered rate and expose diagnostics (#462)

archledger:
## Summary

Refs #462 — final slice.

Enforces an exact delivered-rate floor on every production capture stream and
exposes machine-readable diagnostics for it, so a camera that advertises a
Windows Hello interval but delivers too slowly is caught and reported, not
assumed adequate.

## What it does

- **Exact 30-delta delivered-rate gate** (`rate_gate::RateWindow`) on
  `TrackedStream::next` — the single dequeue choke point for RGB/IR sessions,
  `capture_raw_burst_timed`, `capture_ir_streaming`, `capture_ir_sequence`, and
  recovery. All integer arithmetic (no `f32`/`f64`). Floors: IR 15/1, RGB 15/2;
  98% tolerance, derived from measured dual-stream hardware (14.714 Hz).
- **Independent rate baseline**: a corrupt dequeue does not advance the rate
  window (the next valid delta spans it, fail-closed). Recovery atomically
  resets the window with the two continuity trackers.
- **Typed failure**: `Error::DeliveredRate(Box<CameraStreamRateEvidence>)`
  carries the exact measurement; the message is a fixed stable string.
- **Per-frame evidence**: immutable `DeliveredRateEvidence` (requested/accepted/
  floor/delivered rationals, gaps, drops, clock/source, epoch) attached to every
  frame's runtime provenance.
- **Machine diagnostics**: `irlume camera diagnostics --json` (contract-1
  additive capability `camera-diagnostics`) → `Request/Response::CameraDiagnostics`
  (boxed report), daemon camera-class handler runs the ordinary gated capture per
  present role and reports requested/accepted/delivered, gaps/drops, clock/source,
  and same-domain RGB/IR skew. No device paths, account identity, or template data.

## Validation

- 1,598 guarded workspace tests green; strict clippy/fmt/docs/release/diff clean.
- 4 compile-valid rate-gate mutants killed and restored (inverted comparison,
  tolerance off-by-one, wrong IR floor, no-op reset).
- Measured basis: single-stream and concurrent dual-stream kernel-timestamp probes
  (`target/rate-probe-evidence/…/analysis.json`); 30-delta window is the smallest
  round window above the measured 25 and clears the 98% floor.

## Scope / non-goals

- No change to interval selection, `S_FMT`/`S_PARM` policy, emitter writes, or
  capture-mode selection. Diagnostics are read-only captures; no device writes.
- `camera diagnostics` opens the camera through the same lease/arbiter as other
  camera-class requests.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -27,7 +27,7 @@ pub use irlume_camera::{capabilities, device_identity, select_pair};
 /// Enumerate the Hello camera pairs. Re-exported for the daemon's
 /// camera-class `ListCameras` arm: clients must not enumerate for themselves
 /// (#187), so this is the only path to a listing.
-pub use irlume_camera::{list_pairs, privacy_engaged, CameraPair};
+pub use irlume_camera::{camera_rate_diagnostics, list_pairs, privacy_engaged, CameraPair};
 
 /// Loaded models + camera device selection. Build once, reuse per request.
 pub struct Engine {
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -3147,10 +3147,20 @@ impl Engine {
                 // SAFETY: _cams is Some, and _rs/_is borrow from it. _cams is
                 // declared before _rs/_is so it outlives them.
                 let (ref cam_r, ref cam_i) = _cams.as_ref().unwrap();
-                if let (Ok(rs), Ok(is)) = (
+                if let (Ok(mut rs), Ok(mut is)) = (
                     cam_r.session_with_progress(&progress),
                     cam_i.session_with_progress(&progress),
                 ) {
+                    // Establish the delivered-rate windows for the HELD PAIR up
+                    // front, draining both streams concurrently. Best-effort: a
+                    // failure is logged, and the per-frame serial fill in
+                    // `next()` still re-attempts (and fails closed) on capture.
+                    if let Err(error) = irlume_camera::establish_pair_rate(&mut rs, &mut is) {
+                        irlume_common::dlog!(
+                            "auth: held pair could not establish delivered-rate evidence \
+                             ({error}); the per-frame fill will retry"
+                        );
+                    }
                     held_rgb = Some(rs);
                     held_ir = Some(is);
                 }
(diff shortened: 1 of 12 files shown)


archledger:
## Slice-8 enforcement fix (delivered-rate fill) — final

The serial fill is fixed with a **concurrent** pair-level warm-up (not round-robin):

- `establish_concurrent_rate` spawns one thread per stream; `drain_until_both_ready` fills each rate window then keeps discarding (the window just slides) until BOTH are ready, so the faster stream never idles and overflows its queue while the slower twin finishes.
- Wired into the 4 held-session sites (auth, enrolment, A/B/A check, capture-mode probe) and the physical stress test, which now captures RGB+IR on concurrent threads (matching production) and establishes the rate up front for both dual and RGB-only.
- `install_recovered` now resets the rate window, so a post-recovery fill can't skip re-establishment on a stale "ready" window.
- Stress-test validator updated: discarded-observation check is now a documented minimum (the fill's trailing drain is rate-ratio dependent), and the recovery duration now spans the re-establishment so it matches the recorded timestamp gap.

**Round-robin was tried first and rejected by measurement**: it throttles the faster stream (ASUS RGB 30 fps vs IR 15 fps), overflowing RGB's queue and dropping IR frames (IR measured 14.5 Hz vs the 14.7 Hz floor). The 98 % tolerance was calibrated against a *concurrent* probe (14.714 Hz), so the fill must be concurrent.

**Gates**: fmt, clippy -D warnings, doc -D warnings, release build, guarded workspace tests (1603) — all green. Independent review: SHIP-WITH-NITS (nits addressed).

**Hardware matrix** (60 s stress + validator):
| host | mode | result |
|---|---|---|
| ASUS UX5406S-Fedora (local) | dual | PASS |
| HP-ArchLinuxGaming (archhost) | BRIO rgb-only | PASS |
| minihost (NexiGo N930W) | dual | PASS at `8581e5a` (dual path identical); exact-head re-run aborts on `VIDIOC_DQBUF` — NexiGo still USB-flapping (dmesg `usb 1-3.3 USB disconnect`, `Failed to resubmit video URB`), environmental, needs a physical reseat |
| thinkpad | rgb-only | offline (needs power-on) |

Issue #472, opened 2026-08-16
A GNOME token arm is unrecoverable by design: an unattended fwupd dbx update permanently destroyed my login keyring (follow-up to #250)


Magniquick:
# A GNOME token arm is unrecoverable by design: an unattended fwupd dbx update permanently destroyed my login keyring (follow-up to #250)

First, thanks for irlume — and specifically for #250. That survey is why I could reconstruct this failure completely instead of guessing. This report isn't asking you to revert #250; the root-escalation argument against sealing the Unix password is correct and I'm not disputing it. It's about a property of the *particular* alternative that was chosen, which I don't think was weighed at the time: **it is the only one of the three that cannot be recovered when the seal is lost.**

## What happened

- **Jul 18** — `irlume keyring arm` on a GNOME host. Resolved to `GnomeKeyringToken`, re-keying my login keyring to a random 64-byte token, sealed Tier 3 literal to PCR 7 (Secure Boot off, pcrlock not provisioned).
- **Aug 2, 02:33** — `fwupd` applied a UEFI **dbx** revocation update (`fwupdmgr get-history`: previous version `20160809`, `Update State: Success`, "needs a reboot after installation"). I did not initiate this and was never prompted about consequences.
- **Secure Boot was never enabled on this machine** (`SecureBoot`=0, `SetupMode`=0). dbx is still one of the five variables measured into PCR 7, so PCR 7 changed on the next boot anyway. I flag this because "PCR 7 = Secure Boot" intuition suggests an SB-disabled machine is safe from PCR-7 drift. It isn't.
- Every PCR-7-bound seal died at once — face templates *and* the keyring token. So the face factor that would have released the token was broken by the same event that broke the token.
- **Aug 15, 19:37** — irlumed logged, interleaved with `Esys_Unseal ... ErrorCode (0x0000099d)`:

```
irlumed: SealPassword: sealed a GNOME keyring token for 'magni'
         (reused from the existing envelope); caller must now re-key the login keyring
```

The caller could not complete that re-key — face auth was dead from the same drift. End state: the login keyring is opened by neither my password nor the envelope's token. Contents lost.

I verified that three independent ways before accepting data loss:

1. `gnome-keyring-daemon --unlock` with the correct login password — stayed locked.
[...]

Issue #448, opened 2026-08-14
[bug] keep getting a failure to decrypt when trying to enroll or even run the tui setup anytime i try


orthadoxtesla:
### irlume version and install channel

0.10.0

### Distro and desktop

Arch Hyprland

### What happened, and what you expected

On multiple machines with fresh installs, including one with a completely fresh installation of arch, setup is failing at the enrolling step. there seems to be some issue with the TPM decryption or something like it. and while it says that 

### Relevant log lines

```text
irlume status
irlume status for 'user'
  daemon        : running ✅
  auth method   : Auto
  enrollment    : error: policy: decryption failed (wrong key or tampered data)
  keyring unlock: armed ✅, literal PolicyPCR (Tier 3)
  templates     : encrypted at rest ✅
  recovery pass : set ✅
  biopolicy     : off (default)
  keyring gate  : off (default): the keyring releases after the face match with no nod
  cameras       : rgb=/dev/video0 ir=none
  fingerprint   : none




irlume doctor
[doctor] platform: Arch-family
[doctor] install origin: pacman package (AUR / makepkg)
[doctor] TPM 2.0: /dev/tpmrm0 ✓
[doctor] Secure Boot: disabled ⚠ (TPM PCR-7 binding is weak; enable for trust)
[doctor] boot mode: UKI (signed PCR-11 eligible)
[doctor] capture mode: root-only setting (reads /etc/irlume/cameras.conf); re-run `sudo irlume doctor` to read it
[doctor] signed PCR policy: none (no Tier 1 on this boot chain)
[doctor] pcrlock: not provisioned: seals use the literal PCR-7 policy + recovery passphrase (re-arm/restore after firmware updates); `systemd-pcrlock make-policy` enables Tier 2
[doctor] camera nodes (classified by pixel format):
  /dev/video0: Rgb (uvcvideo, USB)
[doctor] IR stream: no selected node
[doctor] RGB stream: 640x480@30.00fps YUYV ✓ (Windows Hello minimum 480x480@7.5fps)
[doctor] models:
  loaded by the daemon ✓
[doctor] ORT_DYLIB_PATH: (unset)
[doctor] ONNX Runtime: system libonnxruntime.so (1.28.0, mapped from /usr/lib/libonnxruntime.so.1.28.0) ✓
[doctor] TFLite runtime: /usr/share/irlume/tflite/libtensorflowlite_c.so ✓
[doctor] pipeline stages (candidate per this shell's search order):
[...]

PR #486, opened 2026-08-16
fix(tpm): salt policy sessions so unseal returns the sealed secret

Magniquick:
Fixes the root cause behind #472 (and, I believe, #448): on this machine every TPM unseal returned plausible-looking garbage, so enrollment was encrypted with a key the envelope could never yield.

## The bug

`with_session()` turns on ESYS parameter encryption:

```rust
let session = ctx.start_auth_session(
    None,   // tpm_key  (salt)
    None,   // bind
    None, kind, SymmetricDefinition::AES_128_CFB, HashingAlgorithm::Sha256)?;
let (attrs, mask) = SessionAttributesBuilder::new()
    .with_decrypt(true)
    .with_encrypt(true)
    .build();
```

The session is started **unsalted and unbound**, so its session key is empty. ESYS still applies response-parameter decryption to the value coming back from `TPM2_Unseal`, using a key the TPM never encrypted with. The result is 32 well-formed, wrong, different-every-run bytes.

## Why it looked like a storage bug

The sealed object is perfectly fine. Unsealing the *same* envelope with tpm2-tools, under the same SRK (`0x81010002`) and the same PCR-7 policy, returns the exact secret:

```
sealed         : 000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f
tpm2_unseal    : 000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f   <- correct
irlume unseal(): e10425f3ec9623f02b4516a514405c31e6fcbb87e5baa1c060b65429daaa7452   <- garbage
```

Only the ESYS response path is affected, and only where a real secret is returned. `seal()` is untouched because `create` runs under `execute_with_nullauth_session`. So the failure surfaces one layer later, as `decryption failed (wrong key or tampered data)` on read-back, or — after #450 — as `new template key failed persisted TPM round-trip`.

Worth noting: #450's round-trip guard is doing exactly its job here. It correctly refuses to write an enrollment it cannot read back, which turned silent corruption into a clean refusal. It just couldn't fix the cause, so on this host enrollment cannot complete at all.

## The fix

Salt the three policy sessions against the SRK, so both sides derive the same session key. Parameter encryption then does what it was written to do — protect the unsealed secret on the bus — instead of corrupting it. Trial sessions pass `None` and turn the attributes off: they carry no secret and need no salt.

I chose salting over simply dropping `with_encrypt`/`with_decrypt` because the attributes are clearly deliberate, and an unsealed login password or template key is exactly the thing worth encrypting on the bus. Dropping them also fixes the corruption, if you'd prefer the smaller change.

## Verification

Your own `tpm::tests::seal_unseal_roundtrip_real_tpm`, on real hardware (Lenovo 83KF fTPM, tpm2-tss 4.2.0, `/dev/tpmrm0`, run as root):

**Before**
```
test tpm::tests::seal_unseal_roundtrip_real_tpm ... FAILED
assertion `left == right` failed: round-trip must match
  left:  [112, 213, 59, 99, 97, 212, 227, 195, ...]
  right: [105, 114, 108, 117, 109, 101, 45, 107, ...]   // b"irlume-keyring-secret-roundtrip!"
```

**After**
```
test tpm::tests::seal_unseal_roundtrip_real_tpm ... ok
```

Full suite after the change: `124 passed; 0 failed; 26 ignored`.

That test already exists and is `#[ignore]`-gated with "CI does this" via swtpm, so it's worth checking whether swtpm tolerates parameter encryption on an unsalted session where this fTPM does not — that would explain how this reached a release.

A pleasant side effect: with unseal working, `seal()`'s tier ladder stops falling back. Before, every attempt logged `pcrlock seal round-trip mismatch; falling back to literal PCR seal`, because the round-trip check was being defeated by this same corruption. After, a pcrlock host seals at Tier 2 as intended:

```
policy : PcrlockNv { nv_index: 27697224 }   pcrs [15]
MATCH  : yes
```

So this may also be why Tier 1/Tier 2 seals appeared unusable on hosts that had them provisioned.

Happy to split the trial-session part out, add a dedicated regression test, or adjust anything to fit your conventions.

### DIFF
--- a/crates/irlume-core/src/tpm.rs
+++ b/crates/irlume-core/src/tpm.rs
@@ -178,12 +178,23 @@ fn read_pcr_values(ctx: &mut Context, pcrs: &[u32]) -> Result<Vec<PcrValue>> {
 /// regardless of whether `body` succeeded.
 fn with_session<T>(
     ctx: &mut Context,
+    salt: Option<KeyHandle>,
     kind: SessionType,
     body: impl FnOnce(&mut Context, AuthSession) -> Result<T>,
 ) -> Result<T> {
+    // `salt` is load-bearing, not decoration. ESYS parameter encryption derives
+    // its key from the session key, and an UNSALTED, UNBOUND session has an
+    // EMPTY session key: the TPM then returns the response parameter in the
+    // clear while ESYS still "decrypts" it, so `unseal` hands back plausible
+    // garbage instead of the sealed secret. The sealed object is fine; only the
+    // response is mangled, which is why a tpm2-tools unseal of the very same
+    // object succeeds. Salting against the SRK gives both sides the same key,
+    // so the secret is genuinely protected on the bus AND comes back intact.
+    // Trial sessions pass `None`: they carry no secret, and turning encryption
+    // off there keeps them correct without a salt.
     let session = ctx
         .start_auth_session(
-            None,
+            salt,
             None,
             None,
             kind,
@@ -192,9 +203,10 @@ fn with_session<T>(
         )
         .map_err(tpm_err)?
         .ok_or_else(|| Error::Tpm("start_auth_session returned None".into()))?;
+    let encrypted = salt.is_some();
     let (attrs, mask) = SessionAttributesBuilder::new()
-        .with_decrypt(true)
-        .with_encrypt(true)
+        .with_decrypt(encrypted)
+        .with_encrypt(encrypted)
         .build();
     if let Err(e) = ctx.tr_sess_set_attributes(session, attrs, mask) {
         let _ = ctx.flush_context(SessionHandle::from(session).into());
@@ -209,7 +221,7 @@ fn with_session<T>(
 /// Compute the PolicyPCR digest for a given PCR selection using a trial session.
 /// The digest is what the sealed object commits to.
 fn compute_policy_digest(ctx: &mut Context, pcrs: Option<&PcrSelectionList>) -> Result<Digest> {
-    with_session(ctx, SessionType::Trial, |ctx, session| {
+    with_session(ctx, None, SessionType::Trial, |ctx, session| {
         if let Some(sel) = pcrs {
             let policy = PolicySession::try_from(session).map_err(tpm_err)?;
             ctx.policy_pcr(policy, Digest::default(), sel.clone())
@@ -624,7 +636,7 @@ fn unseal_literal(env: &SealedEnvelope) -> Result<Zeroizing<Vec<u8>>> {
 
         // Scoped fallible region so the loaded object is flushed on every exit.
         let result: Result<Zeroizing<Vec<u8>>> = (|| {
-            with_session(ctx, SessionType::Policy, |ctx, session| {
+            with_session(ctx, Some(*srk), SessionType::Policy, |ctx, session| {
                 if !env.pcrs.is_empty() {
                     let sel = pcr_selection(&env.pcrs)?;
                     let policy = PolicySession::try_from(session).map_err(tpm_err)?;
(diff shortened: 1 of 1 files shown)


archledger:
## Verification on the irlume test fleet

Independent verification on three of our TPMs plus the camera regression matrix.

### TPM seal/unseal roundtrip (`seal_unseal_roundtrip_real_tpm`, real hardware)

- ASUS UX5406S fTPM: **PASS**
- minihost (NexiGo host): **PASS**
- archhost (HP): **PASS**
- Contributor's Lenovo fTPM: PASS (per the PR description)

Four independent TPM vendors/firmware, all green.

### Unit suite

`cargo test -p irlume-core`: **124 passed, 0 failed, 26 ignored** (matches the contributor's count exactly).

### Camera regression (slice-4 stress test, at PR head `464746b`)

The change is in `irlume-core`, which the whole daemon links, so we re-ran the camera stress test as a no-regression gate:

- ASUS dual: **PASS**
- archhost Brio sequential: **PASS**
- minihost NexiGo dual: **FAIL (status 101)** — the known NexiGo USB flaking (intermittent, environmental, not reproducible and unrelated to this diff; this PR does not touch the camera path).

## Why this fix is load-bearing (for future reference)

The TPM 2.0 session key is derived as `sessionKey = KDFa(sessionAlg, bind.authValue || salt, ATH, nonceTPM, nonceCallerStart, bits)`. When both `tpmKey` (the salt) and `bind` are `TPM_RH_NULL`, that key is the **EmptyBuffer (all zeros)**. With an empty key the TPM's response-parameter "encryption" is effectively a no-op, but ESYS still applies response decryption, so `TPM2_Unseal` returns well-formed but wrong bytes. That is why the sealed object verified fine under tpm2-tools while irlume's `unseal` returned garbage.

The fix salts the three `unseal*` sessions against irlume's SRK, giving both sides a real, identical session key, and ties the encrypt/decrypt attributes to `salt.is_some()`. Two invariants to preserve in any future change:

1. Never enable `encrypt`/`decrypt` on an unsalted, unbound session. It is the empty-key trap this PR fixes.
2. Never salt a trial session. It carries no secret, so salting would add an SRK dependency to a pure policy computation for no benefit.

References: TPM 2.0 Library spec Part 1 (session key derivation), the Linux kernel TPM security documentation ("every HMAC session … must be created using the null primary key as the salt key"), and the tpm2-tss changelog entry "Fixed mixing salted and unsalted sessions in the same ESAPI context".

archledger (approved):
Independent review (GLM 5.2, different model family): **APPROVE**.

The fix is correct and well-scoped. Verified against the vendored tss-esapi 7.6.0 source that `salt` maps to `start_auth_session`'s `tpm_key` parameter, and that `SessionAttributesBuilder`'s mask always covers both decrypt+encrypt bits, so trial sessions genuinely clear them while unseal sessions genuinely set them. All 7 `with_session` call sites are correctly classified (3 policy sessions salted with `Some(*srk)`, 4 trial sessions unsalted).

Edge cases checked: salting does not perturb the policy digest, so envelopes sealed before this fix unseal correctly after it; transient and persistent SRK are both valid salt keys (the `decrypt` bit is set on `srk_template`, which is what a salt key needs); salting without binding is correct; `KeyHandle` is `Copy`, so `*srk` is a valid value copy of a still-loaded handle.

One non-blocking pre-existing nit (not this PR): in `load_or_create_srk`'s foreign-key-squatting branch the squatter's handle from `tr_from_tpm_public` is never explicitly flushed. Harmless (each unseal opens a fresh Context), but worth a follow-up.

Backed by: real-TPM seal/unseal roundtrip passing on four independent TPMs (contributor's Lenovo fTPM, plus ASUS, minihost, and archhost), and 124 non-ignored irlume-core tests.

PR #501, opened 2026-08-18
feat: add shadow closure calibration selector

### LINKED ISSUE #173: Closure calibration is one snapshot per user, but glasses change the eye it measured

archledger:
## What & why

Advances #173 with the evidence-only slice recommended by the deep-dive research. Glasses and lighting can move eye-closure EAR enough that one stored calibration misses genuine closures, but accepting against every calibration would widen the consent gate.

This PR deliberately does **not** change authentication, enrollment storage, daemon IPC, PAM, or any authorization outcome.

## What changed

- adds a primary-source deep dive covering safe multi-calibration selection and the missing evidence
- adds a pure fail-closed selector returning only `Unique`, `Ambiguous`, or `OutOfRange`
- adds developer-only `IRLUME_DEV=1 irlume blinkcap select` shadow replay
- derives explicit profile ranges from repeated open/closed recording medians
- requires an explicit pre-gesture prefix length, discards that prefix, and runs the existing detector only for a unique profile
- strictly validates bounded regular-file JSON/JSONL evidence, distinct capture paths, labels, frame counts, required fields, and consecutive indices
- validates every attempt before printing, and retains only one parsed trace at a time
- documents the manifest and developer workflow

## Safety boundary

- no nearest-profile fallback
- no OR-across-profiles behavior
- ambiguous, closed-like, out-of-range, malformed, or incomplete evidence produces `shadow-consent=not-run`
- output is marked `SHADOW ONLY; never authorizes`
- selector symbols have no callers in auth, daemon, storage, common IPC, or PAM crates
- no new dependencies or lockfile changes

This is instrumentation for building the held-out adversarial corpus. It does not close #173 and does not commit to a production profile schema or selector threshold.

## Validation

Exact head: `4edf283`

- `cargo test --workspace` — pass, zero runnable failures
- `rustup run 1.88.0 cargo check -p irlume-cli` — pass
- prefix-discard regression verified red under deliberate full-sequence sabotage, then green after restoration
- independent exact-head review: no Critical, Important, or Minor findings; Ready to merge

### DIFF
--- a/crates/irlume-cli/src/blinkcap.rs
+++ b/crates/irlume-cli/src/blinkcap.rs
@@ -17,6 +17,23 @@
 //!
 //!   replay:  `IRLUME_DEV=1 irlume blinkcap replay data/`   (a file or a directory)
 //!
+//!   selector shadow:
+//!   `IRLUME_DEV=1 irlume blinkcap select --profiles profiles.json \
+//!      --attempts attempts/ --prefix-frames 6`
+//!
+//! The selector manifest names repeated open/closed blinkcap recordings for
+//! each condition. Paths are relative to the manifest unless absolute:
+//!
+//! ```json
+//! {"profiles":[{"name":"desk-glasses",
+//!   "open":["open-1.jsonl","open-2.jsonl"],
+//!   "closed":["closed-1.jsonl","closed-2.jsonl"]}]}
+//! ```
+//!
+//! Selector output is evidence only. It never reaches the daemon, enrollment,
+//! or authorization outcome, and the prefix length is deliberately explicit
+//! because no production value has been qualified.
+//!
 //! Labels are free text; the replay summary groups by them. The suggested set
 //! for the consent-gesture campaign: `held-closure` (genuine deliberate closes),
 //! `natural-blink` (passive spontaneous blinks, must NOT pass), `ae-settle`
@@ -25,9 +42,10 @@
 
 use crate::{engine, flag};
 use irlume_liveness::{
-    detect_blink, detect_deliberate_closure, BlinkResult, ClosureCalibration, EarSample,
+    detect_blink, detect_deliberate_closure, select_closure_profile, BlinkResult,
+    ClosureCalibration, ClosureProfileRange, ClosureProfileSelection, EarSample,
 };
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::ExitCode;
 
 /// Derive a [`ClosureCalibration`] from a pool of samples: open = the 75th
@@ -94,11 +112,13 @@ pub fn run(args: &[String]) -> ExitCode {
     match args.get(1).map(String::as_str) {
         Some("capture") => capture(args),
         Some("replay") => replay(args),
+        Some("select") => selector(args),
         _ => {
             eprintln!(
-                "usage: irlume blinkcap <capture|replay>\n  \
+                "usage: irlume blinkcap <capture|replay|select>\n  \
                  capture --label L --det <y.onnx> --model <g.onnx> --mesh <fl.onnx> --out F.jsonl [--ir DEV] [--n 75]\n  \
-                 replay <file.jsonl | dir>   (runs the detectors + sweeps the closure threshold)"
+                 replay <file.jsonl | dir>   (runs the detectors + sweeps the closure threshold)\n  \
+                 select --profiles P.json --attempts <file.jsonl | dir> --prefix-frames N"
             );
             ExitCode::from(2)
         }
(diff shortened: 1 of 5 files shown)


archledger:
## ASUS hardware smoke — 2026-08-18

Tested this PR's exact head `4edf28346d971c6e06e3a0819cc0edc441ba0075` on the built-in ASUS FHD RGB+IR module, glasses off at approximately arm's length.

| Capture | Face/EAR samples | EAR median | Central 80% |
|---|---:|---:|---:|
| bare-open-1 | 450 | 0.2439 | 0.2190–0.2595 |
| bare-open-2 | 150 | 0.2287 | 0.0879–0.2547 |
| bare-closed-1 | 75 | 0.0319 | 0.0143–0.0467 |
| bare-closed-2 | 450 | 0.0433 | 0.0234–0.0592 |

The derived open selector range was the two per-recording medians, `0.2287–0.2439`.

- With a six-face-frame prefix, both genuine open takes were `out-of-range`; the closed-first take was also `out-of-range`. Consent was never run.
- With a one-face-frame diagnostic prefix, open-1 selected uniquely and returned `NoBlink`; open-2 stayed `out-of-range`; closed-first stayed refused.
- No false shadow consent occurred.

This is the intended kind of negative shadow result: complete raw-prefix containment cannot use unexpanded min/max session medians. Real genuine prefix samples reached roughly `0.2116–0.2501`. This does **not** justify widening production bounds; it identifies range construction/prefix statistics as the next offline experiment.

Safety/restoration evidence:

- daemon and socket restored active;
- RGB and IR privacy readback both `0`;
- emitter stream journal had no unresolved record;
- root-owned captures had complete consumer scanning and recovery bookkeeping;
- the first open take completed/restored normally but lacked root crash journaling, so it is treated as preliminary.

Raw JSONL remains private, mode `0600`, outside Git, with verified SHA-256 checksums. Only these aggregate measurements are posted here.

PR #516, opened 2026-08-22
Ship both PAD models default-on: ViT RGB + FLIR IR (ADR-0013)

archledger:
## What

Both measured PAD models ship DEFAULT-ON in every package (user decision 2026-08-22, ADR-0013):

| cue | model | operating point | measured coverage |
|---|---|---|---|
| RGB | `liveness_vit.onnx` (Adedev-W, MIT) | m96 crop, 0.60, **5-frame-median vote** per auth | print/banner 100% across 2 sessions, 0/180 genuine false-fires (desk/dim/close/glasses) |
| IR | `flir.onnx` (DAMO, MIT) | 0.9, lit IR frames | 122/123 banner frames; revalidated 0/3,940 at 197 identities on CBSR 850nm |

Both DENY-ONLY (worst case = password fallback), threshold-pinned by unit tests from both sides, kill-switched (`IRLUME_PAD_VIT=0` / `pad_vit=0`, `IRLUME_PAD_IR=0` / `pad_ir=0`).

## Disclosed gaps (ADR-0013, startup lines, models/README)

- The ViT does **NOT** stop a phone at login distance (measured 0.455, twice, bistable — `2026-08-22-vit-live-qualification.md`). That species remains IR's job; on RGB-only convenience hosts the gap stays open (unchanged: no measured cue closes it — 7-cue survey included).
- +345 MB payload; ViT scores Live frames only (59 ms 5700G / 268 ms N100).
- ADR-0001 gains a blast-radius tier: deny-only cues may ship default-on with disclosed-undocumented training; grant-capable models keep the full bar.

## Implementation

- `PadVit` classifier + pure `pad_vit_input` (m96 clamp arithmetic pinned by 4 tests; weights determinism test)
- Engine: `with_vit_pad` / `with_pad_ir`, per-request/per-enrollment vote ring (`vit_vote_denies` pure core + protocol tests)
- Wiring: cross-spectrum (FLIR lit-frame → ViT Live-frames), RGB-only tier (ViT), dark path (FLIR); enrollment resets the ring (a banner presented to enroll is denied)
- Daemon: startup verification (adapter rule), kill switches, per-cue startup lines with coverage disclosure; doctor/models copy updated
- Distribution: models-v1 assets (hashes verified at upload), SHA256SUMS, fetch-models.sh, systemd unit, spec/deb/PPA/PKGBUILD/nix — parity script green in all five lanes
- Includes the 7-cue screen-attack survey + `moire_replay` instrument (the evidence ADR-0013 cites)

[...]

PR #518, opened 2026-08-22
perf(camera): role-aware rate-gate startup flush from fleet measurement (-3.3s/auth)

archledger:
## What

The rate-gate startup flush discarded 30 frames on both roles before measuring its 30-delta window: ~2s per stream at 15fps, paid serially by each side of a sequential capture (~4.4s of every auth's capture stage on sequential hosts — both dual-camera fleet hosts today).

Re-measured the STREAMON transient the flush exists to skip, with a new raw probe (`startup_probe` + `startup_transient` example), on every node of three fleet hosts, 3 runs each:

- **Zero sequence gaps single-stream on ALL nodes** (ASUS RGB+IR, NexiGo RGB+IR, Brio RGB). The 32-gap startup that justified 30 frames was a dual-load starvation measurement from slice-8 development — a concurrent-mode failure sequential capture never sees.
- The only real transient is the **NexiGo IR's first ~4 frames** (flush-0 window: 14.018fps vs 14.7 floor; settled from dequeue 4).
- **No RGB node needs any flush** (2x-4x floor margin from dequeue 0).

`startup_flush(role)`: IR 10 (2.5x the measured tail), RGB 0. Fail-closed unchanged: a transient exceeding its role's flush fails the window -> capture error -> password fallback.

## Measured effect (ASUS auth path, sequential qualification, real state)

| | before | after |
|---|---|---|
| total auth | 10.38s | **7.04s** |
| RgbCapture | 5136ms | 3127ms |
| IrCapture | 4966ms | 3632ms |
| rgb/ir skew | 4380ms | 3050ms (limit 3000) |

Research note with full evidence tables: `docs/research/2026-08-22-startup-flush-fleet-measurement.md`.

## Open decision documented (not changed here)

The sequential inter-burst gap is now 50ms over the 3s pairing limit and consists almost purely of the second stream's window machinery — sequential pairs can never meet a concurrent-calibrated bound. This also keeps the #516 ViT RGB PAD unreachable in production auth on both dual (sequential) hosts. Options (schedule-aware budget / pre-arm overlap / window 25) are laid out in the research note for an owner decision.

## Fleet validation (camera-change rule)

- [x] Probe evidence: ASUS RGB+IR, NexiGo RGB+IR, Brio RGB (3 runs each)
- [ ] Physical continuity stress (sequential): ASUS, minihost — this session
- [ ] archhost: no IR node; probe + auth-path evidence
- [ ] thinkpad: OFFLINE — **merge blocked until it powers on and validates**


archledger:
## Fleet validation at 000e2edc (physical continuity stress, sequential, production TrackedStream path)

**ASUS UX5406S** (`UX5406S-Fedora`, 60s/phase): PASS —
RGB 854 frames @ 14.22 Hz, IR 857 @ 14.27 Hz, both 0 cumulative drops, gap_total 0, single deliberate mid-run recovery each (2.76s / 2.91s — post-recovery re-fill is also ~3x faster than the 8.99s recorded pre-change, since recovery re-runs the fill with the shorter flush).

**minihost / NexiGo N930W** (`minihost`, 60s/phase): PASS —
RGB 1705 frames @ 28.41 Hz, IR 822 @ 13.69 Hz, both 0 cumulative drops, gap_total 0, recoveries 2.43s / 4.41s. The NexiGo IR fill met the 14.7 floor at flush 10 through the production gate, matching the probe's 14.76 steady-state.

**archhost / Logitech Brio** (RGB-only, no IR node): the pair stress cannot run without IR. Evidence is the probe (3 runs: windows seeded at dequeue 0 deliver 14.88-15.92 fps vs the 7.5 floor, 2x margin) — the production fill consumes the identical validated-dequeue stream the probe measured, so the delta between probe and production for the RGB role is which frames feed the window: the same ones.

**thinkpad**: offline; validation and merge wait for it per the camera-change rule.

All fleet daemons stopped for the runs and restored active; all /tmp staging removed.

archledger:
## Scope addition (commits 9b12ce2, 4066056, 9d40218): the capture-path program continues on this PR

Beyond the role-aware startup flush (original scope), this branch now carries the follow-on changes from the same latency program, all validated on the reachable fleet (thinkpad still offline; merge of everything here waits for its validation):

**1. ADR-0014 — schedule-aware sequential pairing budget (8s; concurrent keeps 3s).** The measured machinery gap (3.05s ASUS / ~3.0s minihost) sat over the 3s concurrent-calibrated limit on every sequential auth, structurally discarding paired RGB — and with it the #516 ViT PAD — on both dual hosts. Derived bound: rate-gate-guaranteed machinery gap ≤3.1s + one retry + one self-heal ≈ 6.2s; 8s bounds it under the 15s login window. The IR-only path GRANTS with zero RGB evidence, so the old behavior removed cues from an already-granting decision (ADR has the full posture). **e2e validated:** ASUS and minihost auth_timing now run the full cross-spectrum gate (content verdicts replace the stale-discard reason) at their natural 3.0-3.05s gaps.

**2. Grace-window retry time-bound.** A retry that can't finish before the deadline no longer starts (it would overrun mid-capture holding the camera for an unusable result). Estimator: costliest attempt so far (held pair ~1s → sequential fallback ~7s transition handled).

**3. Rate tolerance 98%→97%** (IR floor 14.7→14.55 fps). The NexiGo's measured steady delivery is 14.73-14.79 — 0.2% margin at 98%, one drift from failing every auth on a host that passes every other gate. 97% = 1.2% margin below the worst real node; every fractional-rate delivery still rejected by miles (10Hz = 67%). Transport-health evidence, not the anti-injection control (verify_pinned is).

**4. Read-only MSXU probe + research dossiers.** Three research docs (Windows Hello camera control, camera-control safety/bricking, inference optimization) and the fleet stage-cost sweep. Findings worth review attention: the Brio's video2 is GREY 340x340 IR with the Microsoft face-auth XU advertised — **archhost is potentially the third dual-capable host**; and the consent watch runs YuNet per frame (~1.9s/watch on N100), the top CPU-side optimization target.

**Physical stress re-run at 9b12ce2 (tolerance change active in the production gate):** ASUS PASS (sequential, 60s/phase), minihost PASS. Daemons restored on all hosts; /tmp staging removed.

PR #559, opened 2026-08-27
Recognize Omarchy face lock PAM service

rfmyrick:
## Summary

Recognize Omarchy's dedicated `omarchy-lock-face` PAM service as a live-session `ScreenUnlock` service.

This keeps Irlume's existing fail-closed behavior for unknown PAM service names while allowing Omarchy's face lock integration to work when biopolicy is enabled.

## Why

Omarchy's Quickshell lock screen is adding a dedicated face-auth PAM lane (`/etc/pam.d/omarchy-lock-face`) so face authentication remains separate from the password stack. I tested Irlume 0.11.2 with that architecture on a 2026 Dell XPS 13 with an RGB + IR Windows-Hello-style camera.

The integration works for:

- automatic face unlock
- failed face attempt with password fallback
- blank -> wake -> fresh face attempt
- suspend/resume

With Irlume's biopolicy off (the default), the service works today. With biopolicy enforcing, the unrecognized service would classify as unknown and fail closed. Since this is a known local live-session lock service, adding it to the shared service table as `ScreenUnlock` is the narrow compatibility fix.

## Change

- add `("omarchy-lock-face", ServiceKind::ScreenUnlock)` to the shared PAM service table
- add a regression test asserting that classification

No behavior for other or unknown PAM service names is changed.

Related Omarchy face-auth work: basecamp/omarchy#7935.

### DIFF
--- a/crates/irlume-common/src/pam_service.rs
+++ b/crates/irlume-common/src/pam_service.rs
@@ -62,6 +62,7 @@ pub const SERVICES: &[(&str, ServiceKind)] = &[
     ("swaylock", ServiceKind::ScreenUnlock),
     ("i3lock", ServiceKind::ScreenUnlock),
     ("hyprlock", ServiceKind::ScreenUnlock),
+    ("omarchy-lock-face", ServiceKind::ScreenUnlock),
     // Display-manager greeters (cold login), including GDM's separate
     // fingerprint login service, same login class.
     ("sddm", ServiceKind::Greeter),
@@ -164,6 +165,16 @@ mod tests {
         for probe in ["SUDO", " sudo ", "Sudo"] {
             assert_eq!(classify(probe), Some(ServiceKind::Elevation), "{probe}");
         }
+    }
+
+    /// Omarchy's lock shell drives face auth through a dedicated PAM service.
+    /// It is a live-session screen unlock, not a cold-login greeter.
+    #[test]
+    fn omarchy_face_lock_is_screen_unlock() {
+        assert_eq!(
+            classify("omarchy-lock-face"),
+            Some(ServiceKind::ScreenUnlock)
+        );
     }
 
     /// The divergence that prompted this: doas was Elevation for the policy and


archledger (approved):
Approved. Validated the classification change and the premise on real hardware before approving.

**Correctness of the change**

- `omarchy-lock-face` is a live-session lock screen service, used only by Omarchy's Quickshell lock shell through a dedicated `PamContext`. It never serves a cold login, so `ScreenUnlock` (not `Greeter`) is the right class, same semantics as `kde`, `swaylock`, and `hyprlock` in the table.
- The name is sourced: basecamp/omarchy#7935 creates `/etc/pam.d/omarchy-lock-face` and drives it from the lock shell. That PR is still open upstream, but the risk is zero either way: unknown names fail closed today and keep failing closed after this PR, so only machines that actually ship that pam.d file change behavior.
- The regression test pins the classification, no other consumer changes (biopolicy, grace windows, and consent all read this one table), CI green including packit builds.

**Hardware validation** (ThinkPad X13 Yoga G4, fresh Omarchy install, Logitech BRIO rgb+ir, secure tier, stock 0.11.2 vs this branch built as 0.11.2-0.559)

| Build | biopolicy | omarchy-lock-face | Result |
|---|---|---|---|
| 0.11.2 | off | direct PAM client | grant (match 0.887, liveness Live) |
| 0.11.2 | on | direct PAM client | denied, journal: `biopolicy denies verify for service 'omarchy-lock-face' -> password`, camera never engaged |
| this PR | on | direct PAM client | grant (match 0.835, liveness Live, pad-vit 0.376) |
| this PR | on | unknown service name | still denied (regression holds) |

End to end with the omarchy#7935 lock shell on this branch:

- hands-free face unlock through the lane: grants 0.848 / 0.778; a no-face attempt denies without unlocking and retries
- fingerprint lane unlock works alongside it (looking away from the camera)
- blocking the RGB lens still unlocks through the IR identity arm (`match (ir/dark)` 0.940 / 0.817)
- cold greeter face login (SDDM lane) with TPM unseal: `UnsealPassword: OK`, and typing a password at the greeter never engages the camera

**Notes for reproducers**

- The lane is backend neutral on the Omarchy side (the shell keys on the presence of the pam.d file), so testing only needs `auth sufficient pam_irlume.so` plus the account include in that file.
- A test client launched over SSH must strip `SSH_CONNECTION`/`SSH_CLIENT`/`SSH_TTY`, otherwise pam_irlume's always-on remote-session guard silently declines and the daemon journal stays empty.
- Stock Omarchy has no gnome-keyring login keyring, so after a face greeter login `gkr-pam` logs "couldn't unlock the login keyring"; that is platform config noise, the unsealed password did reach the PAM transaction.

Issue #449, opened 2026-08-14
[hw] Enrollment failing in later versions (~ >0.7) using NexiGo HelloCam N930W (60fps)


slickerDev:
### Machine and camera

NexiGo HelloCam N930W [60fps version] (3443:930d)

### Distro and desktop

Fedora 44 KDE (Wayland)

### Output of `irlume doctor`

```text
[doctor] platform: Fedora-family
[doctor] install origin: Fedora Copr (archledger/irlume)
[doctor] TPM 2.0: /dev/tpmrm0 ✓
[doctor] Secure Boot: enabled ✓
[doctor] boot mode: GRUB/loader (self-healing PCR-7)
[doctor] capture mode: root-only setting (reads /etc/irlume/cameras.conf); re-run `sudo irlume doctor` to read it
[doctor] signed PCR policy: none (no Tier 1 on this boot chain)
[doctor] pcrlock: not provisioned: seals use the literal PCR-7 policy + recovery passphrase (re-arm/restore after firmware updates); `systemd-pcrlock make-policy` enables Tier 2
[doctor] camera nodes (classified by pixel format):
  /dev/video0: Rgb (uvcvideo, USB)
  /dev/video2: Ir (uvcvideo, USB)
[doctor] IR stream: 640x360@30.00fps GREY ✓ (Windows Hello minimum 340x340@15fps)
[doctor] RGB stream: 640x480@30.00fps YUYV ✓ (Windows Hello minimum 480x480@7.5fps)
[doctor] models:
  loaded by the daemon ✓
[doctor] ORT_DYLIB_PATH: (unset)
[doctor] ONNX Runtime: /usr/share/irlume/onnxruntime/lib/libonnxruntime.so (1.24.4, mapped from /usr/share/irlume/onnxruntime/lib/libonnxruntime.so.1.24.4) ✓
[doctor] TFLite runtime: /usr/share/irlume/tflite/libtensorflowlite_c.so ✓
[doctor] pipeline stages (candidate per this shell's search order):
  detection: face_detection_yunet_2023mar.onnx — shipped (/usr/share/irlume/models/face_detection_yunet_2023mar.onnx)
  landmarks: face_landmarks_detector.tflite — shipped (/usr/share/irlume/models/face_landmarks_detector.tflite)
  recognition: glintr100.onnx — shipped (/usr/share/irlume/models/glintr100.onnx)
  (third-party models accepted for: recognition, pad; #276. `irlume models` lists them)
[...]

PR #572, opened 2026-08-27
feat: name the accepted face-authentication mode in ir-setup evidence

### LINKED ISSUE #565: Spec-driven emitter discovery from MS-XU GET_MAX answers
## What and why

irlume's emitter discovery (`crates/irlume-camera/src/ir_emitter.rs`) currently derives a value to write from the camera's reported control answers, but the built-in table replays payload blobs validated on specific hardware, and unvalidated USB IDs (like the 60fps NexiGo `3443:930d` in #449) stay outside it until on-device evidence exists.

The Microsoft UVC 1.5 extension unit specification fully defines the Face Authentication Control (0x06) payload shape: `bNumEntries` followed by one `(bInterfaceNumber, bmControlFlags)` pair per streaming interface, where the flags select exactly one of D0 (general purpose), D1, or D2 (face-auth modes; a stream advertises D1 or D2 via GET_MAX, never both). Spec: https://learn.microsoft.com/en-us/windows-hardware/drivers/stream/uvc-extensions-1-5 (section 2.2.2.6). irlume's existing built-in payload `[1, 3, 2, 0, 0, 0, 0, 0, 0]` decodes exactly as one entry, interface 0x03, flags D1, confirming the shape.

## The task

1. Discovery should construct spec-conformant candidate payloads from the camera's own GET_MAX answer (which streaming interfaces are face-auth capable, and which mode each advertises) instead of relying only on camera-reported CUR/DEF derivation, so an unvalidated device gets a spec-valid attempt.
2. Record WHICH mode a module accepted (D1 or D2) as part of the discovery evidence, so USB-ID promotion to the built-in table carries the full chain: descriptor, write, read-back, optical output, restoration, accepted mode.
3. Keep every existing safety property: descriptor-checked writes only, journaled with restore-on-drop, refuse-on-error, optical verification.

## Acceptance

- `sudo irlume ir-setup` on a device with no built-in table entry can light the emitter via a spec-constructed payload when the camera answers GET_MAX sanely
- the discovery transcript names the accepted mode
- unit tests cover payload construction from representative GET_MAX answers (including a camera advertising D2)
- all existing ir_emitter tests keep passing

Labels: camera, enhancement. Hardware validation on the fleet (BRIO + NexiGo variants) before merge per the standing rule.

archledger:
## What & why

Closes #565.

Reading the issue against the code first: the spec-driven payload construction from GET_MAX answers already exists (`face_auth_payload` in ir_emitter.rs, with every structural contradiction a refusal and D2 background-subtraction refused rather than driven, per the existing tests). What was missing is the evidence half of the issue: the ir-setup transcript carried the accepted mode as raw bytes, so the USB-ID promotion chain (descriptor, write, read-back, optical output, restoration) could not be filed without hand-decoding.

This adds `face_auth_mode_evidence`: a pure reader that turns a Face Authentication SET_CUR payload into one clause per streaming interface in the specification's vocabulary. It interprets, never guesses: every shape rule the builder enforces is a None rather than a best effort. Both transcripts now carry it:

- applied control: appends "accepted: interface 0xNN set to D1 (alternative illumination)"
- active by device default: names what GET_CUR and GET_DEF report per interface, instead of assuming D1

## Testing

- [x] `cargo test --workspace` passes (1753 passed, 0 failed; three new tests for the reader, one existing message test updated to the richer wording)
- [x] Docs/CHANGELOG updated if user-facing (message wording only; no CHANGELOG entry)
- [ ] Hardware-validated (if it touches PAM, the daemon, or capture): NOT yet run on hardware this session; the message path is string-level and unit-tested, the fleet BRIO is default-active so a live `sudo irlume ir-setup` on it would exercise the rewritten default-active message. Can run on the thinkpad before merge on request, and the #449 reporter's eventual transcript will carry the new clause on real discovery hardware.

### DIFF
--- a/crates/irlume-camera/src/ir_emitter.rs
+++ b/crates/irlume-camera/src/ir_emitter.rs
@@ -4853,6 +4853,55 @@ pub(crate) fn face_auth_payload(def: &[u8], max: &[u8]) -> std::result::Result<V
     Ok(out)
 }
 
+/// Read a Face Authentication `SET_CUR` payload back as promotion evidence.
+///
+/// The `ir-setup` transcript is the evidence chain for adding a USB ID to the
+/// built-in table: descriptor, on-device write, read-back, optical output,
+/// restoration. The raw bytes leave the reader to decode the accepted mode by
+/// hand; this names each streaming interface and the mode it was set to, in the
+/// specification's own vocabulary, so a transcript can be filed as-is.
+///
+/// It interprets, never guesses. Every shape rule the builder enforces is a
+/// `None` here rather than a best effort, because evidence that might be wrong
+/// is worse than evidence that admits it read nothing.
+pub(crate) fn face_auth_mode_evidence(payload: &[u8]) -> Option<String> {
+    const D0_GENERAL: u8 = 1 << 0;
+    const D1_ALTERNATIVE_ILLUMINATION: u8 = 1 << 1;
+    const D2_BACKGROUND_SUBTRACTION: u8 = 1 << 2;
+    const DEFINED: u8 = D0_GENERAL | D1_ALTERNATIVE_ILLUMINATION | D2_BACKGROUND_SUBTRACTION;
+
+    let len = payload.len();
+    if len < 3 || !(len - 1).is_multiple_of(2) {
+        return None;
+    }
+    let entries = usize::from(payload[0]);
+    let capacity = (len - 1) / 2;
+    if entries == 0 || entries > capacity {
+        return None;
+    }
+    let tail = 1 + entries * 2;
+    if payload[tail..].iter().any(|&b| b != 0) {
+        return None;
+    }
+    let mut parts = Vec::with_capacity(entries);
+    for i in 0..entries {
+        let at = 1 + i * 2;
+        let interface = payload[at];
+        let flags = payload[at + 1];
+        if flags & !DEFINED != 0 || flags.count_ones() != 1 {
+            return None;
+        }
+        let mode = match flags {
+            D0_GENERAL => "D0 (general purpose)",
+            D1_ALTERNATIVE_ILLUMINATION => "D1 (alternative illumination)",
+            D2_BACKGROUND_SUBTRACTION => "D2 (background subtraction)",
+            _ => return None,
+        };
+        parts.push(format!("interface 0x{interface:02x} set to {mode}"));
+    }
+    Some(parts.join(", "))
+}
+
 /// Whether an IR Torch `GET_DEF` value is one the specification says may be
 /// written back.
 ///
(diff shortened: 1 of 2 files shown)


archledger:
Hardware validation (thinkpad, Omarchy, Logitech BRIO rgb+ir, build 0.11.2-1 from this branch):

```
$ sudo irlume ir-setup --dry-run
[ir-setup] extension units: ... unit 12 (Microsoft camera control): advertises [0x06, 0x09]; ...

$ sudo irlume ir-setup
[ir-setup] IR emitter mode active by device default: 12:6:1,2,2,0,0,0,0,0,0 on the camera's
Microsoft camera-control unit; GET_CUR and GET_DEF both report the validated Face
Authentication value, interface 0x02 set to D1 (alternative illumination)
(no camera write or saved config was needed)
```

The BRIO took the active-by-device-default path and the rewritten message reads the accepted mode out of the payload on real hardware (interface 0x02, D1). End-to-end auth on the same build: grant through the omarchy-lock-face lane (biopolicy enforcing). The applied-control wording will additionally surface on the first true discovery device; the #449 reporter's 60fps N930W transcript will carry it when they run the green-lit ir-setup.

archledger:
Fleet hardware pass on this build (3 camera hosts, 2026-08-27):

- thinkpad (Omarchy, BRIO rgb+ir): ir-setup took the active-by-device-default path and the rewritten message named the accepted mode from the payload (transcript in the earlier comment); end-to-end face grant through the omarchy-lock-face lane on the same build (biopolicy enforcing).
- minihost (Arch, NexiGo N930W c803): discovery ran the full safety chain, write and read-back accepted, optical classification honestly inconclusive with nobody in view (0 to 9 mean, needs +20), exact original restored and verified, nothing saved. Daemon healthy after.
- ASUS Zenbook S 14 (Fedora 44, Shinetech rgb+ir): discovery inconclusive via the D1 metadata-transition correlation variant (camera unattended), exact original restored and verified; the built-in known control still lights the emitter at capture and the daemon is healthy on this build.

The applied-control message clause is unit-tested; its first real-hardware transcript will come from the #449 reporter's 60fps N930W when they run the green-lit ir-setup.

archledger:
Post-merge: the applied-control path is now hardware-validated too. archhost (Arch, AppArmor enforcing, NexiGo N930W c803) after fixing an unrelated AppArmor gap (see next PR):

```
[ir-setup] IR emitter enabled: 4:6:1,3,2,0,0,0,0,0,0 on the camera's Microsoft
camera-control unit, using a value built from what the camera reports about that
control (saved; future captures rebuild it the same way); accepted: interface 0x03
set to D1 (alternative illumination)
```

That is every message path exercised on real hardware: applied (archhost), active-by-default (thinkpad BRIO), inconclusive-with-restore (minihost and ASUS). #565 fully closed.

PR #577, opened 2026-08-28
feat: record, report, and fuzz MS-XU metadata illumination (#568)

### LINKED ISSUE #568: Standardize MS-XU metadata illumination as a pipeline capability
## What and why

The MS-XU metadata control (0x09) carries per-frame metadata items including illumination data; it is the only cross-vendor ambient-light signal with a specification behind it. The irlume daemon already parses illumination metadata opportunistically on the Logitech BRIO (an extra /dev/videoN node consumed during capture: "illumination metadata closed after N classified frames" in the journal), but the handling is ad hoc per device rather than a standardized pipeline stage.

Spec: https://learn.microsoft.com/en-us/windows-hardware/drivers/stream/uvc-extensions-1-5 (section 2.2.2.9, Metadata Control).

## The task

- Make MS-XU metadata a first-class, detected capability: during camera qualification, discover whether the pair exposes a metadata node and record it in the qualification context.
- Standardize illumination ingestion: expose the ambient reading in diagnostics (`irlume camera diagnostics`) and consider it as an input to dark-path decisions where it correlates with the existing IR-dark detector (never overriding the optical IR measurement, which stays authoritative).
- Upstream context: uvcvideo is actively fixing partial metadata buffer delivery ("Avoid partial metadata buffers" in current mainline), so handle short/malformed metadata defensively (the parsing is untrusted input; keep it in fuzzed territory).

## Acceptance

- qualification context records the metadata node presence per camera pair
- `irlume camera diagnostics --json` reports the illumination stream state
- malformed metadata never panics or stalls capture (fuzz corpus case)
- no behavior regression on cameras without metadata nodes (the common case)

Labels: camera, enhancement.

archledger:
## What

Closes #568. Three additions, one deliberate non-change:

1. **Qualification records metadata node presence.** `measure_capture_qualification_with_progress` probes the IR node's MS-XU metadata sibling (the existing `TRY_FMT` path, no queue state touched) and stores `ir_illumination_metadata` on the qualification attempt. Presence only, never the `/dev/videoN` path, because node numbering is unstable across reboots.

2. **`camera diagnostics --json` reports the illumination stream state.** New `illumination` section on `CameraDiagnosticsReport`: node present/absent, frames the camera classified, the subset it flagged lit, and whether a real emitter-off frame was observed. Harvested from the same bounded gated capture that already measured delivered rate (its `IrCaptureStats` were computed and discarded before).

3. **The parser is fuzzed territory.** New `uvc_illumination` cargo-fuzz target: determinism, no panic on any byte string (partial first buffers were possible on uvcvideo until the 6.12.97-era "Avoid partial metadata buffers" series, Fixes 088ead255245), plus selection invariants (in-burst, lit beats dark when lit exists). Eight checked-in seeds from the unit fixtures, replayed as a regression gate, and a CI slot alongside the other three targets.

**Non-change, by design:** the presence fact lives on the attempt, never on `QualificationContext`, whose serialized form is the runtime key and whose full-struct equality gates stored authority. Verified by test: records written before this change keep parsing and keep authorizing, and cameras without metadata nodes see no difference. No schema, policy, or engine version bump (old binaries ignore the new field; serde round-trip both ways is tested).

## Evidence

- Strict TDD: runtime RED captured for the presence mapping, the lit-count, and the illumination-state assembly before their implementations; legacy-record and legacy-report compatibility pinned by tests.
- Local gates: `cargo fmt --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --locked` clean, workspace `cargo test` 1766 passed / 0 failed.
- Fuzz: 8 seeds replay clean; 13.6M executions in a 30s local run, zero crashes/hangs (`uvc_illumination`).
- Spec verification (docs/research/2026-08-27-msxu-metadata-illumination.md, in this PR): MS-XU 2.2.3.4.4 record is `{id=6, size=16, Flags bit0, Reserved}`; there is no ambient-lux item in the standard set, so the section reports flag-derived state, nothing invented.
- Dark-path correlation (task bullet) already existed (#167/#264/#221/#312: `DarkEvidence::frames_lit`/`lit_max_mean` feed `ir_dark::diagnose`); untouched here.

## Fleet validation

Camera-path change, so the fleet rule applies; matrix run follows on archhost (BRIO + NexiGo), ASUS (internal pair), and the RGB-only thinkpad (no-IR-pair case) before merge. Results will be posted here.


archledger:
## Fleet validation matrix (camera-change rule)

Scratch-daemon runs (branch binaries, `IRLUME_STATE_DIR`/`IRLUME_CONFIG_DIR`/`IRLUME_SOCKET` overrides, system packages untouched and restored after):

| host | pair | `camera diagnostics --json` illumination | qualification record |
|---|---|---|---|
| archhost | NexiGo N930W (rgb/ir) | `node: present`, 10 classified, 5 lit, ambient observed | `ir_illumination_metadata: present` on last_attempt + authoritative, schema_version 2 unchanged |
| archhost | Logitech BRIO (rgb/ir) | `node: present`, 9 classified, 5 lit, ambient observed | `present` on both attempts |
| archhost | NexiGo, `IRLUME_NO_ILLUM_META=1` | `node: absent`, 0 classified, 0 lit, ambient false | (metadata-less camera simulation: probe and reader agree on absent, no fallback breakage) |
| ASUS Zenbook UX5406S | Shinetech internal pair | shutter engaged: `node: present`, all counts null (capture refused, honest not-measured); shutter open: 10 classified, 5 lit, ambient observed | `present` on both attempts, outcome concurrent_qualified |
| thinkpad | Chicony RGB-only (no IR camera) | ir role `missing`, `illumination: null` | no pair to measure, nothing written |

Bonus paths exercised on real hardware:

- The privacy-shutter fail-closed path (ASUS): IR capture refused, diagnostics answered with not-measured nulls rather than erroring or fabricating counts.
- SELinux-enforcing host: the daemon runs fine from a `bin_t`-labeled path (ops note for scratch testing, not a code issue).
- Stored system qualifications on all three hosts were never touched (scratch state dirs), and every system daemon was restored and verified healthy after its run.

PR #582, opened 2026-08-28
fix: apparmor profiles must allow locking the config lock file (#580)

### LINKED ISSUE #580: AppArmor: irlumed profile denies file_lock on /etc/irlume/cameras.conf.lock, so set-cameras cannot persist
## What and why

On the AppArmor-enforcing host (archhost, 0.11.2-1 Arch package), `sudo irlume set-cameras /dev/video4 /dev/video6` switches the running daemon live but fails to persist:

```
[set-cameras] cameras set to rgb=/dev/video4 ir=/dev/video6 (live only; could not persist: Permission denied (os error 13))
```

The kernel denial:

```
audit: type=1400 ... apparmor="DENIED" operation="file_lock" class="file"
profile="irlumed" name="/etc/irlume/cameras.conf.lock" pid=58881
comm="irlume-camera" requested_mask="k" denied_mask="k" fsuid=0 ouid=0
```

The profile allows creating/writing the lock file but not LOCKING it (`k`), so the persist path dies after the lock file appears, leaving `/etc/irlume/cameras.conf` unwritten and a stray `cameras.conf.lock` behind. After a daemon restart the old pair comes back, which reads as "set-cameras silently reverted".

Same class as #573 (ancestor-fsync chain): a confinement gap found on the one enforcing host, invisible everywhere else. `comm=irlume-camera` indicates the CLI's camera thread runs under the `irlumed` profile, so the fix belongs in the shipped profile (lock rule for the conf lock file) plus the packaging parity ratchet's profile checks, and a live-profile patch on archhost.

## Evidence

- Found during the 2026-08-28 far-range dark-path sweep session on archhost (BRIO pair switch).
- Workaround used (do not ship): writing cameras.conf directly as root, then restarting the daemon.

Labels: bug, security.

archledger:
Closes #580.

Config writes serialize on a lock file beside the config (`config.rs lock_exclusive`). The shipped `irlumed` profiles granted the lock file's create and write through the broad `/etc/irlume/** rw` rule but not `flock`'s `k`, so on an enforcing host `set-cameras` switched the running daemon live, printed the "live only; could not persist" warning, and the next daemon restart silently reverted the pair. Same class as #573: a confinement gap only the enforcing host can hit.

Fix: `rwk` on `/etc/irlume/*.lock` in both profile variants, asserted by the packaging parity ratchet (the ratchet entry landed first and was watched to MISS both profiles before the profile change, test-first). Found enforcing on Arch during the far-range dark-path sweep; the audit denial line is in the issue.

Hardware validation follows below: live patch on the enforcing Arch host, then a persisting `set-cameras` including across a daemon restart.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to irlume are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- **AppArmor: `set-cameras` could not persist the camera pair.** Config
+  writes serialize on a lock file beside the config (`cameras.conf.lock`),
+  and the shipped `irlumed` profiles granted the lock file's create and
+  write but not `flock`'s `k`, so on an enforcing host `set-cameras`
+  switched the running daemon live, reported success with a persist
+  warning, and the next daemon restart silently reverted the pair. Both
+  profile variants now carry `rwk` on `/etc/irlume/*.lock`, the packaging
+  parity ratchet asserts the rule so the gap cannot re-ship, and the fix is
+  hardware-validated (live patch on the enforcing Arch host completed a
+  persisting `set-cameras` immediately after; issue #580).
+
 ### Added
 
 - **Camera triage docs: the kernel quirk table, first.** DEBUGGING.md gains
--- a/packaging/apparmor/usr.bin.irlumed
+++ b/packaging/apparmor/usr.bin.irlumed
@@ -152,6 +152,13 @@ profile irlumed /usr/bin/irlumed flags=(attach_disconnected) {
   # Config + state: enrollments (per-user), template keys, recovery blobs.
   /etc/irlume/ r,
   /etc/irlume/** rw,
+  # Config writes serialize on a lock file beside the config (config.rs
+  # lock_exclusive). The broad rw rule above grants the lock file's create
+  # and write; flock additionally needs `k`, without which set-cameras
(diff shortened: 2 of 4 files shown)


archledger:
## Hardware validation (enforcing Arch host, live-patched profile)

- Backed up the live profile to /tmp/usr.bin.irlumed.pre-580.bak, installed the PR's /etc/apparmor.d/usr.bin.irlumed, `apparmor_parser -r` reloaded it (profile loaded, enforcing).
- `sudo irlume set-cameras /dev/video4 /dev/video6` now answers with NO persist warning and writes the full pin:

```
-rw------- 1 root 0 82 Aug 28 02:48 /etc/irlume/cameras.conf
rgb=/dev/video4
ir=/dev/video4 -> (rgb_id/ir_id 046d:085e:e179cb54 recorded)
```

(corrected: `ir=/dev/video6`, with `rgb_id=ir_id=046d:085e:e179cb54`)

- `systemctl restart irlumed`: the pair SURVIVES the restart (journal: `cameras rgb=/dev/video4 ir=/dev/video6 (secure tier)`), which is exactly what failed before the fix.
- Restored the host: cameras.conf and its lock removed, daemon restarted, auto-pick back on the NexiGo pair, daemon healthy. The patched profile stays live (same content this PR ships), backup retained.

The ratchet half was done test-first: the parity entry landed first and was watched to MISS both profiles (exit 1) before the profile rules were added.

PR #583, opened 2026-08-28
feat: fingerprint wiring for Omarchy's own PAM layout

archledger:
## What

On a fresh Omarchy install, fingerprint was dead at every surface (pkexec, lock screen, greeter): Omarchy is Arch underneath, irlume's Arch arm punts on PAM wiring ("no wiring tool on this distro"), and nothing else wires pam_fprintd. Omarchy, uniquely among Arch systems, OWNS a fingerprint contract its lock and scripts depend on, verified from basecamp/omarchy (`bin/omarchy-setup-security-fingerprint`, `bin/omarchy-apply-lock`) and on a live fresh install:

- sudo and polkit-1 carry a clamshell gate (`pam_exec omarchy-hw-laptop-closed`, skips the reader when the lid is shut) plus `sufficient pam_fprintd.so`
- the lock screen uses a dedicated `/etc/pam.d/omarchy-lock-fingerprint` lane, created exactly when fingers are enrolled and removed otherwise

This PR teaches `irlume fingerprint enable|disable` that layout, byte-identical to upstream so `omarchy-apply-lock` and irlume converge instead of fighting:

- new `omarchy_present()` platform probe (fixture-tested; keyed on `/usr/share/omarchy`)
- the Omarchy arm is a guarded match case: stock Arch, Fedora, and Debian paths are byte-for-byte untouched
- constants pinning the upstream lines are themselves pinned by tests, so an upstream layout change fails our CI before shipping a mismatch
- enable: backup-once `.pre-irlume`, idempotent inserts, the minimal polkit-1 creation matches upstream's, lock lane written when fingers are enrolled; disable reverses all three idempotently

Face-at-stock-lock stays out of scope (upstream omarchy#7935 territory), as is SDDM fingerprint (upstream's own script does not wire SDDM either; upstream parity is the rule).

## Evidence

- Strict TDD: REDs captured for the probe (mutation-verified), the stack-wiring pure functions (top-insert, gate-before-existing-fprintd, idempotency, unwire round-trip), and the upstream byte-pinning tests.
- Gates: fmt clean, clippy `--workspace --all-targets -D warnings` clean, `cargo doc -D warnings` clean, workspace 1789/0.
- Live validation on the fresh Omarchy thinkpad follows below (the three surfaces that were dead).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,22 @@ All notable changes to irlume are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- **Omarchy: `irlume fingerprint enable` now wires the distro's own
+  fingerprint layout.** Omarchy (Arch underneath) owns an opinionated
+  fingerprint contract its lock and scripts depend on: a clamshell gate
+  (skip the reader when the lid is shut) plus `sufficient pam_fprintd.so`
+  in sudo and polkit-1, and a dedicated `omarchy-lock-fingerprint` PAM
+  lane that exists exactly when fingers are enrolled. irlume's Arch arm
+  previously punted ("no wiring tool on this distro"), so on a fresh
+  Omarchy install pam_fprintd was wired nowhere and fingerprint was dead
+  at pkexec, the lock screen, and the greeter. The new arm, active only
+  when the omarchy package is detected (`/usr/share/omarchy`), wires
+  byte-identical lines so `omarchy-apply-lock` and irlume converge instead
+  of fighting; `disable` reverses all three idempotently. Stock Arch,
+  Fedora, and Debian paths are untouched.
+
 ### Fixed
 
 - **AppArmor: `set-cameras` could not persist the camera pair.** Config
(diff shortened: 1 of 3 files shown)


archledger:
## Live validation (fresh Omarchy thinkpad, the machine that found the gap)

`sudo irlume fingerprint enable` (branch binary) produced exactly the upstream contract, verified from the files:

- `/etc/pam.d/sudo`: clamshell gate line 1, `sufficient pam_fprintd.so` line 2
- `/etc/pam.d/polkit-1`: same pair on top; the existing pam_irlume face line intact below (coexist mode)
- `/etc/pam.d/omarchy-lock-fingerprint`: created 0644, byte-identical to `omarchy-apply-lock`'s content
- backups: `sudo.pre-irlume`, `polkit-1.pre-irlume` (the latter still holding the pre-any-irlume state, not overwritten)

The three previously-dead surfaces, user-tested immediately after:

| surface | before | after |
|---|---|---|
| pkexec prompt | fingerprint ignored | finger unlocks |
| lock screen | fingerprint ignored | finger unlocks |
| sudo | fingerprint ignored | finger unlocks |

Face lanes kept working throughout (polkit face via `yes` + Enter, SDDM face via Enter on the empty password field).

Issue #586, opened 2026-08-28
Dell XPS 13: concurrent camera qualification fails continuity despite healthy streams, while camera-tune reports concurrent


rfmyrick:
### irlume version and install channel

Irlume: 0.11.2

### Distro and desktop

Omarchy / Arch Linux

## Summary

On a 2026 Dell XPS 13 running Omarchy/Arch with Irlume 0.11.2, the
internal Windows Hello-style camera is physically capable of fast
concurrent RGB+IR capture, but Irlume consistently persists
`sequential_required / invalid_provenance` because the concurrent arm
fails continuity in 4 of 6 rounds.

The behavior reproduces after a clean reboot.

There is also a reporting issue: `irlume camera-tune` says the capture
mode is concurrent even though the authoritative qualification written
to disk is sequential.

## System

- Irlume: 0.11.2
- OS: Omarchy / Arch Linux
- Kernel: `7.1.9-arch1-2`
- Camera: Dell `Integrated_Webcam_FHD`, USB `2b7e:55c0`, UVC 1.50
- RGB: `/dev/video0`, 640x480 YUYV @ 30 fps
- IR: `/dev/video2`, 360x360 GREY @ 15 fps
- Driver: `uvcvideo`

Kernel log shows normal binding with no UVC errors, resets, bandwidth
failures, or disconnects.

## Reproduction

`sudo irlume camera-tune` reports:

    capture mode concurrent for this camera: concurrent capture keeps
    100% of RGB and 102% of IR brightness and saves 5468ms (6 rounds)

But immediately afterward `irlume camera-mode` reports:

    capture mode: sequential (daemon source: qualification-v2)
    qualification: measured_sequential
    qualification reason: invalid_provenance

The persisted qualification, reproduced after a clean reboot:

    sequential:
      completed: 6
      continuity_ok: 6
      continuity_failures: 0

    concurrent:
      completed: 6
      continuity_ok: 2
      continuity_failures: 4

Everything else in the concurrent arm passes: 6/6 stream contracts,
6/6 rate floors, 6/6 active-IR evidence, zero open/arm/capture/rate
failures, and healthy RGB/IR brightness.

Measured elapsed time was approximately:

    sequential: 6131 ms
    concurrent:  662 ms

So the concurrent arm is roughly 9x faster, but fails only continuity
provenance.

A 90-second `irlume trace` around camera-tune contained only the
CaptureQualification completion event and did not expose the specific
per-round timestamp/sequence continuity failure.

## Questions

[...]

PR #591, opened 2026-08-28
feat: per-round continuity facts name which condition failed (#586)

archledger:
## What

Answers #586's Q2 (the diagnostic path) and arms Q1 (normalizable vs genuinely unreliable) with named facts instead of an aggregate.

The round-continuity verdict decomposes into exactly four nameable conditions: a within-round stream break on either sensor, and per-sensor cross-round cursor drift (stream epoch change, cumulative drops advancing, or a stalled timestamp). Until now all four collapsed into `continuity_failures`, and neither the journal, the trace, nor the persisted record could say which fired, on our own reproduction of the reporter's Dell signature (ASUS internal pair: concurrent 6/6 on contracts, rate floors, and active-IR, brightness retention 96-104%/103-109%, but `continuous_rounds 2/6`) nor on theirs.

This PR:

- adds a pure, mutation-verified classifier (`continuity_fact`) naming all four conditions in check order, with the old boolean shape's test ported to the new vocabulary
- logs each failed round's fact to the journal under `irlume logs debug on` (`qualification round: continuity failed: ir cumulative_drops advanced between rounds (arm round 3)`)
- counts per fact on the arm (in-memory `PairSample.continuity_facts`; the persisted ArmEvidence schema is deliberately untouched)
- extends the #588 verdict message: `4 of 6 concurrent rounds failed the frame-provenance bar (3x ir cumulative_drops advanced between rounds; 1x rgb timestamp did not advance between rounds)` - most frequent fact first

## Evidence

- Gates: fmt clean, clippy `--workspace --all-targets -D warnings` clean, `cargo doc -D warnings` clean, workspace 1799/0.
- Mutation-verified REDs: the classifier mapping (a swapped arm fails its test) and the message extension (the named-fact ordering assertion).
- Hardware validation follows below on the #586 testbed (the ASUS internal pair), expecting the journal to finally name which fact fires in those 4-of-6 rounds - which is then the actual answer to the reporter's Q1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ All notable changes to irlume are documented here. This project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- **Per-round continuity facts: a failing concurrent qualification now says
+  WHICH condition fired (#586).** The round-continuity verdict is four
+  nameable conditions (a within-round stream break on either sensor, and
+  per-sensor cross-round cursor drift: stream epoch change, cumulative
+  drops advancing, or the timestamp stalling), and until now all four
+  collapsed into one aggregate counter that no diagnostic surface could
+  split. Each failed round now names its fact in the journal (under
+  `irlume logs debug on`), the arm counts per fact, and the camera-tune
+  verdict lists them most-frequent-first ("4 of 6 concurrent rounds failed
+  the frame-provenance bar (3x ir cumulative_drops advanced between
+  rounds; 1x ...)"). Diagnostics only; the persisted qualification schema
+  is untouched.
+
 ### Fixed
 
 - **`camera-tune` now reports the verdict it persisted, not the one the
--- a/crates/irlume-camera/src/lib.rs
+++ b/crates/irlume-camera/src/lib.rs
@@ -6420,7 +6420,7 @@ impl CaptureMode {
 }
 
 /// One arm of the contention probe: what a capture produced, averaged.
-#[derive(Clone, Copy, Debug, Default)]
+#[derive(Clone, Debug, Default)]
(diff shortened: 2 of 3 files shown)


archledger:
## Hardware validation (the #586 testbed, this branch's binaries)

The ASUS internal pair (concurrent-capable, 4-of-6 continuity failures on every stock run) under the branch daemon with tracing on:

The verdict message, from the product path:

```
capture mode sequential for this camera: concurrent capture keeps 111% of RGB
and 116% of IR brightness, but 4 of 6 concurrent rounds failed the
frame-provenance bar (4x rgb stream discontinuous within the round), so the
safe one-at-a-time mode is what was measured and persisted
```

and the journal, per failed round:

```
irlume[debug]: qualification round: continuity failed: rgb stream discontinuous within the round (arm round 4)
irlume[debug]: qualification round: continuity failed: rgb stream discontinuous within the round (arm round 5)
irlume[debug]: qualification round: continuity failed: rgb stream discontinuous within the round (arm round 6)
irlume[debug]: qualification round: continuity failed: rgb stream discontinuous within the round (arm round 7)
```

So the aggregate finally splits, and the first named answer to #586's Q1 on this camera class is: **within-round RGB stream breaks under concurrent load** (sequence gaps/timestamp discontinuities during the round's own stream), not cross-round cursor drift (no epoch changes, no drop-count advances, no timestamp stalls between rounds). That is USB isochronous contention corrupting the RGB stream's own continuity discipline while brightness survives, which is precisely the evidence shape the Dell reporter's machine shows, and the exact question their issue asks: this is delivered-but-undiscined frames, normalizable only if the discontinuity class turns out to be benign; the named facts now make that investigation possible per camera.

PR #592, opened 2026-08-28
feat: name the within-round continuity sub-condition (#586)

archledger:
Follow-up to #591 (merged): the within-round arm now names which of the five `is_continuous` sub-conditions fired, per stream, in gate order: timestamp clock not monotonic, timestamp discontinuity, sequence discontinuity, sequence gap, rate-window gap (plus aggregate drops for multi-frame provenance). The coarse "stream discontinuous within the round" fact is gone.

The ASUS testbed's failing rounds (which #591 narrowed to a within-round RGB break) now name the exact condition in the journal and the verdict. Hardware validation below; this is the last instrumentation step before the normalizable-vs-unreliable call for this camera class.

### DIFF
--- a/crates/irlume-camera/src/lib.rs
+++ b/crates/irlume-camera/src/lib.rs
@@ -7715,8 +7715,18 @@ impl ContinuityCursor {
 /// one fired. The string is the journal/verdict vocabulary; stable.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 enum ContinuityFact {
-    RgbWithinRound,
-    IrWithinRound,
+    RgbTimestampClock,
+    RgbTimestampDiscontinuity,
+    RgbSequenceDiscontinuity,
+    RgbSequenceGap,
+    RgbRateGap,
+    RgbAggregateDrops,
+    IrTimestampClock,
+    IrTimestampDiscontinuity,
+    IrSequenceDiscontinuity,
+    IrSequenceGap,
+    IrRateGap,
+    IrAggregateDrops,
     RgbStreamEpochChanged,
     RgbDropsAdvanced,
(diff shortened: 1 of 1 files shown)


archledger:
## Hardware validation (the #586 testbed, this branch's binaries)

```
capture mode sequential for this camera: concurrent capture keeps 90% of RGB
and 104% of IR brightness, but 4 of 6 concurrent rounds failed the
frame-provenance bar (4x rgb sequence gap within the round), so the safe
one-at-a-time mode is what was measured and persisted
```

Journal, per failed round:

```
irlume[debug]: qualification round: continuity failed: rgb sequence gap within the round (arm round 4)
irlume[debug]: qualification round: continuity failed: rgb sequence gap within the round (arm round 5)
irlume[debug]: qualification round: continuity failed: rgb sequence gap within the round (arm round 6)
irlume[debug]: qualification round: continuity failed: rgb sequence gap within the round (arm round 7)
```

The last instrumentation layer answers #586's Q1 for this camera class: the failing condition is **RGB sequence gaps within concurrent rounds**, with timestamps staying monotonic, contracts and rate floors holding, and brightness surviving. The round pattern (first rounds clean, then every round failing once load has built) is USB isochronous bandwidth pressure dropping RGB payloads mid-round. That is real frame loss, not a normalizable timestamping artifact: the provenance bar exists precisely because a gapping stream cannot prove which frames informed a pairing decision. The conservative sequential verdict is correct for this class, and normalizing gaps away would weaken the bar against a genuinely lossy transport. The Dell reporter's shape matches this signature; their persisted record's continuity counts (once they attach it) will confirm or refute the same subclass on 2b7e:55c0.

PR #605, opened 2026-08-29
feat: make the privileged `yes` confirmation optional per machine

### LINKED ISSUE #602: Let privileged surfaces opt out of the per-attempt yes, treating the PAM wiring as the consent
## The request

An opt-in setting that lets a privileged face attempt start without the literal `yes`, so wiring `pam_irlume` into `sudo` and `polkit-1` is itself the consent and the camera runs on the prompt appearing. Default unchanged.

I know this is deliberate — `docs/SETUP.md` says "This confirmation is mandatory and has no disable setting", and every gesture knob explicitly adds a factor after it rather than replacing it. So this is a request to make a designed-in decision configurable, not a bug report, and the reasoning below is the case for why it is worth making optional.

## Why the standing consent already covers it

Nobody has `pam_irlume.so` in `/etc/pam.d/sudo` by accident. It takes `login enable --with-sudo --apply`, which you have deliberately separated from the base wiring precisely because granting root by face is its own decision. Having made that decision, running a command that triggers the prompt is the act of asking for face authentication. The per-invocation `yes` re-asks a question already answered at wiring time.

## The asymmetry that makes the case

irlume already accepts standing consent instead of per-attempt consent — on the lock screen. There, the scan starts by itself with no keyword, and on Omarchy's dedicated lane it is fully hands-free. That is the surface a user hits most often in a day, and the one most likely to face a camera while someone is not deliberately authenticating.

So the principle is not "every camera open needs an explicit act". It is already "some surfaces take standing consent and some take per-attempt consent", and the line currently sits between screen unlock and privileged operations. The request is only to let the machine's owner move that line for their own privileged surfaces.

## What the default should stay

Everything as it is. The reason your current behaviour is right by default is worth keeping intact:

- with `pam_irlume` first and `sufficient`, an unconditional scan means every `sudo` opens the camera, including ones typed by someone else at the machine and ones from scripts
- empty Enter and a typed password never opening the camera is a genuinely good property, and it should stay the default

An opt-in — a `settings.conf` key alongside `enforce_biopolicy` and `forbid_external_cameras`, per-service if that fits your model better — preserves all of it for everyone who does not set it.

## Ergonomics, concretely

Typing `yes` and then holding still for a scan is slower than typing a password on this machine, where a scan is about 5.5s. The keystroke count is not the problem; the mode switch is — the field asks for a password, and the correct answer for the fast path is a word that is not a password. If the gate has to stay in some form, an empty Enter as the trigger would be a large improvement on its own: it is the gesture the on-demand greeter recipe already uses, so users would meet one convention instead of two.

The `yes` prompt also makes `sudo` unusable from a context with no controlling terminal — PAM prints `Type yes to use face authentication`, then sudo fails on `a terminal is required to read the password`. That is arguably sudo's behaviour rather than yours, but a non-interactive `sudo` under the previous backend simply fell through to the password provider without the extra prompt.

## Environment

- irlume 0.11.3 (AUR), Arch, wired to sddm, sudo, polkit-1, and Omarchy's dedicated `omarchy-lock-face` lane
- Logitech BRIO, sequential capture, ~5.5s per attempt

Context: this came out of running irlume as the backend for basecamp/omarchy#7935, where face auth is set up as a deliberate, discoverable step from a menu — the same "you already opted in" shape.

FizykPSX:
Closes #602.

Adds `privileged_face_consent` to `settings.conf`, defaulting **on**, so nothing changes for anyone who does not set it. With `privileged_face_consent=0` a face attempt on `sudo` and polkit starts the way one at the lock screen does, with no keyword.

## Correction to the case I originally made

> **The argument this PR was opened on was wrong, and I am leaving the correction visible rather than quietly rewriting it.**

I argued that irlume already accepts standing consent on the lock screen and was therefore inconsistent in refusing it for privileged services. That hands-free behaviour is basecamp/omarchy#7935's doing, not irlume's: that shell arms a `PamContext` on its own PAM service continuously, so the scan starts with no gesture because the shell starts it. Every surface irlume wires itself takes an explicit act — empty Enter at the greeters and the KDE lock, the typed `yes` on the stock Omarchy lane. As @archledger put it in #602, the line is between a shell that took responsibility for arming and everything else.

So the asymmetry case does not hold. What remains is a plainer argument: `login enable --with-sudo` is already a deliberate, separate decision, exactly because granting root by face is its own choice, and this makes the per-invocation prompt optional for an owner who has made it. That is a weaker case than the one I opened with, and the maintainer's reason for holding the line — with `pam_irlume` first and `sufficient`, every `sudo` then opens the camera, scripted ones included — is a real cost.

I am leaving the branch up because the implementation may be useful whatever is decided about the policy, particularly the daemon-side half below. Per #602 the maintainer's leaning is toward unifying the privileged trigger with the on-demand empty-Enter convention instead; I am happy to rework this into that shape, or to have it closed in favour of their own implementation.

## The daemon stays the authority

`intent_confirmation_gate` refuses a privileged request without a trusted confirmation, so a client-side skip on its own would be rejected — which is the right design and worth preserving rather than working around.

So the waiver is a distinct `IntentAttestation::PolicyWaived` rather than the PAM module claiming a `PamConversation` it never collected, and **the daemon re-reads the same policy before honouring it**. A root client asserting "policy waived this" on a machine whose policy did not waive anything is refused exactly like a missing confirmation. Without that, the option would turn into a way for any root client to skip the gate — worse than the annoyance it removes.

Absent, unset, or unreadable settings all mean the confirmation is required. A policy this process cannot read is not a policy that waived anything.

## What is in the diff

- `privileged_face_consent_visible()` / `privileged_face_consent_required()` in `config.rs`, same env-over-settings shape as `enforce_biopolicy` and `forbid_external_cameras`, with the default inverted because this key turns a protection off rather than on
- `IntentAttestation::PolicyWaived`
- the PAM module skips the prompt when policy waives it
- the daemon honours the waiver only when its own read agrees, and only from a root peer
- `SETUP.md` and `FAQ.md`, including the line that read "This confirmation is mandatory and has no disable setting", plus both the settings and environment tables

## Tests

- `privileged_face_consent_defaults_on_and_env_wins_over_settings`: absent means on, the file turns it off, the env var wins in both directions
- `policy_waiver_is_honoured_only_when_the_daemon_reads_the_same_policy`: refused on policy mismatch, refused from a non-root peer, a real `PamConversation` still passes, and a waived policy still does not accept a missing attestation

`cargo test` clean for `irlume-common`, `irlume-pam` and `irlume-daemon` (117 / 138 / 141). `cargo fmt --all --check` clean. `cargo clippy` adds no new warnings — the two that fire are pre-existing, in `uvc_descriptor.rs:405` and `main.rs:3037`.

## How it was exercised

Running on Arch with irlume 0.11.3 as the face backend for Omarchy's dedicated lock lane, with sudo and polkit wired. With the key set, `sudo -k && sudo true` authenticates by face with no keyword and no password prompt; with the camera covered, the same command falls through to the password prompt as it should. The lock screen is unchanged.

Happy to adjust the naming, the default's shape, or the attestation variant if you would rather express this differently — the part I would argue for keeping is the daemon checking the policy itself rather than trusting the client's word for it.

Context, since it came up in #602 and basecamp/omarchy#7935: I have been testing irlume as the backend for that PR's dedicated face lane, and this was the one part of daily use that consistently grated.

### DIFF
--- a/crates/irlume-common/src/config.rs
+++ b/crates/irlume-common/src/config.rs
@@ -953,10 +953,76 @@ pub fn forbid_external_cameras_visible() -> Option<bool> {
     }
 }
 
+/// Whether privileged services must collect the literal `yes` before a face
+/// attempt (`privileged_face_consent`), or `None` when settings.conf exists and
+/// this process may not read it.
+///
+/// The odd one out among these keys: it defaults **on**, so an absent file or an
+/// absent key means the confirmation is required, and the key exists to turn a
+/// protection off rather than on. That asymmetry is deliberate. No surface
+/// irlume wires runs hands-free by default, and this key does not make one: it
+/// puts privileged prompts on a standing-consent footing as the machine owner's
+/// own choice, distinct from every default surface, without moving that line
+/// for anyone who does not ask (ADR-0018).
+///
+/// Read by both sides. The PAM module skips the prompt, and the daemon consults
+/// it again before honouring [`crate::IntentAttestation::PolicyWaived`], so a client
+/// cannot waive a confirmation the machine's own policy still requires.
+pub fn privileged_face_consent_visible() -> Option<bool> {
+    if let Ok(v) = std::env::var("IRLUME_PRIVILEGED_FACE_CONSENT") {
+        return Some(truthy(&v));
+    }
+    match observe_kv("settings.conf", "privileged_face_consent") {
+        KvObservation::Value(v) => Some(truthy(&v)),
+        // Absent is unambiguous here too, but the default is ON.
+        KvObservation::Absent => Some(true),
+        KvObservation::Unknown(_) => None,
+    }
+}
+
+/// [`privileged_face_consent_visible`] resolved for a decision: unreadable
+/// settings keep the confirmation, because a policy this process cannot read is
+/// not a policy that waived anything.
+#[must_use]
+pub fn privileged_face_consent_required() -> bool {
+    privileged_face_consent_visible().unwrap_or(true)
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::testenv;
+
+    /// The privileged-consent key reads env-over-settings like its neighbours,
+    /// but Absent means ON: it exists to switch a protection off, so anything
+    /// that fails to say otherwise has to leave it standing.
+    #[test]
+    fn privileged_face_consent_defaults_on_and_env_wins_over_settings() {
+        let _g = testenv::lock();
+        let dir = std::env::temp_dir().join(format!("irlume-cfg-consent-{}", std::process::id()));
+        let _ = std::fs::remove_dir_all(&dir);
+        std::fs::create_dir_all(&dir).unwrap();
+        std::env::set_var("IRLUME_CONFIG_DIR", &dir);
+        std::env::remove_var("IRLUME_PRIVILEGED_FACE_CONSENT");
+
+        // Absent key: required, and unambiguously so.
+        assert_eq!(privileged_face_consent_visible(), Some(true));
+        assert!(privileged_face_consent_required());
+
+        // Settings file turns it off.
+        write_kv("settings.conf", "privileged_face_consent", "0").unwrap();
+        assert_eq!(privileged_face_consent_visible(), Some(false));
+        assert!(!privileged_face_consent_required());
+
+        // The env override wins over the file, in both directions.
+        std::env::set_var("IRLUME_PRIVILEGED_FACE_CONSENT", "1");
+        assert_eq!(privileged_face_consent_visible(), Some(true));
+        std::env::set_var("IRLUME_PRIVILEGED_FACE_CONSENT", "no");
+        assert_eq!(privileged_face_consent_visible(), Some(false));
+
+        std::env::remove_var("IRLUME_PRIVILEGED_FACE_CONSENT");
+        let _ = std::fs::remove_dir_all(&dir);
+    }
 
     /// The external-camera prohibition reads env-over-settings with Absent
     /// meaning off, exactly like the biopolicy gate it mirrors.
--- a/crates/irlume-common/src/lib.rs
+++ b/crates/irlume-common/src/lib.rs
@@ -428,6 +428,11 @@ pub fn write_atomic_reporting(
 pub enum IntentAttestation {
     /// A root PAM client reports collecting the conventional confirmation.
     PamConversation,
+    /// A root PAM client reports that this machine's policy waives the
+    /// confirmation (`privileged_face_consent=0`). Never trusted on its own:
+    /// the daemon re-reads the same policy before honouring it, so a client
+    /// cannot waive a confirmation the machine still requires.
+    PolicyWaived,
 }
 
 /// Request from an (untrusted) client to the (privileged) daemon.
--- a/crates/irlume-daemon/src/main.rs
+++ b/crates/irlume-daemon/src/main.rs
@@ -2888,11 +2888,18 @@ fn intent_confirmation_gate(req: &Request, peer: &Peer) -> Option<Response> {
         .as_deref()
         .and_then(irlume_common::pam_service::classify)
         .is_some_and(ServiceKind::requires_face_intent_confirmation);
+    // A waiver is honoured only when the daemon's own read of the policy agrees
+    // with the client's. The client is root, but "root said so" is not the
+    // guarantee here: the guarantee is that the machine's configuration waived
+    // the confirmation, and the daemon is the one that decides it did.
     let has_trusted_confirmation = peer.uid == 0
-        && matches!(
-            intent_confirmation,
-            Some(IntentAttestation::PamConversation)
-        );
+        && match intent_confirmation {
+            Some(IntentAttestation::PamConversation) => true,
+            Some(IntentAttestation::PolicyWaived) => {
+                !irlume_common::config::privileged_face_consent_required()
+            }
+            None => false,
+        };
     let valid = if requires_confirmation {
         has_trusted_confirmation
     } else {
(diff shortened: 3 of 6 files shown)


FizykPSX:
Correcting the record on this PR: the argument I opened it with was wrong.

I claimed irlume already accepts standing consent at the lock screen and was inconsistent in refusing it for privileged services. That hands-free behaviour comes from basecamp/omarchy#7935 — that shell arms a `PamContext` continuously on its own PAM service, so the scan starts because the shell starts it, not because irlume waives anything. Every surface irlume wires itself takes an explicit act. @archledger's framing in #602 is the accurate one: the line is between a shell that took responsibility for arming and everything else.

I have updated the description rather than silently rewriting it, and the same correction is on #602. The commit message carries the same false claim; I will amend it if this branch goes anywhere, and if it does not, the issue thread is the honest record.

What I would still offer for it, independent of the policy question: the daemon-side half. `intent_confirmation_gate` re-reading the policy before honouring a waiver is what keeps any such option from becoming a root-client bypass, and that shape would be worth keeping in whatever form this eventually takes — including the empty-Enter direction, if a waiver ever needs to be expressible at all.

Per #602 I have also confirmed that Omarchy's polkit agent accepts empty submits, which is one of the per-agent data points the empty-Enter direction needs. Happy to rework this branch into that shape, or to close it in favour of your own implementation — your call, no attachment to this patch.

archledger:
@FizykPSX One merge requirement on this PR, separate from the policy question: the commit has no `Signed-off-by` trailer, and this repo requires exactly one per commit (DCO-1.1, see CONTRIBUTING.md). You mentioned you would amend the commit message anyway to correct the retracted standing-consent claim, so the same amend can carry the sign-off: `git commit --amend -s --no-edit` (or `git rebase --signoff`), then force-push.

One more thing so the checks mean something: as a first PR from a fork, GitHub holds your Actions workflows (CI, CodeQL, DCO) in "action required" until a maintainer approves them to run, which is why the checks panel currently only shows the external packit builds. Push the amended commit and I will approve the workflows so the full suite, including the DCO check, actually runs against it.

FizykPSX:
Amended and force-pushed: `9ca99dd`. Ready for the workflow approval whenever suits you.

The amend carries both things at once, as you suggested:

- `Signed-off-by: FizykPSX <86354781+FizykPSX@users.noreply.github.com>` — exactly one trailer, matching the author of the commit.
- The retracted standing-consent claim is gone from the message. Rather than deleting the paragraph I replaced it with the retraction itself, so the commit records that the argument was made and withdrawn, and names #7935 as where the hands-free behaviour actually comes from. What is left as the case for the change is the narrower one: a machine owner removing friction on their own box, default unchanged, no packaged configuration touched. Also downgraded `Closes #602` to `Refs #602`, since the policy question is yours to settle and this branch should not close your issue by merging.

Local state before the push, on Rust 1.98 rather than CI's 1.88: `cargo fmt --all --check` clean, `cargo test -p irlume-common -p irlume-pam -p irlume-daemon` green including both new tests. `cargo clippy -D warnings` fails on `crates/irlume-common/src/secureboot.rs:139` and one other `chunks_exact` site, but that reproduces on unmodified `main` — it is `clippy::chunks_exact_to_as_chunks`, which did not exist in 1.88, so it should not appear on your CI. Flagging it only so it is not a surprise if you ever bump the pinned toolchain; nothing in this branch touches that file.

The policy question is still open and I have no attachment to this shape — if you would rather implement the empty-submit direction yourself, close this and I will happily test it on hardware instead.

FizykPSX:
Force-pushed a one-line fix for the failing check. `9ca99dd` is now `1cc6e50`, and the only difference is:

```diff
-/// it again before honouring [`IntentAttestation::PolicyWaived`], so a client
+/// it again before honouring [`crate::IntentAttestation::PolicyWaived`], so a client
```

`doc (broken intra-doc links are errors)` was the one red check:

```
error: unresolved link to `IntentAttestation::PolicyWaived`
  --> crates/irlume-common/src/config.rs:968:33
      no item named `IntentAttestation` in scope
```

The enum is in `lib.rs`, so it needs a path from inside the `config` module. `crate::` matches what the crate already does at `config.rs:255` with `[`crate::write_0600_atomic`]`. Verified locally with the same command CI runs, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --locked`, plus `cargo fmt --check`.

The push reset the workflow approvals, so `CI`, `dco` and `CodeQL` are sitting at `action_required` on the new head whenever you next get to it. Nothing else in the branch changed.

archledger:
@FizykPSX Decision made, and it is your shape: the typed confirmation stays the default exactly as ADR-0010 and ADR-0011 fixed it, and `privileged_face_consent=0` becomes the owner opt-in for the hands-free experience. ADR-0018 (amending both) is drafted on my side and lands before or with your branch; the empty-Enter direction is withdrawn, so no rework of the mechanism is needed.

What stands between this PR and merge, from my review of 9ca99dd:

1. The retracted standing-consent premise is still in three in-diff doc sites, so the branch documents the feature with an argument you withdrew: the `privileged_face_consent_visible` doc comment in `crates/irlume-common/src/config.rs` ("screen unlock already takes the PAM wiring as standing consent and scans hands-free"), the SETUP.md paragraph ("same footing as the lock screen, where the PAM wiring is itself the consent and the scan runs without a keyword"), and the settings-table phrasing ("starts the way it does at the lock screen"). The honest framing on stock irlume: the greeters and the KDE lock arm on an explicit empty-field Enter, and nothing irlume wires itself is hands-free; the waiver puts privileged prompts on a standing-consent footing as the owner's choice. ADR-0018 carries that wording; aligning these three sites with it closes the finding.
2. The commit message carries an em dash; CONTRIBUTING's plain-punctuation rule asks for none. One character, but it is the repo's stated style.

With those two fixed and ADR-0018 landed, this merges. #602 gets the same answer: the request is accepted in this form, default unchanged, and I will say so there when the ADR is in.

archledger:
@FizykPSX ADR-0018 is accepted and landing (#615), so the decision this implements is on the record: the typed confirmation stays the shipped default, and `privileged_face_consent=0` is the owner-elected exception, daemon-authoritative, fail-closed on unreadable settings.

Two doc sites still carry the withdrawn premise, and both ship in the release docs, so they are the last thing between this PR and merge:

1. The `privileged_face_consent_visible` doc comment in `crates/irlume-common/src/config.rs` still opens with "screen unlock already takes the PAM wiring as standing consent and scans hands-free". On stock irlume no wired surface is hands-free: the greeters and the KDE lock arm on an explicit empty-field Enter, and nothing else scans without an act. ADR-0018 carries the wording that fits: the waiver puts privileged prompts on a standing-consent footing as the machine owner's choice, distinct from every default surface.
2. The settings-table row still says a face attempt "starts the way it does at the lock screen". Same correction: name the waiver itself, not a lock-screen analogy.

With those two aligned this merges; I will turn the merge around immediately when your push lands.

archledger:
@FizykPSX One mechanical note alongside the two doc lines above: I tried to merge and the maintainer-side update-branch commit I created does not trigger your fork's workflows (only the packit builds ran for it), so the required contexts have no runs on that head and the merge cannot proceed. The clean fix is a rebase from your side: git fetch origin && git rebase origin/main && git push --force-with-lease, with the two doc-site fixes from my previous comment folded in. That gives a fork-authored head with full CI (I will re-approve the workflows the moment it lands) and also drops the update-branch merge commit, which the linear-history rule prefers anyway. The moment your checks are green this merges with the ADR-0018 reference in the squash message.

FizykPSX:
Rebased and pushed: `f86af8f` is now `39a4ee7`, sitting directly on `81622f5`. The three maintainer-side update-branch merges are gone, so the head is fork-authored and linear, and the workflows should pick it up.

The doc premise is corrected in four places rather than two. You named the `config.rs` doc comment and the settings-table row; the same lock-screen analogy also appears in the SETUP.md prose above the table and once in FAQ.md, and both ship in the release docs, so leaving them would have left the retracted argument in the same documents.

`crates/irlume-common/src/config.rs`, the doc comment now reads:

> No surface irlume wires runs hands-free by default, and this key does not make one: it puts privileged prompts on a standing-consent footing as the machine owner's own choice, distinct from every default surface, without moving that line for anyone who does not ask (ADR-0018).

The settings-table row and the two prose sites now name the waiver instead of the lock screen:

> `privileged_face_consent=0` is the machine owner's waiver of the literal `yes` on privileged services, so the scan starts when the privileged PAM prompt appears, with no per-attempt word

Diffed against the previous head, the push changes nothing outside those three files: `crates/irlume-common/src/config.rs`, `docs/SETUP.md`, `docs/FAQ.md`.

Verified locally with the commands CI runs: `cargo fmt --check` clean, and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --locked` builds `irlume-common` with no warnings.

archledger (approved):
Verified at 39a4ee79: the two remaining doc sites now state the owner waiver per ADR-0018 (config.rs doc and both settings tables), SETUP.md no longer claims the confirmation has no disable setting, and your commit carries the DCO sign-off. Merging now as promised.

PR #611, opened 2026-08-29
fix: stop the TPM first-use expected miss printing as esys ERROR lines

### LINKED ISSUE #601: tpm2-tss ERROR lines in the journal during a successful first enroll
## What happens

`irlume enroll` succeeds, and while it runs the daemon logs three tpm2-tss lines at WARNING and ERROR level:

```
Aug 29 14:06:33.837314 omarchy irlumed[27019]: WARNING:esys:src/tss2-esys/api/Esys_ReadPublic.c:311:Esys_ReadPublic_Finish() Received TPM Error
Aug 29 14:06:33.837314 omarchy irlumed[27019]: ERROR:esys:src/tss2-esys/esys_tr.c:278:Esys_TR_FromTPMPublic_Finish() Error ReadPublic ErrorCode (0x0000018b)
Aug 29 14:06:33.837314 omarchy irlumed[27019]: ERROR:esys:src/tss2-esys/esys_tr.c:402:Esys_TR_FromTPMPublic() Error TR FromTPMPublic ErrorCode (0x0000018b)
```

Nothing is actually wrong. The enroll completed, and `doctor` reports the result as sealed:

```
[doctor] templates (pawel): ENCRYPTED ✓ · recovery passphrase not set
[doctor] TPM 2.0: /dev/tpmrm0 ✓
```

Face authentication has worked on every attempt since, so this looks like a probe of a persistent handle that does not exist yet on a first enroll, with tpm2-tss printing its own diagnostics straight to stderr at ERROR level on the way past.

## Why it is worth filing

It is only log noise, so this is low priority, but it is log noise of an unhelpful shape. Three lines that say `ERROR` and `Received TPM Error`, emitted during the one operation a new user is least sure about, on a subsystem where a real failure would be serious and silent. Anyone checking `journalctl` after enrolling finds them and cannot tell from the text that the enroll they just did was fine, particularly since the message carries a raw `ErrorCode (0x0000018b)` rather than a description.

If the probe is expected on a first enroll, silencing it — or setting the tpm2-tss log level around that call and reporting the expected miss in irlume's own words — would make the journal say what actually happened. `irlume logs` is the place users are pointed to, and these lines come through it.

## Environment

- irlume 0.11.2, built from the AUR `PKGBUILD` (`irlume 0.11.2-1`) on Arch
- tpm2-tss 4.2.0, kernel 7.1.9-arch1-2, `/dev/tpmrm0`, Secure Boot enabled, UKI boot
- pcrlock not provisioned; seals use the literal PCR-7 policy, per `doctor`
- Logitech BRIO, `/dev/video0` RGB + `/dev/video2` IR

Found while testing irlume against Omarchy's dedicated face-lock PAM lane (basecamp/omarchy#7935).

archledger:
## What

On a fresh install's first enroll, `load_or_create_srk` probes the persistent SRK handle with `tr_from_tpm_public` and `nv_index_handle` probes the pcrlock NV index the same way. The handle does not exist yet - the expected state - and tpm2-tss's own esys logging prints that expected miss as one WARNING and two ERROR lines (`TPM2_RC_HANDLE`, 0x18b) straight into the journal, on the one operation a new user watches closest. Reported with the full mechanism in #601; the reporter's diagnosis (expected miss, not a fault) was confirmed.

## Fix

Both probes now ask `GetCapability` first and only call `tr_from_tpm_public` against a handle the capability list says exists, so the expected miss never reaches the library's error logging:

- the SRK first-use derivation is journaled in irlume's own words (`first TPM use on this machine: deriving and persisting the storage root key`),
- the unprovisioned-pcrlock miss returns a message naming the state (`NV index ... is not defined on this TPM (pcrlock not provisioned)`) instead of a raw handle code.

Suppression was rejected as the mechanism: `TSS2_LOG=esys+none` hides real esys errors too; the pre-check hides only the expected miss (the `TSS2_LOG` syntax itself is now documented in the issue thread).

## Tests

- `capability_list_has_matches_exact_persistent_and_nv_handles`: exact-handle match, the empty-TPM first-use state, and the cross-type miss; mutation-verified (breaking the matcher fails the test)
- `cargo test --workspace`: 1810 passed, 0 failed; clippy `-D warnings` clean; fmt; zero em dashes

## Hardware evidence (local TPM, branch daemon)

`irlume diag` through the new probe path: both the keyring and template seals unseal clean, **zero esys lines in the entire session journal** (previously the miss paths printed the three lines). The fresh-install first-use path runs in the CI hardware lane (real TPM).

Fixes #601

### DIFF
--- a/crates/irlume-core/src/tpm.rs
+++ b/crates/irlume-core/src/tpm.rs
@@ -31,7 +31,7 @@ use std::str::FromStr;
 use zeroize::Zeroizing;
 
 use tss_esapi::attributes::{ObjectAttributesBuilder, SessionAttributesBuilder};
-use tss_esapi::constants::tss::{TPM2_HT_NV_INDEX, TPM2_HT_PERSISTENT};
+use tss_esapi::constants::tss::{TPM2_HR_NV_INDEX, TPM2_HR_PERSISTENT};
 use tss_esapi::constants::{CapabilityType, SessionType};
 use tss_esapi::handles::{
     KeyHandle, NvIndexHandle, NvIndexTpmHandle, ObjectHandle, PersistentTpmHandle, SessionHandle,
@@ -328,7 +328,7 @@ fn load_or_create_srk(ctx: &mut Context) -> Result<(KeyHandle, bool)> {
     // missing handle is the EXPECTED state on a fresh install's first enroll.
     // Ask GetCapability first, so the expected miss never prints as an error;
     // the probe below then only runs against a handle that is really there.
-    if tpm_handle_exists(ctx, u32::from(TPM2_HT_PERSISTENT), wanted)? {
+    if tpm_handle_exists(ctx, TPM2_HR_PERSISTENT, wanted)? {
         if let Ok(object) = ctx.tr_from_tpm_public(wanted) {
             let key_handle = KeyHandle::from(ESYS_TR::from(object));
             let is_ours = match ctx.read_public(key_handle) {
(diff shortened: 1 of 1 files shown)


archledger:
Correction to the hardware-evidence claim in the description: the local verification run used a daemon binary built before the change, so it exercised none of it. The first cut also passed the handle-type constants where GetCapability wants the handle-range values, which is what broke both CI TPM lanes (empty capability list, existing SRK seen as absent, evict_control hitting TPM2_RC_NV_DEFINED). Both fixed in e6a743aa and re-verified against a binary confirmed to contain the change: existing-SRK path unseals both seals, zero esys lines, first-use notice correctly silent.

PR #643, opened 2026-08-31
fix: explain persistent IR saturation denials

### LINKED ISSUE #636: A saturated IR source inside the frame makes the auth path refuse a correctly illuminated face as too dark
## Summary

On this camera, a saturated IR source inside the frame makes the authentication path judge a correctly illuminated face as too dark and refuse it as a spoof. Rotating the laptop so the source leaves the frame, with nothing else in the room changed, takes the same face from refused to granted. irlume's own `ambient` reading barely moves across that change, so the deciding factor is not how much infrared is in the room; it is that a blown-out region is inside the frame.

The source here is a north-facing window with its blinds closed. In visible light the room is unremarkable. In infrared the window saturates completely.

## The pair

Same room, same lamps, same daylight, same distance, about a minute apart. The only change is the laptop's angle.

| | window in frame | window out of frame |
| --- | --- | --- |
| `ir_bright` (face region) | 17 to 29 | **123** |
| `ambient` | 28 to 34 | 22 |
| `ir_center_edge_ratio` | 1.43 to 1.49 | 1.29 |
| whole-frame mean, illuminated / non-illuminated | 109.5 / 91.7 | 51.9 / 28.4 |
| verdict | `Spoof: IR face too dark`, `score=0.0000` | `granted=true score=0.905` |

Refused:

```
liveness(cross-spectrum): Spoof (IR face too dark (29); not reflecting IR like skin);
ir_bright=29 ir_center_edge_ratio=1.43 glint=252.00 ambient=34 face_frac=0.193 ir_clipped=0.0%
face auth 'pawel': granted=false live=false score=0.0000
```

Granted:

```
ir_bright=123 ir_center_edge_ratio=1.29 glint=254.00 ambient=22 face_frac=0.214 ir_clipped=0.0%
face auth 'pawel': granted=true live=true score=0.905
```

Note the direction of the whole-frame numbers: the frame is roughly twice as bright in the case that is refused for darkness.

## Raw frames

Captured straight from `/dev/video2` with ffmpeg, no irlume involved, after applying the emitter control. Consecutive frames, so an illuminated one and its neighbour.

Window in frame, illuminated frame. The face is lit; the window is pure white:

window in frame, illuminated

The very next frame, non-illuminated. The face is nearly black; the window is unchanged:

window in frame, non-illuminated

Window out of frame, illuminated. The face is now the brightest thing in the picture:

window out of frame, illuminated

And its neighbour:

window out of frame, non-illuminated

## What the emitter is actually doing

Worth stating because it rules out the obvious explanation. The emitter works throughout, and it works with nobody in front of the camera. Centre-crop means of 12 consecutive frames, nobody in frame, dark room:

```
without the control:  0 0 0 0 0 0 0 0 0 0 0 0
with the control:     30.4  0.0004  44.1  0.0004  59.1  0.0009  58.4  0.0008  57.5  0.0005  0.03  0.0006
```

Two things fall out of that. Illumination alternates frame by frame, and one write of the control buys roughly five illuminated frames before it stops, which lines up with the daemon's own `illumination metadata closed after 10 classified frames`.

The same crop with a face present:

| condition | illuminated frames | non-illuminated frames | irlume's `ir_bright` |
| --- | --- | --- | --- |
| dark room | 101 to 108 | 0 | 138 |
| window in frame | 40 to 60 | 12 to 18 | 17 |

In the dark, the reported value sits at the illuminated level. With the window in frame it sits at the non-illuminated level, while illuminated frames of the same burst are three to four times brighter. I cannot see from outside which frame the pipeline ends up scoring, so I am reporting the correspondence rather than claiming the mechanism.

## The advice loops

The refusal, and the enrollment refusal from #621, both say to check the emitter and run `sudo irlume ir-setup`. On this machine that had been done minutes earlier and reported success:

```
[ir-setup] IR emitter enabled: 14:6:1,3,2,0,0,0,0,0,0 on the camera's Microsoft camera-control
unit ... accepted: interface 0x03 set to D1 (alternative illumination)
```

So the operator is told to apply a remedy that is already applied, gets the same refusal, and has no way to reach the real one, which is to keep a bright infrared source out of the frame. This is the same shape as #613, where the advice pointed at an emitter that was already working, except that #624 fixed the enrollment preflight and this is the authentication path.

That may be the residual you already named when closing #613, the burst diagnostic still judging by whole-frame means. The numbers above fit it: the frame that is refused for darkness is the brighter frame.

## Suggested handling

Nothing here needs a new gate as far as I can tell. What would have saved the hour:

- when the face region is dark but the frame contains a saturated region, say so, rather than pointing at the emitter
- treat a large clipped region as its own situation, the way `ir_clipped` is already tracked per attempt
- do not repeat the `ir-setup` remedy when a control is in force and was applied during this capture

## Environment

ThinkPad T14s Gen 3, Arch, irlume built from `main` at `27c2a7b`. Chicony `04f2:b74f`, RGB interface 0, IR interface 2, same USB device. Stored verdict `sequential_required / signal_loss`. Emitter control `14:6:1,3,2,0,0,0,0,0,0`, applied by both `irlume ir-setup` and `linux-enable-ir-emitter`, with identical results.

Raw frames, `camera-tune` records or a support report available on request; the four frames above are on a branch of my fork if you want them at full size.

archledger:
## What & why

Fixes #636.

A localized IR-bright source can remain clipped in both emitter-lit and metadata-dark frames, pin camera exposure, and leave the face dark or flat. The existing attempt diagnosis then recommends emitter setup even though the emitter is working.

This change records whole-frame clipping for the selected metadata-lit frame, its adjacent metadata-dark partner, and the exact same-pixel overlap. Persistent clipping only selects a more accurate denial reason and stable `IR source` situation. It does not change liveness verdicts, grants, scores, retryability, existing cue thresholds, or single-frame enrollment preflight behavior.

PAM now recommends repositioning or password fallback rather than `ir-setup` for this situation.

## Testing

- [x] `cargo test --workspace` passes, 1,881 passed, 0 failed, 100 environment-gated ignored
- [ ] Docs/CHANGELOG updated if user-facing
- [x] Hardware-validated (if it touches PAM, the daemon, or capture)

Additional evidence:

- Focused camera, liveness, auth, and PAM suites: 807 passed, 0 failed.
- Five mutation probes detected changes to the strict boundary, dark metadata requirement, exact overlap, auth precedence, and PAM mapping.
- The production helper matched all six public issue frames. ThinkPad window-in overlap was 0.16487 at the GREY8 ceiling; window-out was 0.00067; BRIO overlap was 0.0.
- A sandboxed ASUS RGB+IR capture exercised the ordinary no-face path without false `IR source` wording, then production state was restored and revalidated.
- Final metadata-complete ASUS hardware validation measured lit saturation 0.47828, dark saturation 0.46420, and exact persistent same-pixel overlap 0.45612. The centered face remained detected, the existing cue returned `Spoof`, and the exact `AuthResult` wire carried `IR source` with `granted=false` and `live=false`.

Two independent implementation reviews found no Critical, Important, or Minor findings. A final pre-merge differential review covered all seven changed files, found no Critical or Important issues, recorded two non-blocking test-hardening suggestions, and recommended approval.

### DIFF
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -331,6 +331,7 @@ enum AttemptSituation {
     TooFar,
     OffCenter,
     LookingAway,
+    IrSource,
     TooDark,
     GlintBelow,
     BelowScore,
@@ -347,6 +348,7 @@ const fn attempt_situation_label(situation: AttemptSituation) -> &'static str {
         AttemptSituation::TooFar => "too far",
         AttemptSituation::OffCenter => "off-center",
         AttemptSituation::LookingAway => "looking away",
+        AttemptSituation::IrSource => "IR source",
         AttemptSituation::TooDark => "too dark",
         AttemptSituation::GlintBelow => "glint below",
         AttemptSituation::BelowScore => "below score",
@@ -367,6 +369,7 @@ struct AttemptFacts {
     rgb_face_brightness: f32,
     glint: Option<f32>,
     ir_bright: f32,
+    persistent_ir_source_overwhelms: bool,
 }
 
 impl AttemptFacts {
@@ -378,6 +381,7 @@ impl AttemptFacts {
             rgb_face_brightness: a.signals.rgb_face_brightness,
             glint: a.signals.ir_eye_glint,
             ir_bright: a.ir_brightness,
+            persistent_ir_source_overwhelms: a.signals.persistent_ir_source_overwhelms(),
         }
     }
 }
@@ -407,6 +411,9 @@ fn auth_attempt_situation(kind: OutcomeKind, f: &AttemptFacts) -> AttemptSituati
     }
     if f.yaw_asym > FRAME_YAW_ASYM_MAX {
         return AttemptSituation::LookingAway;
+    }
+    if f.persistent_ir_source_overwhelms {
+        return AttemptSituation::IrSource;
     }
     if f.rgb_face.is_some() && f.rgb_face_brightness < DIM {
         return AttemptSituation::TooDark;
@@ -3573,6 +3580,7 @@ impl Engine {
             face_frac: face_frac_of(rgb_top.as_ref().map(|f| &f.bbox), rgb.width),
             // RGB-only path: no IR frame exists to clip.
             ir_saturated_frac: None,
+            ir_persistent_saturated_frac: None,
             ir_ceiling_known: false,
             rgb_face_brightness: rgb_brightness,
             rgb_specular_frac: rgb_specular,
@@ -4367,6 +4375,7 @@ impl Engine {
                 ir_top.as_ref().map(|f| &f.bbox),
                 ir_stats.white_level,
             ),
+            ir_persistent_saturated_frac: ir_stats.persistent_saturated_frac,
             // Whether the FORMAT could be measured, which is not the same
             // question as whether this capture produced a number: the call
             // above also yields None when no face was found (#358).
@@ -11192,6 +11201,7 @@ mod engine_tests {
             rgb_face_brightness: 120.0,
             glint: Some(250.0),
             ir_bright: 140.0,
+            persistent_ir_source_overwhelms: false,
         };
         let line = |kind, score, facts: &AttemptFacts| {
             let text = attempt_situation_line(kind, score, facts);
@@ -11279,6 +11289,7 @@ mod engine_tests {
             rgb_face_brightness: 0.0,
             glint: None,
             ir_bright: 140.0,
+            persistent_ir_source_overwhelms: false,
         };
         assert_eq!(
             attempt_situation_line(OutcomeKind::NoFace, 0.0, &facts),
@@ -11298,6 +11309,7 @@ mod engine_tests {
             rgb_face_brightness: 120.0,
             glint: Some(72.0),
             ir_bright: 140.0,
+            persistent_ir_source_overwhelms: false,
         };
         assert_eq!(
             auth_attempt_situation(OutcomeKind::Spoof, &turned),
@@ -11383,6 +11395,7 @@ mod engine_tests {
                 ir_ambient: 30.0,
                 face_frac: 0.22,
                 ir_saturated_frac: None,
+                ir_persistent_saturated_frac: None,
                 ir_ceiling_known: false,
                 rgb_face_brightness: 90.0,
                 rgb_moire_score: 0.0,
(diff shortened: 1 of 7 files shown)


archledger:
Hardware session 1, ASUS FHD pair (RGB /dev/video0, IR /dev/video2):

- Installed-production baseline with the user upright at normal distance granted profile BEN at score 0.746.
- A second production attempt with the ceiling LED in the background also granted at 0.745.
- Branch direct-probe control with the user upright measured lit whole-frame saturation 0.000949, ambient 0.0, persistent overlap 0.0. The face read brightness 137, center/edge 1.48, glint 241, and the liveness verdict remained Live.
- The only available source tonight was a visible ceiling LED, not a continuous IR source. It therefore did not reproduce issue #636. Earlier source positions likewise produced no persistent clipping.
- Temporary developer-probe instrumentation was removed. The committed tree is clean, the focused CLI verification passed 591 tests with 0 failures and 11 declared ignores, production service/socket remain active and enabled, the camera pair and enrollment are unchanged, and the serving daemon hash still matches the installed binary.

The live positive hardware checkbox remains open. Positive live validation needs daylight through the window or a dedicated continuous IR illuminator that produces more than 10% same-pixel clipping in both metadata phases, without direct eye exposure.

archledger:
Offline historical-hardware replay from /home/wisbfime/irlume-suncal:

- Source: raw GREY8 frames captured outdoors on 2026-07-16 with the same ASUS FHD IR camera pair documented in SESSION-2026-07-16.md.
- The historical burst_dump format retained pixels, means, and dequeue timing, but not V4L2 illumination metadata. Therefore this is an offline hardware replay, not a metadata-complete live production reproduction.
- The standing-outdoor burst has a stable even-lit/odd-dark cadence in all 8 adjacent pairs. Reconstructing only those phase labels and running the checked-in production selection and paired-saturation helper selected frame 06 with adjacent dark frame 07.
- Production-helper result at the GREY8 ceiling 255: lit saturation 0.7464453, ambient saturation 0.7400742, persistent same-pixel overlap 0.7334766. This passes the strict persistent > 0.10 diagnostic threshold by a wide margin.
- Independent read-only recomputation across all 8 adjacent pairs found persistent overlap from 0.73348 to 0.74006, so the result is not dependent on one selected pair. A separate laptop-outside burst exceeded 0.10 in 7 of 8 pairs; its first warm-up pair measured 0.09066.
- The temporary local ignored test used to call the private production helper was removed after execution; the committed worktree is clean. No frames or biometric payloads were copied or attached.

Conclusion: the corpus supplies strong positive real-sensor pixel evidence for the new overlap calculation and threshold. It does not close the live metadata hardware checkbox because the 0.2.1 capture utility did not persist the camera phase flags. The PR should remain draft until a daylight or dedicated-IR live run records explicit Lit/Dark metadata.

archledger:
Final metadata-complete hardware validation, ASUS internal RGB+IR pair:

- Ran the branch daemon with every mutable input and socket redirected into an isolated sandbox, using byte-identical copies of production state. A normal-user `irlume auth test --events=jsonl` exercised the exact daemon and `AuthResult` wire path without releasing credentials.
- The decisive final attempt measured lit saturation `0.47828`, dark saturation `0.46420`, and exact persistent same-pixel overlap `0.45612`.
- The face remained detected and centered (`face_frac 0.227`, yaw `0.11`).
- The existing dark or flat cue returned `Spoof`; the persistent-source selector chose `IR source`.
- The attempt journal recorded `IR source`, and the exact `AuthResult` wire carried situation `IR source` with `granted=false` and `live=false`.
- A preceding high-clipping attempt selected the higher-priority `looking away` situation. Earlier centered captures above the clipping threshold remained `Live` with an empty situation when no independent dark or flat denial existed. These controls confirm that clipping alone does not deny and that framing precedence is preserved.
- Temporary instrumentation and the sandbox were removed. Focused verification passed 15 of 15 tests with 2 declared environment-gated ignores. The branch worktree is clean.
- Production service and socket are active and enabled on their original installed hashes. Enrollment, emitter configuration, settings, template key, and capture qualification records were byte-identical after restoration.
- No frames, identities, biometric payloads, credentials, or raw capture artifacts were retained or uploaded.

The live hardware gate is complete.

PR #694, opened 2026-09-11
Improve status responsiveness, model ownership, and performance benchmarks

archledger:
Routine status polling could run live PCR diagnosis, wait through several independent daemon timeouts, and block TUI startup on external probes. This change separates metadata from explicit diagnostics, shares a two-second observation deadline in machine status, and moves TUI observations into background refreshes while preserving unknown readings and historical diagnostic freshness.

Model loading now parses the PAD bytes accepted by checksum policy, releases serialized recognizer weights before auxiliary sessions, and transfers owned native-mesh bytes into TFLite. Authentication gates, model pins, degraded-policy decisions and on-disk formats remain unchanged. Benchmark examples now execute current production wrappers, reject failed samples, consume results, and label construction, synthetic stages and live-engine timing accurately.

The commits separate model ownership, observation primitives, status/TUI behavior and benchmark corrections. Contract 1 retains existing field types and adds optional camera/fingerprint observation-known fields. Older daemons use the cheap armed-bit fallback.

Validation: guarded workspace suite **2,244 passed**; Rust 1.88 all-target Clippy with warnings denied; rustfmt; warnings-denied workspace docs; Rust 1.88 optimized workspace build. Targeted native TFLite ownership/landmark tests passed with both exact names enforced by the existing CI guard. All 18 benchmark example test executions and synthetic runtime/error-path smokes passed. Independent review findings for stale diagnostic wording and test-global PATH interference were fixed and verified.

The CI follow-up declares the new observation fields as optional booleans in the published schema and retains legacy fixtures. Camera census already documents canonical video nodes in its node field; the conformance checker now recognizes only that exact field while continuing to reject paths elsewhere. Regression tests cover optional boolean types, undocumented fields, census scope, malformed nodes and error responses. The strict conformance step and full main CI job now pass.


archledger:
The ASUS comparison supports a **17.5% reduction in startup peak RSS** for the default full-model configuration and faster repeated status queries for an enrolled, armed account. It does **not** demonstrate a substantive authentication speedup. All 20 measured attended authentications succeeded; cancellation and look-away refusal also passed on both versions.

Measurements were collected on 2026-09-11. The baseline is `6049fba7ae0c6d8fdc8f6cf6dbaf468d805aa9a8`; the candidate is `d82b8de0aa6904bea0acc35d05da1469ec6a84aa`. Both use Rust 1.88.0, locked dependencies, the same default release profile/features, and four build jobs. The measured candidate binaries were originally built at `0f97f7ef`; the subsequent schema/conformance correction changed no production Rust source, dependency lockfile, build configuration, or model manifest. A fresh release build at the final candidate produced CLI and daemon SHA-256 hashes identical to the frozen measured binaries.

The machine was an ASUS UX5406S running Fedora 44 and Linux `7.1.13-200.fc44.x86_64`. Both versions used the same hash-verified public model files, installed ONNX Runtime 1.28.1, and native TFLite library. AC power, the balanced platform profile, powersave governor, and `balance_power` EPP were preserved. Builds finished before measurements. The desktop and normal system workload remained active, so these are controlled comparisons on one working laptop rather than a quiet laboratory benchmark.

The startup campaign used empty private state, the default dual-camera configuration, and the full model stack. Both PAD sessions and native mesh were verified ready. One warmup per version preceded five paired comparisons in AB, BA, AB, BA, AB order. Each start included four production JSON status commands: 12 successful starts and 48 successful commands overall, with warmups excluded from these medians.

| Startup/default empty-state metric | Baseline median | Candidate median |
|---|---:|---:|
| Peak RSS | 1499.5 MiB | 1237.4 MiB |
| Sampled peak PSS | 1485.1 MiB | 1221.0 MiB |
| RSS after two seconds idle | 923.9 MiB | 910.6 MiB |
| Process start to model-ready response | 2005.4 ms | 1904.6 ms |
| Ready JSON status command | 16.0 ms | 19.2 ms |

Peak RSS fell in every pair, by 246.7–266.4 MiB. The difference between the two medians is **262.08 MiB, or 17.48%**; the median of paired reductions is 257.4 MiB. These are different statistics. Idle-memory ranges overlap, so the strongest finding concerns startup peak memory. This is consistent with earlier release of serialized model buffers, without attributing the entire reduction to one change.

RSS peak uses the kernel high-water mark; PSS was sampled every 50 ms and can miss an intervening peak. Starts were process-cold with warmed filesystem caches. There was no cache eviction or cold boot, and the installed daemon remained running during this campaign, potentially affecting shared-library PSS. Model readiness was polled approximately every 20 ms.

Startup was faster in three pairs and slower in two. Its paired median improvement was only 30.9 ms, and maxima were essentially identical: 2164.3 versus 2165.3 ms. Empty-state status was slower on the candidate. Its 136.6 ms maximum remains in the data; a 138.8 ms baseline warmup observation is also retained. Neither outlier was diagnosed or discarded. The 20 measured status calls per version are clustered within five processes, not 20 independent startup trials.

The separate attended campaign used the existing enrollment and installed **IR-only experimental** policy, with unchanged camera/model settings. Each fresh daemon completed two identity-only authentication requests after model readiness, followed by four status commands. Five alternating pairs yielded 20/20 successful authentications with camera release; four successful pilot requests are excluded from the statistics.

| Enrolled-account metric | Baseline median | Candidate median | Samples per version |
|---|---:|---:|---:|
| First authentication after readiness | 4322.8 ms | 4337.7 ms | 5 |
| Second, warm authentication | 4182.0 ms | 4171.0 ms | 5 |
| First post-auth status | 298.3 ms | 285.9 ms | 5 |
| Later status commands | 26.20 ms | 19.42 ms | 15, within five processes |

First and warm authentication had paired median changes of +13.0 ms and −7.7 ms, respectively: no substantive demonstrated speedup. First-status differences were variable, with a candidate maximum of 418.4 ms versus 393.3 ms, so no reliable first-status gain is claimed. For later status, all five per-process medians improved, with paired reductions of 4.65–7.06 ms and a median reduction of 6.67 ms. No TPM-command counts were measured. These small samples do not estimate population tail latency reliably.

One cancellation per version confirmed an open camera before disconnect and subsequent release. Disconnect-to-confirmed-quiet took 531.1 ms and 550.3 ms, including a fixed 200 ms quiet-confirmation interval; this is functional evidence, not a speed comparison. One attended look-away test per version returned a no-face refusal in 11.540 s and 11.530 s, released the camera, and allowed subsequent management queries.

The initial pilot reported a restoration warning because its executable check raced a `Type=simple` service start. Independent verification confirmed the expected installed executable and readiness. The corrected harness waits for both, with a bounded failure if restoration remains wrong; nine harness tests passed. The original pilot record was preserved. Subsequent campaigns completed restoration successfully. Final verification confirmed the original daemon/service state, package integrity, all 26 protected files unchanged, and no remaining camera holders, test mounts, private runtimes, or pending emitter records. No raw frames, biometric scores, templates, or credentials were retained in the measurement evidence.

These tests exclude PAM/desktop login, wallet release, reboot/suspend, spoof evaluation, new enrollment, power consumption, and full nightly qualification. The next optimization should instrument capture, liveness, and matching to explain the roughly 4.2-second authentication cost before changing policy. Retain the measured model-memory improvement and investigate empty-state status overhead with comparable measurements.

Issue #606, opened 2026-08-29
Concurrent capture refused on a BRIO behind chained hubs, where the same model qualified on another host


FizykPSX:
As requested in basecamp/omarchy#7935: the concurrent-capture refusal on a Logitech BRIO, on a host where you qualified the same camera model successfully.

## Verdict here

```
$ sudo irlume camera-tune
capture mode sequential for this camera: it cannot stream RGB and IR at once
(all 6 concurrent attempts errored; 6 sequential round(s) completed, 0 errored;
 a trailing one-at-a-time control confirmed the camera still answers)
```

Reproduced on 0.11.2 and again on 0.11.3, identical wording both times. Worth noting for your per-round diagnostics work: all six attempts **errored** rather than failing the frame-provenance bar, so the verdict names no continuity fact — the failure is at open/stream level, before continuity can be measured. The #586 per-fact reporting does not reach this case.

## The likeliest difference from your host

The BRIO here is not on a root port. It is two hub levels down:

```
$ lsusb -t
|__ Port 002: Dev 005, If 0, Class=Hub, Driver=hub/4p, 5000M
    |__ Port 002: Dev 006, If 0, Class=Video, Driver=uvcvideo, 5000M
    |__ Port 002: Dev 006, If 1, Class=Video, Driver=uvcvideo, 5000M
    |__ Port 002: Dev 006, If 2, Class=Video, Driver=uvcvideo, 5000M
    |__ Port 002: Dev 006, If 5, Class=Human Interface Device, Driver=usbhid, 5000M
```

and your own qualification context records the path:

```
"usb_devpath": ".../usb4/4-2/4-2.2/4-2.2.2"
```

`4-2/4-2.2/4-2.2.2` is a chained-hub topology at 5000M. If your qualifying host had the BRIO on a root port or a single hub, that is the first variable I would test, since concurrent RGB+IR on one UVC device is exactly where bandwidth allocation and transfer scheduling across chained hubs would bite while sequential capture stays fine.

I have not verified this by moving the camera — say the word and I will replug it directly into a root port and re-run `camera-tune`, which would turn the hypothesis into a measurement on identical hardware and software.

## Host

- irlume 0.11.3, Arch, kernel 7.1.9-arch1-2, uvcvideo 1.1.1
- Logitech BRIO `046d:085e`, `/dev/video0` RGB (YUYV/MJPG/NV12), `/dev/video2` IR (GREY only), `/dev/video1` metadata-only
[...]

PR #703, opened 2026-09-13
feat(auth): let privileged services collect grouped PAD evidence

FizykPSX:
## What & why

On a camera pair that cannot capture RGB and IR concurrently, `sudo`, `su`, `doas` and polkit cannot finish a face authentication within any budget an operator would want to sit through. This adds an owner opt-in that lets those services use the same bounded sequential PAD collection the greeter and lock screen already use.

`qualify_rgb_pad_evidence` needs `VIT_PAD_VOTE_N` = 5 ViT scores, and an attempt scores exactly one RGB frame (the final one). The ordinary retry loop **does** accumulate those votes across attempts when the budget holds five of them — `pending_pad_retry_loop_completes_five_assessments_when_they_fit` is the case that works. What does not work is the arithmetic on a pair whose attempt costs seconds: the privileged default is 5000 ms, one attempt here measured 4.4-4.9 s, so the ring never fills. `grouped_auth` sidesteps it by gathering the whole vote window inside one transaction at one attempt's cost, which is why the lock screen grants on this hardware — but `eligible_configuration` scoped that collector to `ServiceKind::Greeter | ScreenUnlock`, with `AppConsent => false`.

Measured on the reporting host (Logitech BRIO, `camera-tune` qualifies only `MeasuredSequential` — the pair from #606 that you reproduced on a direct root port):

- `sudo` / polkit, stock 0.12.0: `granted=false live=false score=0.0000 (collecting RGB PAD evidence)`, one attempt of 4930 ms, every time.
- `irlume identify`, seconds earlier on the same face: `score 0.961` ✅
- Lock screen (`omarchy-lock-face`), same daemon and minute: `granted=true live=true score=0.938`, 7251 ms.
- With `IRLUME_GRACE_MS=15000`: polkit ran three full attempts (13694 ms) and cast three votes (`p_spoof` 0.389 / 0.380 / 0.393, all far below `VIT_PAD_THRESHOLD`), then settled — five attempts of that cost need ~22 s, which is not a budget worth shipping.
- `pad_vit=0` is not a workaround either: it turns the deferred refusal into a hard `RGB PAD is unavailable`.

With the key on, same host: polkit `granted=true live=true score=0.857` in 7262 ms with five `pad-vit` votes logged `from grouped capture`, and `sudo` `granted=true score=0.868` with the session opened and no password.

## What this does and does not change

`privileged_grouped_pad_evidence=1` in `/etc/irlume/settings.conf`, or `IRLUME_PRIVILEGED_GROUPED_PAD=1`, opts `Elevation` into the `Verify` arm and lifts the hard `false` on `AppConsent`. It defaults **off**, and reads off when `settings.conf` exists but cannot be read, so privileged surfaces are untouched for anyone who does not ask.

It widens **which services may collect** the evidence, never how much evidence a grant needs. `VIT_PAD_VOTE_N`, `VIT_PAD_THRESHOLD`, the IR PAD thresholds, liveness and the identity thresholds are unchanged, and a privileged grant still closes the whole five-vote window. `CredentialRelease` deliberately keeps its existing rule — an owner preference is not a reason to let a sealed secret leave the recognized local login and lock services.

**The budget follows the route, not the service name** (review round 2). The collector is gated on `window >= GRACE_WINDOW_MS`, so the key needs a longer privileged budget to do anything — but deciding that in `grace_window_ms` was too early: the capture route is unknown there, so a concurrent pair, an unqualified or runtime-demoted one, a models-missing request and experimental IR-only all took the longer failure budget without ever reaching the collector. IR-only is the clearest case, since it returns on its own route in `authenticate_in_window_inner` before grouped eligibility is consulted.

`grace_window_ms` is therefore back to its service table, and `privileged_budget_for_route` decides at the point where the mode, its qualification and runtime standing, the models and the purpose are all known. Only the **default** short window is replaced, the replacement is measured from the request's entry instant (`deadline - window`, never from now) so the request keeps exactly one deadline with time already spent still counted, and an explicit `IRLUME_GRACE_MS` still decides the budget alone — including a value below the login window and the legacy one-shot zero. Exact runtime authority stays in the outer gate, as in `eligible`.

## Testing

- [x] `cargo test -p irlume-common -p irlume-auth` passes (227 + 146).
- [x] New coverage: config default and env precedence; scope widening with the credential-release exclusion held; the budget replacement against each non-grouped route with the flag still on (concurrent, unqualified, demoted, models absent, IR-only's budget half); default-window-only and override-wins; and the newly eligible privileged services driven through the collector itself — complete evidence granting, incomplete evidence never reaching identity, and a deadline or cancellation terminal with the evidence discarded.
- [x] `cargo fmt --all --check` clean. Clippy clean on the touched code; the two `chunks_exact_to_as_chunks` errors in `diagnostics.rs` and `secureboot.rs` come from a newer clippy than the pinned 1.88 and predate this branch.
- [x] DCO sign-off added.
- [x] Hardware: re-verified on the sequential BRIO pair after the review changes — `sudo` and polkit granting, greeter and lock unaffected.
- [ ] Not verified on a pair that qualifies concurrent capture: the key changes nothing I can exercise there, since those hosts complete the vote window in one session already.
- [x] Docs: SETUP.md configuration reference and a diagnosis section keyed on the `collecting RGB PAD evidence` symptom, plus CHANGELOG — both carrying the narrower rationale.

If you would still rather size the privileged window from the measured attempt cost (the engine tracks `costliest_attempt` and knows the pair is sequential from `qualification-v2`), or collect the vote window sequentially inside one attempt so no key is needed at all, say so and I will rework it that way — this shape is the smallest change I could justify, not a preference.

No ADR is included — the scope of a PAD collection feels like your call to record, and I did not want to presume ADR-0018's shape. Happy to write one if you want it here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


crates/irlume-common/src/config.rs
```diff
@@ -562,6 +562,37 @@ pub fn privileged_face_consent_required() -> bool {
...
+/// run the bounded sequential PAD collection the greeter and lock screen already
+/// use (`privileged_grouped_pad_evidence`).
+///
+/// Defaults **off**, and an unreadable settings.conf reads as off, so without the
+/// key privileged surfaces behave exactly as they do upstream.
+///
+/// It exists for a camera pair that cannot capture RGB and IR concurrently. There
+/// one authentication attempt scores exactly one RGB frame, so it casts one ViT
+/// vote, and `VIT_PAD_VOTE_N` votes never accumulate inside a privileged window:
+/// every attempt settles as `RgbPadPending` however long that window is. The
```
archledger: Please qualify the 'however long' claim. The vote ring is cleared at request entry, while `authentication_attempt_loop_with` can retry healthy pending evidence until another observed-cost attempt no longer fits. `pending_pad_retry_loop_completes_five_assessments_when_they_fit` demonstrates five ordinary assessments completing within a sufficient budget, and the explicit override accepts up to 60 seconds. The 15-second experiment establishes an insufficient budget for that host's measured attempt cost, not an inability to accumulate votes at any window length. The efficiency benefit of grouped capture remains a good motivation. Please align the other documentation and PR description with that narrower claim.

crates/irlume-auth/src/lib.rs
```diff
@@ -929,7 +929,20 @@ fn grace_window_ms(service: Option<&str>) -> u64 {
...
+        // window in one capture session. A pair that can only capture
+        // sequentially casts one vote per attempt, so the owner who opts into
+        // `privileged_grouped_pad_evidence` needs the grouped collector, and that
+        // collector is itself gated on `window >= GRACE_WINDOW_MS`: leaving the
+        // short window here would make the opt-in a no-op. The cost is the one
+        // #362 measured in the other direction — a refused privileged attempt now
+        // holds the camera for the login window before the password prompt.
+        Some(kind)
+            if kind.wants_short_grace()
+                && !irlume_common::config::privileged_grouped_pad_evidence_enabled() =>
```
archledger: This flag changes every short privileged window from 5 to 15 seconds before collector eligibility is known. That includes concurrent pairs, unqualified or runtime-demoted pairs, and experimental IR-only requests, which take their separate route before `grouped_auth::eligible`. Those requests get the longer failure/fallback budget without this collector's benefit. Please retain the short default for requests that cannot use grouped collection, while preserving the explicit numeric override and the request's deadline contract. Add coverage with the flag enabled for eligible sequential and non-grouped configurations.

archledger (changes requested):
Thanks for the focused change and the hardware report. Reusing the existing grouped collector is a reasonable approach here. I reviewed 0c77ce88 and ran the scoped checks locally on Rust 1.98.1: 225 authentication tests passed (3 existing ignored), 146 common-library tests passed, formatting passed, and Clippy for both crates with all targets and warnings denied passed. The existing five-vote, PAD/identity threshold, runtime qualification, credential-release scope and separate privileged-consent checks are preserved.

Before merge, please address these items:

1. **DCO:** the commit currently has a Co-Authored-By trailer but no Signed-off-by. CONTRIBUTING.md and the DCO workflow require exactly one Signed-off-by per commit. Please add your own sign-off.
2. **Rationale:** correct the claim that ordinary sequential authentication cannot finish at any window length. Healthy retries can accumulate votes within one authentication request; the existing `pending_pad_retry_loop_completes_five_assessments_when_they_fit` test passes. The reported default/15-second budgets were insufficient for the measured attempt cost, which still motivates the more efficient grouped path. Please update the config documentation, setup guide, changelog and PR description consistently.
3. **Timeout scope:** keep the longer default privileged budget tied to grouped-collection eligibility. Currently the flag also lengthens requests on concurrent, unqualified/demoted and experimental IR-only paths, where this collector is not used. Details are inline. Please include regression coverage for those non-grouped cases and behavioral coverage of complete/incomplete evidence and deadline/cancellation on the newly eligible privileged path.

The local tests complement the required GitHub checks and maintainer hardware validation. I am releasing the held workflows for this reviewed revision so we can get their results as well.

archledger:
CI update for reviewed head `0c77ce88` after approving the held workflows:

- All six CI jobs passed, including the MSRV checks, stable tests, fuzzing, cargo-deny, systemd checks and Nix.
- All three CodeQL analyses passed.
- Fedora 43 and 44 package builds passed.
- DCO failed with the expected message: the commit has **0 Signed-off-by trailers**, where exactly one is required.

The only failed automated check is DCO. The requested rationale correction, timeout-scope changes and regression coverage in the review still need addressing, followed by maintainer hardware validation before merge. Thanks again for the contribution and the detailed reproduction.

FizykPSX:
Thanks — all three items are addressed, and the timeout-scope one turned out to be a real defect rather than only a tidiness point. Pushed as `8c10394` and `94732f9`.

**1. DCO.** Signed off. The branch now carries exactly one `Signed-off-by` per commit, and the author identity matches my earlier merged commits here.

**2. Rationale.** Corrected, and thank you for the pointer to `pending_pad_retry_loop_completes_five_assessments_when_they_fit` — my claim was wrong as written. The ordinary loop does accumulate votes across healthy retries; what the reporting host establishes is only that its measured attempt cost (4.4–4.9 s) does not fit five of them into the privileged default, and that the 15-second experiment bought three votes rather than five. Five at that cost needs ~22 s, which is not a budget worth shipping — the efficiency of one-transaction collection is the motivation, not an impossibility. The narrower claim is now in the config doc comment, SETUP.md, the CHANGELOG and the PR description.

**3. Timeout scope.** You were right, and the first fix I tried was wrong in an instructive way. I moved the decision into `authenticate_in_window_inner`, where the route is known, and extended the window from the request's entry instant. On hardware every privileged attempt then died silently at the original 5000 ms: five grouped samples collected, IR armed, and no outcome line at all, with PAM falling through to the password. The window is the deadline contract *shared with admission* — the daemon hands one `AuthenticationWindow` to the engine and admits the response against that same window, so widening it mid-request widens capture and not admission.

So the budget is now decided where the window is created. `Engine::authentication_window_for` asks `grouped_route_possible`, which answers from facts available without opening a camera:

- the owner's opt-in,
- the sensor policy — experimental IR-only returns on its own route before grouped collection is consulted, so it must not buy the collector's budget,
- the purpose — credential release keeps its own scope,
- the service class,
- PAD and IR model availability,
- and the pair's **stored** qualification verdict, the same stored verdict `grouped_auth` requires.

The daemon calls that instead of `AuthenticationWindow::for_service`, so capture and admission share one window measured once from entry. `grace_window_ms` is back to its service table untouched, only the default short window is ever replaced, and an explicit `IRLUME_GRACE_MS` still decides alone — including a value below the login window and the legacy one-shot zero.

**What this shape concedes,** stated in the doc comment too: the runtime facts that can still refuse the collector — the pair's live contract, a runtime demotion — are unknowable at entry, so a stored-sequential pair that is demoted *during* the request keeps the longer budget while taking the ordinary path. That is the remaining over-grant. I could not see a way to close it without either breaking the one-deadline contract or paying a camera probe before the window exists; if you would rather have the short window everywhere and let the collector request its own extension through an explicit admission-aware mechanism, that is a bigger change I am happy to attempt instead.

**Coverage added:**

- `grouped_route_possible_from` as a pure decision in the style of `ir_selection_available`, tested across the matrix without a qualification store, models directory or config file: each privileged service with its own purpose; each requirement load-bearing on its own (opt-in off, non-sequential stored verdict, models absent, IR-only policy, credential release); and `login`, `sddm`, `omarchy-lock-face`, `sshd`, an invented name and `None` all staying out.
- `privileged_budget_for_route`: default-short-window-only, already-long untouched, and `IRLUME_GRACE_MS` winning in every direction.
- The newly eligible privileged services driven through the collector itself: complete evidence granting, incomplete evidence never reaching identity, and a deadline or a cancellation terminal with the evidence discarded.
- 228 authentication tests and 146 common-library tests pass; `fmt --check` clean.

**Hardware, on the reworked revision:** `sudo` `granted=true live=true score=0.865`, `verify total 7284 ms`, five `pad-vit` votes at 0.298–0.307; polkit `0.834` in 7271 ms; greeter and lock screen unchanged. I re-verified after the regression above rather than trusting the earlier run, since that run was on the shape you rejected.

On the ADR: still happy to write one if you want the scope of a PAD collection recorded that way, and equally happy to drop this PR if you would rather fix it for everyone by collecting the vote window sequentially inside one attempt.

crates/irlume-auth/src/lib.rs
```diff
@@ -5325,6 +5400,58 @@ impl Engine {
...
+        service: Option<&str>,
+        purpose: AuthenticationPurpose,
+        policy: irlume_common::config::FaceSensorPolicy,
+    ) -> bool {
+        grouped_route_possible_from(
+            service,
+            purpose,
+            policy,
+            self.ir_available && self.has_vit_pad() && self.has_pad_ir(),
+            stored_pair_requires_sequential(),
```
archledger, line 5450: [P1] This is live camera work, evaluated even when the opt-in/policy/model checks reject the route. `stored_pair_requires_sequential` calls `stored_capture_qualification`; camera/lib.rs:7812 forwards to `stored_capture_qualification_state`, which calls `collect_qualification_context` (8074-8079). That acquires a Diagnostics lease and performs `open_rgb` and `open_ir` (7736-7781). Thus configured IR-only requests touch RGB, and even opt-out/non-privileged/explicit-override requests can open/configure the pair before `AuthenticationWindow::new` establishes their deadline. I reproduced entry into this live-lookup boundary through the actual engine window API with a cfg(test)-only counter: opt-in=0 and IR-only each invoke it once, where the no-probe invariant expects zero. Please short-circuit exclusions and use a genuinely camera-free budget hint; preserve the authoritative live validation for the timed capture path. Merely calling the path selector `configured_pair_no_probe` does not make the subsequent qualification lookup non-probing.

crates/irlume-auth/src/grouped_auth.rs
```diff
@@ -72,6 +95,38 @@ pub(super) fn eligible_configuration(
...
+/// the capture route, its qualification and runtime standing, the models, and
+/// the purpose/service scope — is already decided by then, so this answers
+/// "could this configuration use the collector, given enough window?" and
+/// nothing else. Exact runtime authority stays in the outer gate, as in
+/// [`eligible`], so the caller ANDs this with the runtime contract.
+#[expect(
+    clippy::too_many_arguments,
+    reason = "mirrors eligible_configuration, whose arguments it shares"
+)]
+pub(super) fn eligible_configuration_ignoring_budget(
```
archledger, line 109: [P2] This helper appears to be left over from the earlier in-request budget approach and is now referenced only by tests. Both local and hosted CI Clippy fail with `function eligible_configuration_ignoring_budget is never used` in the non-test build. The nearby `#[expect(clippy::too_many_arguments)]` also fails as unfulfilled: this helper has seven arguments. Please remove the obsolete helper or restrict it appropriately to tests, remove the unnecessary expectation, and have route tests exercise the decision used by production.

archledger (changes requested):
Thanks for the revision. The sign-offs are now present, the rationale is corrected, and restoring the service-only short default addresses the original timeout regression. Keeping one admission/capture deadline is the right concern.

I carried an independently authored audit overlay onto 94732f9. The scoped suite passed 677 tests, including 20 reviewer-authored tests; the old short-window regression now passes. However, two new no-probe regression checks fail: with the opt-in disabled, and with IR-only policy selected, `Engine::authentication_window_for` still invokes the live pair-qualification lookup once (expected zero). The tests use configured synthetic paths and a test-only interception point immediately before `stored_capture_qualification`, so they require no camera.

The important API distinction is that `configured_pair_no_probe` does not open devices, but `stored_capture_qualification` does. It collects current context through a Diagnostics lease and `open_rgb`/`open_ir`, including negotiation. This happens before `AuthenticationWindow::new`, so the work is outside the new request budget. Because the argument is eagerly evaluated, it also happens for configurations that the cheap policy/model/opt-in checks subsequently reject. That is a regression with the opt-in off and for IR-only requests. Details and the call chain are inline.

There is also a concrete CI failure: the leftover `eligible_configuration_ignoring_budget` helper is unused in a non-test build, and its `too_many_arguments` expectation is unfulfilled. Local Clippy with warnings denied and the hosted CI Clippy step report the same errors. The revised DCO workflow has passed.

For clarity on hardware evidence: the separate minihost/NexiGo and Archhost/BRIO audit completed on the earlier 0c77ce88 production code. Real PAM negative checks and cancellation/recovery passed, and BRIO reached grouped collection for sudo and polkit. Those were unattended tests with zero detected faces, not genuine/PAD-inference qualification, and they do not qualify this revised binary. I have not activated 94732f9 on physical hardware.

Please make budget routing genuinely non-probing, short-circuit the cheap exclusions, keep any routing work inside a clearly accounted request-time contract, and remove or appropriately test-gate the obsolete helper. The documentation should also retain the late-runtime-demotion caveat you described rather than claiming every demoted request keeps the short window.

archledger:
Here is a tested reference fix against `94732f9`. Feel free to adapt it to your branch.

### What the code does

1. **Checks the cheap exclusions first.** Opt-out, IR-only, credential release, non-privileged requests, missing models, forced capture schedules and explicit time overrides never evaluate the hint.
2. **Uses metadata for a time-budget hint only.** The new reader uses `stat`, sysfs and one bounded, validated v2 record. It never opens a video node, issues an ioctl or acquires a camera lease. Missing/mismatched/unreadable records return false. The hint does not replace live capture qualification or authorize a grant.
3. **Keeps one deadline origin.** The start instant is captured before budget-selection reads, the explicit override is read once, and the resulting window is shared with response admission. The convenience engine API follows the same path.
4. **Removes the obsolete helper and lint expectation.** Its tests call the production predicate instead.

The central lazy guard is:

```rust
fn privileged_budget_for_route(
    window_ms: u64,
    explicit_override: bool,
    candidate: bool,
    hint: impl FnOnce() -> bool,
) -> Option<u64> {
    (!explicit_override && window_ms == SUDO_GRACE_WINDOW_MS && candidate && hint())
        .then_some(GRACE_WINDOW_MS)
}
```

The full patch includes the metadata reader, start-anchored window, caller wiring and regression tests. It intentionally retains the fixed-deadline caveat: a metadata hint may reserve time even if a later live stream mismatch or runtime degradation prevents grouped capture. It is a hint, not runtime authority.

**Verification:** 1,357 scoped tests passed (232 auth, 696 camera, 146 common, 283 daemon), including eight new focused tests; formatting and all-target Clippy with warnings denied passed. API docs passed with broken intra-doc links treated as errors. Read-only syscall checks on minihost/NexiGo and Archhost/BRIO recorded zero camera-node opens and zero ioctls: NexiGo returned no stored hint, BRIO returned its stored sequential hint. Services and live state were unchanged. No authentication or image capture was performed for this reference fix; final integrated hardware qualification is still needed.

Patch SHA256: `79a6d3324edca6194bf16ce91bbd4019165af54c4b426ddb2626c0a21c48c706`.

<details>
<summary>Complete tested reference patch</summary>

```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index 235577e74915c280da129c9e64ad0770fafde5b6..05cfaddfa4bb7fd7fcebde577a7acc81f5946e55 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,9 @@ All notable changes to irlume are documented here. This project adheres to
 - `privileged_grouped_pad_evidence=1` in `/etc/irlume/settings.conf` (or
   `IRLUME_PRIVILEGED_GROUPED_PAD=1`) lets `sudo`, `su`, `doas` and polkit use the
   bounded sequential PAD collection the greeter and lock screen already use, and
-  gives those services the login grace window that collection requires, and only
-  when the request can actually use that collection. Default off, so privileged
+  reserves that collection's login grace window only after cheap eligibility
+  checks and a matching stored sequential budget hint, without opening cameras.
+  Live capture qualification remains mandatory. Default off, so privileged
   surfaces are unchanged without it. On a camera pair that cannot capture RGB and
   IR concurrently one attempt casts a single ViT vote, so filling the five-vote
   RGB PAD ring takes five observed-cost attempts: affordable for the ordinary
diff --git a/crates/irlume-auth/src/authentication_window.rs b/crates/irlume-auth/src/authentication_window.rs
index 1e33cfa60c030f30e32eebfdb262c3d65e3e74b7..6c0a746b5060f74408cbb65b99548de5da046429 100644
--- a/crates/irlume-auth/src/authentication_window.rs
+++ b/crates/irlume-auth/src/authentication_window.rs
@@ -15,8 +15,12 @@ impl AuthenticationWindow {
     }
 
     pub(crate) fn new(milliseconds: u64) -> Self {
+        Self::from_started(Instant::now(), milliseconds)
+    }
+
+    pub(crate) fn from_started(started: Instant, milliseconds: u64) -> Self {
         Self {
-            deadline: Instant::now() + Duration::from_millis(milliseconds),
+            deadline: started + Duration::from_millis(milliseconds),
             milliseconds,
         }
     }
diff --git a/crates/irlume-auth/src/engine_tests/budget_suggestion_tests.rs b/crates/irlume-auth/src/engine_tests/budget_suggestion_tests.rs
new file mode 100644
index 0000000000000000000000000000000000000000..ebd3dc51da07cd948509b7327537909d1b80fd84
--- /dev/null
+++ b/crates/irlume-auth/src/engine_tests/budget_suggestion_tests.rs
@@ -0,0 +1,269 @@
+use super::*;
+use irlume_common::config::FaceSensorPolicy::{Dual, IrOnlyExperimental};
+use std::cell::Cell;
+use std::time::{Duration, Instant};
+
+struct Environment(Vec<(&'static str, Option<std::ffi::OsString>)>);
+impl Environment {
+    fn new() -> Self {
+        let keys = [
+            "IRLUME_PRIVILEGED_GROUPED_PAD",
+            "IRLUME_GRACE_MS",
+            "IRLUME_SEQUENTIAL_CAPTURE",
+            "IRLUME_CONFIG_DIR",
+        ];
+        let result = Self(
+            keys.iter()
+                .map(|key| (*key, std::env::var_os(key)))
+                .collect(),
+        );
+        for key in keys {
+            std::env::remove_var(key);
+        }
+        result
+    }
+}
+impl Drop for Environment {
+    fn drop(&mut self) {
+        for (key, value) in &self.0 {
+            match value {
+                Some(value) => std::env::set_var(key, value),
+                None => std::env::remove_var(key),
+            }
+        }
+    }
+}
+
+struct ReadyEngine<'a> {
+    engine: &'a mut Engine,
+    old_ir: bool,
+    old_vit: Option<irlume_vision::PadVit>,
+}
+impl<'a> ReadyEngine<'a> {
+    fn new(engine: &'a mut Engine) -> Self {
+        let model =
+            irlume_vision::PadVit::load_from_file(&model_path("liveness_vit.onnx")).unwrap();
+        let old_vit = engine.vit_pad.replace(model);
+        let old_ir = std::mem::replace(&mut engine.ir_available, true);
+        assert!(engine.has_pad_ir());
+        Self {
+            engine,
+            old_ir,
+            old_vit,
+        }
+    }
+}
+impl Drop for ReadyEngine<'_> {
+    fn drop(&mut self) {
+        self.engine.ir_available = self.old_ir;
+        self.engine.vit_pad = self.old_vit.take();
+    }
+}
+
+#[test]
+fn budget_hint_is_not_evaluated_for_excluded_requests() {
+    let _guard = env_guard();
+    let _env = Environment::new();
+    let mut s = shared();
+    let ready = ReadyEngine::new(&mut s.engine);
+    for (enabled, policy, purpose, service, expected) in [
+        ("0", Dual, AuthenticationPurpose::Verify, "sudo", 5_000),
+        (
+            "1",
+            IrOnlyExperimental,
+            AuthenticationPurpose::Verify,
+            "sudo",
+            5_000,
+        ),
+        (
+            "1",
+            Dual,
+            AuthenticationPurpose::CredentialRelease,
+            "sudo",
+            5_000,
+        ),
+        ("1", Dual, AuthenticationPurpose::Verify, "kde", 15_000),
+        ("1", Dual, AuthenticationPurpose::Verify, "sshd", 15_000),
+    ] {
+        std::env::set_var("IRLUME_PRIVILEGED_GROUPED_PAD", enabled);
+        let window = ready.engine.authentication_window_from_with_hint(
+            Instant::now(),
+            Some(service),
+            purpose,
+            policy,
+            || panic!("excluded request evaluated the hint"),
+        );
+        assert_eq!(window.milliseconds, expected);
+    }
+    std::env::set_var("IRLUME_PRIVILEGED_GROUPED_PAD", "1");
+    for value in ["0", "5000", "8000", "60000"] {
+        std::env::set_var("IRLUME_GRACE_MS", value);
+        let window = ready.engine.authentication_window_from_with_hint(
+            Instant::now(),
+            Some("sudo"),
+            AuthenticationPurpose::Verify,
+            Dual,
+            || panic!("explicit override evaluated the hint"),
+        );
+        assert_eq!(window.milliseconds, value.parse::<u64>().unwrap());
+    }
+    std::env::remove_var("IRLUME_GRACE_MS");
+    for value in ["0", "1", ""] {
+        std::env::set_var("IRLUME_SEQUENTIAL_CAPTURE", value);
+        let window = ready.engine.authentication_window_from_with_hint(
+            Instant::now(),
+            Some("sudo"),
+            AuthenticationPurpose::Verify,
+            Dual,
+            || panic!("forced capture schedule evaluated the hint"),
+        );
+        assert_eq!(window.milliseconds, 5_000);
+    }
+    std::env::remove_var("IRLUME_SEQUENTIAL_CAPTURE");
+    ready.engine.ir_available = false;
+    let window = ready.engine.authentication_window_from_with_hint(
+        Instant::now(),
+        Some("sudo"),
+        AuthenticationPurpose::Verify,
+        Dual,
+        || panic!("missing IR evaluated the hint"),
+    );
+    assert_eq!(window.milliseconds, 5_000);
+    ready.engine.ir_available = true;
+    let old_vit = ready.engine.vit_pad.take();
+    let calls = Cell::new(0);
+    let window = ready.engine.authentication_window_from_with_hint(
+        Instant::now(),
+        Some("sudo"),
+        AuthenticationPurpose::Verify,
+        Dual,
+        || {
+            calls.set(calls.get() + 1);
+            true
+        },
+    );
+    ready.engine.vit_pad = old_vit;
+    assert_eq!(calls.get(), 0);
+    assert_eq!(window.milliseconds, 5_000);
+    let old_ir_pad = ready.engine.pad_ir.take();
+    let window = ready.engine.authentication_window_from_with_hint(
+        Instant::now(),
+        Some("sudo"),
+        AuthenticationPurpose::Verify,
+        Dual,
+        || {
+            calls.set(calls.get() + 1);
+            true
+        },
+    );
+    ready.engine.pad_ir = old_ir_pad;
+    assert_eq!(calls.get(), 0);
+    assert_eq!(window.milliseconds, 5_000);
+}
+
+#[test]
+fn budget_hint_reserves_time_once_and_never_resets_request_origin() {
+    let _guard = env_guard();
+    let _env = Environment::new();
+    let mut s = shared();
+    let ready = ReadyEngine::new(&mut s.engine);
+    std::env::set_var("IRLUME_PRIVILEGED_GROUPED_PAD", "1");
+    for service in ["sudo", "polkit-1"] {
+        for hint in [false, true] {
+            let started = Instant::now() - Duration::from_secs(2);
+            let calls = Cell::new(0);
+            let window = ready.engine.authentication_window_from_with_hint(
+                started,
+                Some(service),
+                AuthenticationPurpose::for_service(Some(service)),
+                Dual,
+                || {
+                    calls.set(calls.get() + 1);
+                    hint
+                },
+            );
+            let expected = if hint { 15_000 } else { 5_000 };
+            assert_eq!(calls.get(), 1);
+            assert_eq!(window.milliseconds, expected);
+            assert_eq!(window.deadline, started + Duration::from_millis(expected));
+        }
+    }
+    let window = ready.engine.authentication_window_from_with_hint(
+        Instant::now(),
+        Some("sudo"),
+        AuthenticationPurpose::Verify,
+        Dual,
+        || {
+            std::env::set_var("IRLUME_GRACE_MS", "0");
+            true
+        },
+    );
+    assert_eq!(
+        window.milliseconds, 15_000,
+        "the entry override snapshot must not be reread after the hint"
+    );
+}
+
+#[test]
+fn budget_convenience_entry_keeps_refusal_state_and_cancellation_precedence() {
+    let _guard = env_guard();
+    let _env = Environment::new();
+    let mut s = shared();
+    let dir = std::env::temp_dir().join(format!("irlume-budget-policy-{}", std::process::id()));
+    std::fs::create_dir_all(&dir).unwrap();
+    std::fs::write(dir.join("settings.conf"), "face_sensor_policy=invalid\n").unwrap();
+    std::env::set_var("IRLUME_CONFIG_DIR", &dir);
+    s.engine.last_attempt_situation = Some(AttemptSituation::NoFace);
+    let result = s.engine.authenticate_for_with_diagnostics(
+        "budget-fixture",
+        Some("sudo"),
+        AuthenticationPurpose::Verify,
+        &(),
+    );
+    assert!(result.is_err());
+    assert!(s.engine.last_attempt_situation.is_none());
+    let prior = s
+        .engine
+        .request_cancelled
+        .replace(std::sync::Arc::new(|| true));
+    let result = s.engine.authenticate_for_with_diagnostics(
+        "budget-fixture",
+        Some("sudo"),
+        AuthenticationPurpose::Verify,
+        &(),
+    );
+    s.engine.request_cancelled = prior;
+    std::fs::remove_dir_all(dir).unwrap();
+    assert!(matches!(result, Err(irlume_common::Error::Preempted(_))));
+}
+
+#[test]
+fn budget_selection_cost_can_expire_window_but_zero_stays_one_shot() {
+    let _guard = env_guard();
+    let _env = Environment::new();
+    let mut s = shared();
+    let ready = ReadyEngine::new(&mut s.engine);
+    std::env::set_var("IRLUME_PRIVILEGED_GROUPED_PAD", "1");
+    let started = Instant::now() - Duration::from_secs(20);
+    let window = ready.engine.authentication_window_from_with_hint(
+        started,
+        Some("sudo"),
+        AuthenticationPurpose::Verify,
+        Dual,
+        || true,
+    );
+    assert!(
+        window.check().is_err(),
+        "metadata lookup must not grant a fresh 15 seconds afterward"
+    );
+    std::env::set_var("IRLUME_GRACE_MS", "0");
+    let window = ready.engine.authentication_window_from_with_hint(
+        started,
+        Some("sudo"),
+        AuthenticationPurpose::Verify,
+        Dual,
+        || panic!("one-shot evaluated hint"),
+    );
+    assert!(window.remaining().is_none());
+    assert!(window.check().is_ok());
+}
diff --git a/crates/irlume-auth/src/engine_tests/grouped_tests.rs b/crates/irlume-auth/src/engine_tests/grouped_tests.rs
index 28f284557f2c4873c477620beea79c6fccadc753..ed0d13c4a613fafd9f2e2e7329bc7563e7b2fa83 100644
--- a/crates/irlume-auth/src/engine_tests/grouped_tests.rs
+++ b/crates/irlume-auth/src/engine_tests/grouped_tests.rs
@@ -956,11 +956,12 @@ fn privileged_grouped_collection_grants_refuses_and_expires_like_a_greeter() {
 fn privileged_budget_follows_the_capture_route_not_the_service_name() {
     use irlume_common::diagnostics::QualificationState;
     let ready = |mode: &CaptureModeSelection, ir, rgb_pad, ir_pad, service| {
-        crate::grouped_auth::eligible_configuration_ignoring_budget(
+        crate::grouped_auth::eligible_configuration(
             mode,
             ir,
             rgb_pad,
             ir_pad,
+            crate::GRACE_WINDOW_MS,
             AuthenticationPurpose::Verify,
             service,
             true,
@@ -969,7 +970,7 @@ fn privileged_budget_follows_the_capture_route_not_the_service_name() {
     let mode = measured_sequential_configuration();
     assert!(ready(&mode, true, true, true, Some("sudo")));
     assert_eq!(
-        crate::privileged_budget_for_route(crate::SUDO_GRACE_WINDOW_MS, true),
+        crate::privileged_budget_for_route(crate::SUDO_GRACE_WINDOW_MS, false, true, || true),
         Some(crate::GRACE_WINDOW_MS)
     );
 
@@ -993,7 +994,12 @@ fn privileged_budget_follows_the_capture_route_not_the_service_name() {
         let route_ready = ready(&m, true, true, ir_pad, Some("sudo"));
         assert!(!route_ready, "{label} must not reach grouped collection");
         assert_eq!(
-            crate::privileged_budget_for_route(crate::SUDO_GRACE_WINDOW_MS, route_ready),
+            crate::privileged_budget_for_route(
+                crate::SUDO_GRACE_WINDOW_MS,
+                false,
+                route_ready,
+                || true
+            ),
             None,
             "{label} must keep the short privileged window"
         );
@@ -1007,7 +1013,7 @@ fn privileged_budget_follows_the_capture_route_not_the_service_name() {
     // says. Asserted as the sensor policy's own early return in
     // `authenticate_in_window_inner`; here we pin the budget half.
     assert_eq!(
-        crate::privileged_budget_for_route(crate::SUDO_GRACE_WINDOW_MS, false),
+        crate::privileged_budget_for_route(crate::SUDO_GRACE_WINDOW_MS, false, false, || true),
         None
     );
 }
diff --git a/crates/irlume-auth/src/grouped_auth.rs b/crates/irlume-auth/src/grouped_auth.rs
index 0136a22b9a2744da2b394e34e3662c58ff035437..5e10ff6f02f31d104e144edfa4ab9a5addb7e85c 100644
--- a/crates/irlume-auth/src/grouped_auth.rs
+++ b/crates/irlume-auth/src/grouped_auth.rs
@@ -95,38 +95,6 @@ pub(super) fn eligible_configuration(
         && window >= GRACE_WINDOW_MS
 }
 
-/// [`eligible_configuration`] without its budget term, for deciding what budget
-/// the request needs in the first place. Every other configuration requirement —
-/// the capture route, its qualification and runtime standing, the models, and
-/// the purpose/service scope — is already decided by then, so this answers
-/// "could this configuration use the collector, given enough window?" and
-/// nothing else. Exact runtime authority stays in the outer gate, as in
-/// [`eligible`], so the caller ANDs this with the runtime contract.
-#[expect(
-    clippy::too_many_arguments,
-    reason = "mirrors eligible_configuration, whose arguments it shares"
-)]
-pub(super) fn eligible_configuration_ignoring_budget(
-    mode: &CaptureModeSelection,
-    has_ir: bool,
-    has_rgb_pad: bool,
-    has_ir_pad: bool,
-    purpose: AuthenticationPurpose,
-    service: Option<&str>,
-    privileged_opt_in: bool,
-) -> bool {
-    eligible_configuration(
-        mode,
-        has_ir,
-        has_rgb_pad,
-        has_ir_pad,
-        GRACE_WINDOW_MS,
-        purpose,
-        service,
-        privileged_opt_in,
-    )
-}
-
 fn expired() -> Outcome {
     Outcome::deny(
         OutcomeKind::DeadlineExpired,
diff --git a/crates/irlume-auth/src/lib.rs b/crates/irlume-auth/src/lib.rs
index 7af980e362cc87007402a637345980955b288fb4..84644c6d59a767c4640c1639856697f8d7c899c0 100644
--- a/crates/irlume-auth/src/lib.rs
+++ b/crates/irlume-auth/src/lib.rs
@@ -921,6 +921,10 @@ fn grace_window_ms(service: Option<&str>) -> u64 {
     if let Some(v) = grace_window_override_ms() {
         return v;
     }
+    default_grace_window_ms(service)
+}
+
+fn default_grace_window_ms(service: Option<&str>) -> u64 {
     // From the shared table, not a local list. The list this replaced was
     // missing `doas`, which is Elevation for the policy, so a doas prompt held
     // the camera for the 15s login window instead of the 5s one (#362).
@@ -941,23 +945,25 @@ fn grace_window_ms(service: Option<&str>) -> u64 {
 /// collector — and that collector is itself gated on
 /// `window >= GRACE_WINDOW_MS`, so nothing would change without this.
 ///
-/// `grouped_ready` is [`Engine::grouped_route_possible`], so a request that
-/// cannot use the collector keeps the short window: a concurrent or unqualified
-/// pair, a request whose PAD models are missing, credential release, and
-/// experimental IR-only.
+/// `candidate` contains the inexpensive policy/model checks. The metadata-only
+/// hint is lazy: excluded requests never read camera metadata or stored records.
+/// A true hint reserves time only, not capture or grant authority. Stale stream
+/// contracts or later runtime degradation can still prevent grouped capture.
 ///
 /// Only the DEFAULT short window is replaced. An explicit `IRLUME_GRACE_MS`
 /// still decides the budget on its own, including a smaller one and the legacy
 /// one-shot zero, because an operator who names a number has named it for every
 /// service.
 ///
-/// This is answered before the request starts, because the budget IS the
-/// request's deadline: the daemon admits the response against the same window
-/// it hands the engine, so a window widened mid-request would be widened for
-/// capture and not for admission, and the request would die at the original
-/// deadline with no outcome at all.
-fn privileged_budget_for_route(window_ms: u64, grouped_ready: bool) -> Option<u64> {
-    (grouped_ready && grace_window_override_ms().is_none() && window_ms == SUDO_GRACE_WINDOW_MS)
+/// The caller anchors the resulting window before these reads. Capture and
+/// response admission retain that same window; routing does not reset its origin.
+fn privileged_budget_for_route(
+    window_ms: u64,
+    explicit_override: bool,
+    candidate: bool,
+    hint: impl FnOnce() -> bool,
+) -> Option<u64> {
+    (!explicit_override && window_ms == SUDO_GRACE_WINDOW_MS && candidate && hint())
         .then_some(GRACE_WINDOW_MS)
 }
 
@@ -987,20 +993,6 @@ fn grouped_route_possible_from(
         && stored_sequential
 }
 
-/// Does the configured pair's STORED qualification say sequential capture?
-///
-/// Camera-free, and deliberately the same stored verdict `grouped_auth`
-/// requires: read from the qualification store by device path, so the budget
-/// decision can precede the request without opening anything.
-fn stored_pair_requires_sequential() -> bool {
-    irlume_camera::configured_pair_no_probe().is_some_and(|(rgb, ir)| {
-        matches!(
-            irlume_camera::stored_capture_qualification(&rgb, &ir),
-            Ok(QualificationResolution::SequentialRequired(_))
-        )
-    })
-}
-
 /// The operator's explicit window, when set and within bounds.
 fn grace_window_override_ms() -> Option<u64> {
     std::env::var("IRLUME_GRACE_MS")
@@ -5400,15 +5392,11 @@ impl Engine {
         self.authenticate_for_with_diagnostics(user, service, purpose, &())
     }
 
-    /// The window this request must run under, chosen before it starts.
-    ///
-    /// The budget IS the request's deadline: the daemon admits the response
-    /// against the same window it hands this engine, so it cannot be widened
-    /// once capture is under way — a request whose capture believed in a longer
-    /// window than its admission would simply die at the original deadline with
-    /// no outcome. Callers that build a window for an authentication should use
-    /// this rather than [`AuthenticationWindow::for_service`], which knows only
-    /// the service name.
+    /// One request window anchored before non-probing budget selection.
+    /// Capture and response admission retain the same deadline, including time
+    /// spent reading configuration and the metadata-only hint. A hint reserves
+    /// time only; live stream qualification can still refuse grouped capture.
+    /// Use this rather than the service-only [`AuthenticationWindow::for_service`].
     #[must_use]
     pub fn authentication_window_for(
         &self,
@@ -5416,39 +5404,54 @@ impl Engine {
         purpose: AuthenticationPurpose,
         policy: irlume_common::config::FaceSensorPolicy,
     ) -> AuthenticationWindow {
-        let base = grace_window_ms(service);
-        AuthenticationWindow::new(
-            privileged_budget_for_route(
-                base,
-                self.grouped_route_possible(service, purpose, policy),
+        self.authentication_window_from(std::time::Instant::now(), service, purpose, policy)
+    }
+
+    fn authentication_window_from(
+        &self,
+        started: std::time::Instant,
+        service: Option<&str>,
+        purpose: AuthenticationPurpose,
+        policy: irlume_common::config::FaceSensorPolicy,
+    ) -> AuthenticationWindow {
+        self.authentication_window_from_with_hint(started, service, purpose, policy, || {
+            irlume_camera::capture_qualification::sequential_budget_hint(
+                &self.rgb_dev,
+                &self.ir_dev,
             )
-            .unwrap_or(base),
-        )
+        })
     }
 
-    /// Could this request reach grouped collection, on the facts available
-    /// before it starts?
-    ///
-    /// Every fact here is re-checked by [`grouped_auth::eligible`] once the pair
-    /// is open, so a yes only ever buys the budget that collector needs and
-    /// never admits evidence. The runtime facts that can still refuse it — the
-    /// pair's live contract and a runtime demotion — are unknowable at entry and
-    /// can only narrow eligibility, so such a request spends the longer budget
-    /// on the ordinary path. That is the one over-grant this shape keeps, and it
-    /// is the price of one deadline per request.
-    fn grouped_route_possible(
+    fn authentication_window_from_with_hint(
         &self,
+        started: std::time::Instant,
         service: Option<&str>,
         purpose: AuthenticationPurpose,
         policy: irlume_common::config::FaceSensorPolicy,
-    ) -> bool {
-        grouped_route_possible_from(
-            service,
-            purpose,
-            policy,
-            self.ir_available && self.has_vit_pad() && self.has_pad_ir(),
-            stored_pair_requires_sequential(),
-            irlume_common::config::privileged_grouped_pad_evidence_enabled(),
+        hint: impl FnOnce() -> bool,
+    ) -> AuthenticationWindow {
+        // Snapshot the explicit override once and count all routing reads in
+        // the one window later retained by both capture and response admission.
+        let override_ms = grace_window_override_ms();
+        let base = override_ms.unwrap_or_else(|| default_grace_window_ms(service));
+        // Passing `true` for stored availability screens the cheap facts only.
+        // The actual hint is deferred until every exclusion has passed.
+        let candidate = override_ms.is_none()
+            && base == SUDO_GRACE_WINDOW_MS
+            && std::env::var("IRLUME_SEQUENTIAL_CAPTURE").is_err()
+            && grouped_route_possible_from(
+                service,
+                purpose,
+                policy,
+                self.ir_available && self.has_vit_pad() && self.has_pad_ir(),
+                true,
+                true,
+            )
+            && irlume_common::config::privileged_grouped_pad_evidence_enabled();
+        AuthenticationWindow::from_started(
+            started,
+            privileged_budget_for_route(base, override_ms.is_some(), candidate, hint)
+                .unwrap_or(base),
         )
     }
 
@@ -5462,11 +5465,23 @@ impl Engine {
         purpose: AuthenticationPurpose,
         diagnostics: &dyn irlume_common::diagnostics::DiagnosticSink,
     ) -> irlume_common::Result<Outcome> {
-        self.authenticate_for_in_window(
+        let started = std::time::Instant::now();
+        self.last_attempt_situation = None;
+        self.check_request_cancelled()?;
+        let policy = irlume_common::config::observe_face_sensor_policy().resolve()?;
+        let window = self.authentication_window_from(started, service, purpose, policy);
+        if let Err(error) = self.check_authentication_completion(window) {
+            if matches!(error, irlume_common::Error::DeadlineExpired) {
+                self.last_attempt_situation = Some(AttemptSituation::TimedOut);
+            }
+            return Err(error);
+        }
+        self.authenticate_for_in_window_with_policy(
             user,
             service,
             purpose,
-            AuthenticationWindow::for_service(service),
+            window,
+            policy,
             diagnostics,
         )
     }
@@ -9011,24 +9026,31 @@ mod tests {
         std::env::remove_var("IRLUME_GRACE_MS");
         // Ready for the collector: the default short window is replaced.
         assert_eq!(
-            privileged_budget_for_route(SUDO_GRACE_WINDOW_MS, true),
+            privileged_budget_for_route(SUDO_GRACE_WINDOW_MS, false, true, || true),
             Some(GRACE_WINDOW_MS)
         );
         // Not ready (concurrent, unqualified, demoted, models absent,
         // credential release, IR-only): nothing moves.
         assert_eq!(
-            privileged_budget_for_route(SUDO_GRACE_WINDOW_MS, false),
+            privileged_budget_for_route(SUDO_GRACE_WINDOW_MS, false, false, || panic!(
+                "excluded hint"
+            )),
             None
         );
         // A login/lock request is already on the long window.
-        assert_eq!(privileged_budget_for_route(GRACE_WINDOW_MS, true), None);
+        assert_eq!(
+            privileged_budget_for_route(GRACE_WINDOW_MS, false, true, || panic!(
+                "long window hint"
+            )),
+            None
+        );
         // An explicit override decides on its own, in both directions, and the
         // legacy one-shot zero is only reachable that way.
         for value in ["8000", "0", "30000"] {
             std::env::set_var("IRLUME_GRACE_MS", value);
             let named = grace_window_ms(Some("sudo"));
             assert_eq!(
-                privileged_budget_for_route(named, true),
+                privileged_budget_for_route(named, true, true, || panic!("override hint")),
                 None,
                 "IRLUME_GRACE_MS={value} must keep deciding the budget"
             );
@@ -11296,6 +11318,7 @@ mod pad_cue_tests {
 /// build (the 512-D recognizer session), so one instance is shared.
 #[cfg(test)]
 mod engine_tests {
+    mod budget_suggestion_tests;
     mod grouped_tests;
     mod managed_pad_tests;
     mod pair_identity_tests;
diff --git a/crates/irlume-camera/src/capture_qualification.rs b/crates/irlume-camera/src/capture_qualification.rs
index d6f0ceeb7343a94828032721cd3f6972b177f1d0..827381181e8a296bb4bc03b0867152acbf2e2793 100644
--- a/crates/irlume-camera/src/capture_qualification.rs
+++ b/crates/irlume-camera/src/capture_qualification.rs
@@ -221,6 +221,33 @@ pub struct CameraEndpoint {
 }
 
 impl CameraEndpoint {
+    // Private: a path observation may supply filename/identity hints, never an
+    // fd-derived runtime contract. Only a boolean escapes sequential_budget_hint.
+    fn for_budget_hint(path: &str, role: QualifiedStreamRole) -> Option<Self> {
+        let (identity, connection) =
+            crate::uvc_descriptor::identity_and_connection_for_budget_hint(path).ok()?;
+        if connection.driver != "uvcvideo" {
+            return None;
+        }
+        Self::new(
+            irlume_common::sha256_hex(&identity.descriptors),
+            identity.vid,
+            identity.pid,
+            identity.serial,
+            identity.interface_number,
+            identity.usb_devpath,
+            role,
+            ConnectionContext::new(
+                connection.controller_devpath,
+                connection.speed_millimbps,
+                connection.driver,
+                "uvc-v4l2".into(),
+            )
+            .ok()?,
+        )
+        .ok()
+    }
+
     /// Construct one fd-derived endpoint identity.
     ///
     /// # Errors
@@ -1382,7 +1409,64 @@ pub struct QualificationStore {
     dir: PathBuf,
 }
 
+/// Non-authoritative hint for reserving a sequential-collection time budget.
+/// Reads path metadata, sysfs and one bounded validated v2 record; never opens
+/// a camera, negotiates a format, or acquires a camera lease. Missing, changed,
+/// malformed or unreadable evidence returns false. Stream contracts are NOT
+/// observed here: live qualification remains mandatory before grouped capture.
+#[must_use]
+pub fn sequential_budget_hint(rgb_path: &str, ir_path: &str) -> bool {
+    let Some(rgb) = CameraEndpoint::for_budget_hint(rgb_path, QualifiedStreamRole::Rgb) else {
+        return false;
+    };
+    let Some(ir) = CameraEndpoint::for_budget_hint(ir_path, QualifiedStreamRole::Ir) else {
+        return false;
+    };
+    QualificationStore::system().sequential_budget_hint_for_endpoints(&rgb, &ir)
+}
+
 impl QualificationStore {
+    fn sequential_budget_hint_for_endpoints(
+        &self,
+        rgb: &CameraEndpoint,
+        ir: &CameraEndpoint,
+    ) -> bool {
+        use std::os::unix::fs::OpenOptionsExt;
+        let read = || -> Option<bool> {
+            if rgb.role != QualifiedStreamRole::Rgb || ir.role != QualifiedStreamRole::Ir {
+                return None;
+            }
+            rgb.validate().ok()?;
+            ir.validate().ok()?;
+            let path = self
+                .dir
+                .join(format!("{}.json", pair_endpoint_filing_key(rgb, ir)));
+            let file = std::fs::OpenOptions::new()
+                .read(true)
+                .custom_flags(libc::O_CLOEXEC | libc::O_NOFOLLOW | libc::O_NONBLOCK)
+                .open(path)
+                .ok()?;
+            if !file.metadata().ok()?.is_file() {
+                return None;
+            }
+            let mut body = Vec::new();
+            file.take((MAX_RECORD_BYTES + 1) as u64)
+                .read_to_end(&mut body)
+                .ok()?;
+            let record = CaptureQualificationRecord::from_json(&body).ok()?;
+            let authoritative = record.authoritative()?;
+            Some(
+                authoritative.context().rgb_endpoint == *rgb
+                    && authoritative.context().ir_endpoint == *ir
+                    && matches!(
+                        authoritative.outcome(),
+                        AttemptOutcome::SequentialRequired(_)
+                    ),
+            )
+        };
+        read().unwrap_or(false)
+    }
+
     /// The production machine-state store.
     #[must_use]
     pub fn system() -> Self {
@@ -1479,8 +1563,12 @@ impl QualificationStore {
 }
 
 fn pair_filing_key(context: &QualificationContext) -> String {
-    let rgb = context.rgb_endpoint.filing_key();
-    let ir = context.ir_endpoint.filing_key();
+    pair_endpoint_filing_key(&context.rgb_endpoint, &context.ir_endpoint)
+}
+
+fn pair_endpoint_filing_key(rgb_endpoint: &CameraEndpoint, ir_endpoint: &CameraEndpoint) -> String {
+    let rgb = rgb_endpoint.filing_key();
+    let ir = ir_endpoint.filing_key();
     irlume_common::sha256_hex(format!("rgb:{}:{rgb}|ir:{}:{ir}", rgb.len(), ir.len()).as_bytes())
 }
 
@@ -1870,6 +1958,106 @@ mod tests {
         concurrent_attempt_with_presence(port, None)
     }
 
+    fn sequential_hint_attempt() -> QualificationAttempt {
+        let mut attempt = concurrent_attempt("/devices/pci0000:00/usb3/3-2");
+        attempt.concurrent.rgb_mean = 1.0;
+        attempt.outcome = AttemptOutcome::SequentialRequired(SequentialReason::SignalLoss);
+        attempt.validate().unwrap();
+        attempt
+    }
+
+    #[test]
+    fn budget_hint_uses_only_matching_conclusive_sequential_record() {
+        let dir = TempStore::new("budget-hint");
+        let store = dir.store();
+        let seq = sequential_hint_attempt();
+        let c = seq.context();
+        assert!(!store.sequential_budget_hint_for_endpoints(&c.rgb_endpoint, &c.ir_endpoint));
+        store
+            .save_attempt(concurrent_attempt("/devices/pci0000:00/usb3/3-2"), None)
+            .unwrap();
+        assert!(!store.sequential_budget_hint_for_endpoints(&c.rgb_endpoint, &c.ir_endpoint));
+        store.save_attempt(seq.clone(), Some(1)).unwrap();
+        assert!(store.sequential_budget_hint_for_endpoints(&c.rgb_endpoint, &c.ir_endpoint));
+        let mut later = seq.clone();
+        later.outcome = AttemptOutcome::Inconclusive(InconclusiveReason::IncompleteRounds);
+        store.save_attempt(later, Some(2)).unwrap();
+        assert!(
+            store.sequential_budget_hint_for_endpoints(&c.rgb_endpoint, &c.ir_endpoint),
+            "last attempt must not replace retained conclusive evidence"
+        );
+        for change in 0..7 {
+            let mut rgb = c.rgb_endpoint.clone();
+            match change {
+                0 => rgb.connection.speed_millimbps /= 2,
+                1 => rgb.connection.driver = "different".into(),
+                2 => rgb.connection.backend = "different".into(),
+                3 => rgb.descriptor_sha256 = "cd".repeat(32),
+                4 => rgb.serial = Some("different".into()),
+                5 => rgb.usb_devpath = "/devices/pci0000:00/usb3/3-9".into(),
+                _ => rgb.interface_number += 1,
+            }
+            assert!(
+                !store.sequential_budget_hint_for_endpoints(&rgb, &c.ir_endpoint),
+                "change={change}"
+            );
+        }
+        assert!(!store.sequential_budget_hint_for_endpoints(&c.ir_endpoint, &c.rgb_endpoint));
+    }
+
+    #[test]
+    fn budget_hint_fails_closed_for_bad_or_nonregular_records() {
+        use std::os::unix::ffi::OsStrExt;
+        let dir = TempStore::new("budget-hint-errors");
+        let store = dir.store();
+        let seq = sequential_hint_attempt();
+        let c = seq.context();
+        store.save_attempt(seq.clone(), None).unwrap();
+        let path = store.record_path(c);
+        for body in [
+            b"{".to_vec(),
+            vec![b' '; MAX_RECORD_BYTES + 1],
+            b"{}".to_vec(),
+        ] {
+            std::fs::write(&path, body).unwrap();
+            assert!(!store.sequential_budget_hint_for_endpoints(&c.rgb_endpoint, &c.ir_endpoint));
+        }
+        std::fs::remove_file(&path).unwrap();
+        std::os::unix::fs::symlink("missing", &path).unwrap();
+        assert!(!store.sequential_budget_hint_for_endpoints(&c.rgb_endpoint, &c.ir_endpoint));
+        std::fs::remove_file(&path).unwrap();
+        let fifo = std::ffi::CString::new(path.as_os_str().as_bytes()).unwrap();
+        // SAFETY: fifo is a live NUL-terminated path inside this test's private directory.
+        assert_eq!(unsafe { libc::mkfifo(fifo.as_ptr(), 0o600) }, 0);
+        assert!(!store.sequential_budget_hint_for_endpoints(&c.rgb_endpoint, &c.ir_endpoint));
+        std::fs::remove_file(&path).unwrap();
+        let mut record =
+            CaptureQualificationRecord::new(1, seq.clone(), Some(seq.clone())).unwrap();
+        record.policy_version += 1;
+        std::fs::write(&path, serde_json::to_vec(&record).unwrap()).unwrap();
+        assert!(!store.sequential_budget_hint_for_endpoints(&c.rgb_endpoint, &c.ir_endpoint));
+    }
+
+    #[test]
+    fn budget_hint_is_not_stream_qualification() {
+        let dir = TempStore::new("budget-hint-not-authority");
+        let store = dir.store();
+        let seq = sequential_hint_attempt();
+        let mut changed = seq.context().clone();
+        let record = store.save_attempt(seq, None).unwrap();
+        changed.ir_stream.requested.width += 1;
+        assert!(
+            store.sequential_budget_hint_for_endpoints(&changed.rgb_endpoint, &changed.ir_endpoint)
+        );
+        assert!(
+            matches!(
+                record.resolve(&changed),
+                QualificationResolution::Unqualified(_)
+            ),
+            "the hint must not replace the exact live stream gate"
+        );
+    }
+
     fn concurrent_attempt_with_presence(
         port: &str,
         ir_illumination_metadata: Option<IlluminationMetadataPresence>,
diff --git a/crates/irlume-camera/src/uvc_descriptor.rs b/crates/irlume-camera/src/uvc_descriptor.rs
index eedc453e3884feece798723a314e95b59bcd807c..1128f2fd041d3507982a9f9bc22d7d618d8c8295 100644
--- a/crates/irlume-camera/src/uvc_descriptor.rs
+++ b/crates/irlume-camera/src/uvc_descriptor.rs
@@ -285,6 +285,30 @@ pub(crate) fn identity_and_connection_from_fd(
 
 fn fd_usb_dirs(fd: std::os::raw::c_int) -> std::io::Result<(PathBuf, PathBuf)> {
     let (major, minor) = device_numbers(fd)?;
+    usb_dirs_for_numbers(major, minor)
+}
+
+/// Metadata-only observation for a non-authoritative time-budget hint.
+/// Unlike the fd collector, this never opens the video node or issues ioctls.
+/// Path/stat/sysfs races are acceptable only because capture revalidates its
+/// own fd-derived contract; this observation must never authorize capture.
+pub(crate) fn identity_and_connection_for_budget_hint(
+    path: &str,
+) -> std::io::Result<(CameraIdentity, UsbConnectionFacts)> {
+    use std::os::unix::fs::{FileTypeExt, MetadataExt};
+    let metadata = std::fs::metadata(path)?;
+    if !metadata.file_type().is_char_device() {
+        return Err(bad("budget hint requires a character-device path".into()));
+    }
+    let (iface_dir, dev_dir) =
+        usb_dirs_for_numbers(libc::major(metadata.rdev()), libc::minor(metadata.rdev()))?;
+    Ok((
+        identity_from_dirs(&iface_dir, &dev_dir)?,
+        connection_facts_from_dirs(&dev_dir, &iface_dir, Path::new("/sys"))?,
+    ))
+}
+
+fn usb_dirs_for_numbers(major: u32, minor: u32) -> std::io::Result<(PathBuf, PathBuf)> {
     let node = std::fs::canonicalize(format!("/sys/dev/char/{major}:{minor}"))?;
 
     let iface_dir = ancestor_with(&node, "bInterfaceNumber").ok_or_else(|| {
@@ -541,6 +565,20 @@ fn bad(msg: String) -> std::io::Error {
 #[cfg(test)]
 mod tests {
 
+    #[test]
+    fn budget_hint_metadata_rejects_non_usb_paths_without_opening_them() {
+        assert!(super::identity_and_connection_for_budget_hint("/dev/null").is_err());
+        assert!(
+            super::identity_and_connection_for_budget_hint("/dev/irlume-missing-budget-hint")
+                .is_err()
+        );
+        let path =
+            std::env::temp_dir().join(format!("irlume-budget-regular-{}", std::process::id()));
+        std::fs::write(&path, b"not a camera").unwrap();
+        assert!(super::identity_and_connection_for_budget_hint(path.to_str().unwrap()).is_err());
+        std::fs::remove_file(path).unwrap();
+    }
+
     #[test]
     fn connection_facts_bind_controller_speed_and_driver_without_bus_numbers() {
         use std::os::unix::fs::symlink;
diff --git a/docs/SETUP.md b/docs/SETUP.md
index 29513556b95ae7dc46ed3d1701a4286e25696f05..726c9eab4c812e2d8d24f8b53c1ece23c7d66fb8 100644
--- a/docs/SETUP.md
+++ b/docs/SETUP.md
@@ -408,10 +408,15 @@ that phrase in `irlume logs`.
 
 `privileged_grouped_pad_evidence=1` in `/etc/irlume/settings.conf` (or
 `IRLUME_PRIVILEGED_GROUPED_PAD=1`) puts `sudo`, `su`, `doas` and polkit on the
-same collection, and gives them the login grace window it requires — only for a
-request that can actually use it, so a concurrent, unqualified or
-runtime-demoted pair, and experimental IR-only, keep the short window they have
-today. An explicit `IRLUME_GRACE_MS` still decides the budget on its own.
+same collection. Eligible privileged dual-camera requests may reserve its login
+grace window using a metadata-only hint from the pair's validated stored
+sequential record. No camera is opened for that hint. Opt-out, experimental
+IR-only, missing-model, forced-schedule and explicit-window cases skip it;
+missing, mismatched or invalid records keep the short default. The window starts
+before configuration/hint reads, and an explicit `IRLUME_GRACE_MS` still decides
+the budget on its own. The hint is not capture qualification: a changed live
+stream contract or later runtime demotion can still prevent grouped collection
+while that request retains its already chosen deadline.
 Evidence requirements do not move: the full vote window still has to close
 before a grant, and every PAD and liveness threshold is unchanged. Two costs are
 real. A privileged authentication now takes as long as that collection (measured
```
</details>

archledger:
Migrating this PR into #711 so the audited revision can land on current main.

@FizykPSX thank you for this contribution and for the two responsive revision rounds. Your three commits are preserved verbatim in the new PR with your authorship and DCO sign-offs, and the final commit there applies the reference revision posted above: the metadata-only lazy budget hint that removes the eager camera-probe regression and the unused helper found in review.

Closing this one in favor of the migration; please review the final shape there, your feedback is welcome.

PR #711, opened 2026-09-14
feat(auth): let privileged services collect grouped PAD evidence

archledger:
## What & why

Migrates #703 by @FizykPSX onto current main with the revisions agreed during its review. The original commits are preserved with their authorship and DCO sign-offs; the final commit applies the tested reference revision posted in the review thread (https://github.com/archledger/irlume/pull/703#issuecomment-5655215917).

A camera pair that cannot capture RGB and IR concurrently scores one RGB frame per authentication attempt, so one attempt casts a single ViT vote of the five the RGB PAD decision needs. The greeter and lock screen already close that vote window through bounded sequential collection, but the collector was scoped to those consented surfaces, leaving privileged surfaces with one vote per attempt: on the reporting host every sudo and polkit attempt settled as `collecting RGB PAD evidence` while `irlume identify` scored the same face seconds earlier.

`privileged_grouped_pad_evidence=1` in `/etc/irlume/settings.conf` (or `IRLUME_PRIVILEGED_GROUPED_PAD=1`) opts Elevation services such as `sudo`, `su`, `doas` and polkit into the same bounded sequential collection and gives those services the login grace window that collection requires. It defaults off, so privileged surfaces are unchanged without it. The key widens which services may collect the evidence, never how much evidence a grant needs: PAD vote count and thresholds, liveness and identity thresholds are untouched, and a privileged grant still closes the whole vote window. Credential release keeps its existing scope.

The privileged budget is decided where the window is created, so capture and response admission share one deadline. The route fact comes from a metadata-only hint: device node stat, sysfs identity and connection metadata, and one bounded validated v2 qualification record. It never opens a video node, issues an ioctl or acquires a camera lease. The hint is lazy: opt-out, experimental IR-only, credential release, non-privileged requests, missing models, forced capture schedules and explicit window overrides never evaluate it. Missing, changed, malformed or unreadable evidence answers false, and live capture qualification remains mandatory before grouped capture.

## Credits

The feature and both review revisions were authored by Pawel Waszkiewicz (@FizykPSX); the first three commits are his, cherry-picked with authorship and sign-offs preserved. His hardware observations (sudo and polkit grouped grants with five PAD votes on a sequential pair) drove the design. The final commit is the maintainer revision from the #703 review thread.

## Testing

Rebased onto current main (`02e399b`) with only a CHANGELOG entry conflict; all four commits are verbatim cherry-picks with authorship and sign-offs preserved.

- `cargo fmt --all -- --check`: pass
- `cargo clippy --locked --workspace --all-targets -- -D warnings`: pass
- `cargo test --locked --workspace`: 1742 passed. Three tests failed only under this shared host's full-suite load and pass in isolation on both this branch and plain main in the same environment, so they are pre-existing host flakes, not failures of this change: `emitter_journal::tests::this_cameras_record_is_found_whatever_the_directory_order`, `ir_emitter::tests::a_caught_signal_is_noted_and_then_handed_back`, and the cli_dispatch `diag_reports_a_missing_template_key_as_recoverable_when_a_recovery_wrap_exists`.
- The reference revision itself was previously verified on its exact tree (1357 scoped tests, eight new focused tests covering lazy exclusions, override snapshotting, expiry and zero semantics, cancellation precedence, stored-record matching, identity and connection mismatch, malformed records and symlinks, and separation from live stream qualification), and a read-only probe on two real hosts observed zero video-node opens and zero ioctls while the hint answered from stored metadata. See the review thread on #703.
- Reporter hardware (in the commit messages): sudo and polkit grouped grants with five PAD ViT votes on a sequential pair, where the same attempts previously settled as `collecting RGB PAD evidence`.


archledger:
The migrated work landed in main as #711 with your three commits preserved, and it will ship in the next release (it merged after the v0.12.1 tag, so today's release packages do not contain it yet). Thank you again for the feature and the two revision rounds; the grouped PAD grants you measured on the sequential pair are the reason it exists. If you re-run your sudo and polkit observations against a build from current main once the next release is out, the results would be a welcome data point.

PR #717, opened 2026-09-15
perf(camera): reuse delivered-rate evidence within bounds (ADR-0021)

archledger:
## What & why

A sequential capture pair cannot stream RGB and IR at once, so every
authentication pays, per stream and serially: open, negotiate, the role
startup flush, and a 30-delta delivered-rate observation window whose frames
are all discarded. On the measured BRIO sequential pair that machinery is
roughly 4-6 s of a 10.7 s authentication: the rate windows paid twice,
serially, are the dominant contributor, while the IR evidence burst itself
is only ~1.9 s.

This implements ADR-0021 (docs/adr/0021-rate-evidence-amortization.md): when
the same camera node and stream role completed a full floor-passing
30-delta window in this process within five minutes, and no stream error,
recovery epoch or failed probe intervened since, a new session may admit on
a continuity probe:

- the role startup flush still runs (the probe measures THIS session)
- five positive deltas, judged by the same exact integer `meets_floor`
  arithmetic used everywhere else, must clear the same floor
- a failed probe invalidates the cached evidence and falls back to the full
  fill, so the change cannot turn a failing stream into a passing one
- the per-frame sliding floor judgment keeps running for the whole burst
  exactly as before, with the same typed BelowFloor refusal
- reported evidence carries the honest window count, which grows during the
  burst; no downstream consumer gates on it (the auth path reads sequence
  gaps)
- `IRLUME_RATE_AMORTIZATION=0` restores the previous behavior

Security posture, from the rate gate's own contract: the window is
transport-health evidence (degraded link, starved stream), not the
anti-injection control; camera identity is pinned per open by
`verify_pinned`. Five-delta windows were measured invalid only as SOLE
admission evidence (the slice-8 design); here they are a continuity
confirmation of a prior full window, and the sole-evidence path is
unchanged at 30 deltas.

## Expected effect

Second and later sequential attempts per process: per-stream machinery
drops from ~2.7 s to ~0.4-1.0 s (flush plus probe), cutting full
authentication by roughly 4-5 s. The first attempt per process is
unchanged (its full fills populate the cache). The estimate is derived
from the measured machinery cost (PR700/PR702 evidence); attended
hardware timing validation on the sequential pair follows before release.

## Testing

- Tests written first and watched failing: cache hit/stale/invalidate
  semantics, the kill switch, the inverted-unset default bug it exposed,
  second-session probe admission (accounting pins seed plus five discards
  and the following per-frame `next` succeeding), below-floor probe
  fallback (typed BelowFloor on the next frame, cache invalidated and not
  re-recorded by a below-floor window), and kill-switch-forced full fill.
- fmt, all-target Clippy with warnings denied, and the full camera suite
  (726 passed) plus auth, common and daemon suites pass on the branch.


archledger:
Attended hardware validation on the BRIO sequential pair completed (candidate on an isolated socket, packaged daemon and state untouched, engagement instrumented per stream).

Measured, real Authenticate grants, refusals never pooled:

- control (IRLUME_RATE_AMORTIZATION=0): median 11146.7 ms, 3 grants
- intermediate state with only the session sites wired: median 10404 ms, 3 grants
- final wiring (RGB session + both IR one-shot helpers + adaptive startup probing): warm grants 9298.6 and 9153.4 ms

About 2.0 s saved (roughly 18 percent) on this camera. Engagement was proven per stream: probe admissions for both roles on warm attempts, full fills only on the cold attempt, and the staged timings matched the staged wiring. The head commit wires the paths the first slice missed and adds the adaptive-path regression test. Hosts whose IR delivers at the 15 fps floor should see roughly double this saving, per the ADR-0021 window math.

PR #725, opened 2026-09-16
feat(camera): observe the largest inter-frame gap in the rate window (ADR-0023)

archledger:
Third instrumentation slice for the ADR-0023 measurement authority (#722). Stacked on #724 (the same-math wiring); base branch feat/camera-tune-live-evidence-20260915. If #724 merges first, this branch will be rebuilt as a single commit on main before merge (same recovery as #724 itself).

## What it adds

- The rate window already holds every delta it counts; this slice exposes the largest one: `RateWindow::max_delta_us()` is a fold over the held deltas, `DeliveredRateEvidence` carries `max_inter_frame_gap_us` (field, constructor arg, accessor), and the production construction site passes the observation through.
- `measurement::RateRound::from_window_facts` gains the gap parameter; camera-tune rounds now record `Some(gap)`, so the aggregate `RateEvidence.max_inter_frame_gap_us` is a real observation rather than `None`. A gap exceeding the span rejects the round at construction.
- The v1 tune policy deliberately still sets no gap limit: the observation exists so a future policy version can set one with evidence beneath it. No wire-type, verdict, or policy behavior changes.

The motivating measurement: the N930W holds its floor on average while containing a ~625 ms in-window gap - exactly the signature an average-only window hides.

## Validation

- New test: `max_delta_reports_the_largest_held_gap_not_the_average` (34 ms, 1 ms, 626 ms deltas: max names the stall the average hides).
- Updated tests: from_window_facts records the observed gap; a gap larger than the span rejects construction; the gap-unverifiable acceptance test now constructs gap-less rounds explicitly (the production source always observes gaps since this slice).
- 19 measurement + 13 rate-gate tests green; `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean (the provenance enum takes `#[expect(clippy::large_enum_variant)]` with the by-value hot-path reason recorded); camera+daemon suites green except the known pre-#721 tmpfs readdir flake, failing identically on this lineage's main (fixed separately by #721).

## Follow-ups

`--emit-profile` / `--verify-profile` once the evidence set is complete; the loader last.


archledger:
Superseded by #728: same three changes (gap observation, evidence emission, drift verification) cherry-picked as originals onto current main after this branch's base moved and conflicted. All validation reran on the rebuilt tree.

PR #727, opened 2026-09-16
feat(cli): camera-tune --verify-record drift comparison (ADR-0023)

archledger:
Fifth instrumentation slice for the ADR-0023 measurement authority (#722): the verification half of the tooling. Stacked on the #725 branch (which carries the full stack after #726 merged into it).

## What it adds

- `sudo irlume camera-tune --verify-record FILE`: re-measures the camera and prints a per-arm drift report against a reference evidence artifact produced by `--emit-record` (real artifacts exist since the 2026-09-16 Phase B run on the N930W and BRIO).
- The comparison uses the reference's completed round count for the fresh run, so both runs judge the same workload.
- Drift lines name: p50 direction by exact rational cross-multiplication (`faster`/`slower`/`same` - never float equality), fps rendering for humans only; max inter-frame gap change; rounds; both acceptance verdicts. Missing arms or roles are named explicitly, never silently skipped.
- The fresh artifact goes to a process-keyed temporary path and is removed after comparison. Verification installs nothing, persists nothing, and modifies no qualifications (ADR-0023: `--verify-profile` semantics, record-level while the profile schema is still gated).

## Validation

- New test: `drift_report_names_direction_gaps_and_missing_arms` (slower/same detection by rational direction; missing fresh arm named).
- `cargo fmt --check`; `cargo clippy --workspace --all-targets -- -D warnings` clean; full cli suite green.
- Real-artifact smoke executed against the sealed Phase B artifacts during development of this slice's parent session.

## Not in this slice

The loader and the TOML `--emit-profile` remain gated on the ADR's Phase B per-field decisions per the review sequence.


archledger:
Superseded by #728 (same reason as #725): the verify-record change rides #728's third commit.

PR #732, opened 2026-09-16
feat(core): multi-camera secondary store (ADR-0024 Phase 1 foundation)

archledger:
First Phase 1 slice of ADR-0024 (#731): the secondary-store data model in `irlume-core::multi_camera`. No authentication consumer yet - this is the storage-boundary foundation the ADR's acceptance matrix tests against.

## What it adds

- **`SecondaryStore`**: separate, versioned (`SECONDARY_STORE_VERSION = 1`), strictly parsed (deny_unknown_fields on every secondary type) home for secondary camera groups, never read by pre-multi-camera binaries (outside legacy discovery; nothing points legacy code at it).
- **Activation binding (§1.1)**: the store carries `primary_snapshot_sha256` - a digest of the exact primary-file bytes the authorization was captured against. `activation_against(current_primary_bytes)` yields Active or InactivePrimaryChanged; any primary change (including a semantically equivalent legacy rewrite) or a missing/unreadable primary is a change. Fail-closed, no guessing path.
- **Groups (§2)**: immutable `CameraGroupId` + complete role-labelled `GroupPair`; `group_for_pair` matches the COMPLETE pair (a hybrid of two groups' endpoints never matches; a bound side is not a wildcard). Duplicate ids rejected.
- **Whole-store semantics (§1.2)**: `load_secondary` returns `Ok(None)` only for a missing file; unsupported version, corrupt JSON, or any structural violation rejects the entire store with a distinct diagnostic kind (Corrupt / IncompatibleVersion / Invalid / Io). Never partially accepted, never treated as empty.
- **Bounds (§3)**: max 8 groups, 30 scans per profile-group, 64 total scans, 8 MiB store ceiling, identifier length limits - validated on load AND before any write.
- **Durable save (§4.1's durability note)**: validate -> serialize -> temporary file -> file fsync -> rename -> directory fsync. Invalid stores are never persisted.
- **Stated honestly in the module docs**: `FaceScan`'s serde contract is fixed by the legacy format and tolerates scan-level unknown fields; all SECONDARY-structural strictness is enforced here. Also documented: profile-by-name references are safe *because* activation is snapshot-bound (a primary rename deactivates the store before a stale association could authorize).

## Validation

- 5 tests: valid round-trip (byte-equal serialization; `FaceScan` has no PartialEq by legacy contract) + missing-file-is-None; activation follows the primary bytes exactly (match / changed / missing); whole-store rejection for unsupported version, corrupt JSON, and duplicate group ids; complete-pair matching with hybrid refusal; bounds enforced with invalid stores never persisted.
- `cargo fmt --check`; `cargo clippy -p irlume-core --all-targets -- -D warnings` clean; full irlume-core suite green.

## Next Phase 1 slices

Camera-scoped enrollment views (§3's table wired into the matching/calibration consumers), the cross-store commit protocol and grant-boundary validation (§4), then the daemon authorization primitives - each landing under the acceptance matrix rows they satisfy.


archledger:
Second Phase 1 slice of ADR-0024 (#731), added to #732's branch (same workstream, one review surface).

## What it adds

`irlume-core::multi_camera::views` - the validated per-camera-group views ADR-0024 §3 requires:

- **Composition is the only door**: `CameraScopedViews::compose(primary, primary_bytes, secondary)` validates the secondary store, checks the activation digest against the primary's exact bytes, and yields one view per group (primary first). Authentication consumers take the views, never the raw stores - the structural guarantee §3 demands.
- **Scoped derived state, pooled arithmetic unchanged**: `pitch_neutral` (median, pitches > 0, ≥2) and `ir_ratio_floor` (75% of minimum, ≥2 applicable) are computed only from the group's own scans - scoping is a pure restriction, so the primary group's results are identical to today's pooled behavior.
- **Candidate selection scoped by group and pipeline**: RGB by recognizer embedding space (legacy `None` = legacy space, per the existing rule), IR by adapter space + dimension.
- **Readiness predicates kept distinct** (§3): capture target (`DEFAULT_ENROLL_SCANS`), calibration-fittable (`MIN_FIT_PAIRS` over compatible IR pairs), and raw compatible counts - observations, not policy; the RGB and IR pipelines are named separately.
- **Fail-closed composition**: an inactive secondary store contributes nothing while the primary view stays valid (§1.2); an invalid secondary store refuses composition with its diagnostic rather than guessing.

## Validation (6 new tests, 11 total in the module)

- Primary-alone views with exact scoped pitch (0.12) and ratio floor (1.5).
- Active secondary contributes only its own group: desk pitch 0.52 / floor 0.9, while the primary's values are untouched - foreign-group isolation by measurement.
- Changed primary bytes: secondary deactivated, primary view intact.
- Candidate selection: group + space + dimension filters (including zero-match cases for wrong spaces/dims).
- Readiness distinctness: 3 scans = below capture target yet calibration-fittable (the predicates answer different questions).
- Duplicate-id secondary store refuses composition with the diagnostic.

`cargo fmt --check`; `cargo clippy -p irlume-core --all-targets -- -D warnings` clean; full irlume-core suite green.

## Next Phase 1 slices

Cross-store commit + grant-boundary validation (§4.1/§4.2), daemon authorization primitives, then the wiring into the live auth path under the acceptance matrix.

archledger:
Third Phase 1 slice of ADR-0024 (#731), added to #732's branch: `multi_camera::commit` - the cross-store transaction and the serialized grant boundary.

## Commit protocol (§4.1)

- `publish_with_intent`: durable intent journal FIRST (serialized new secondary store + its primary-digest context, fsync file + dir), then the durable rename of the new store (the COMMIT POINT), then journal removal. Every step fsyncs its file and directory - the durability detail the ADR calls out.
- Recovery is deterministic **recover-forward**: `resolve_commit` completes any found journal from the journal's own payload (the operation was authorized before the journal existed), then removes it. Never guesses, never mixes generations; the activation digest still governs whether the recovered store is active. Returns `Clean` (no journal) or `Completed` (recovered) - a corrupt or unsupported journal is an error the caller must surface, never silently treat as absent.
- A binding disagreement between the store and the transaction's primary digest refuses publication before anything is written (journal included).

## Grant boundary (§4.2)

- `grant_boundary_check(pinned, current_primary_bytes, current_secondary)`: pure, exhaustive-testable final decision input. Refusals each name their ADR clause:
  - secondary absent at the boundary
  - generation changed during the attempt
  - **primary snapshot changed during the attempt with the secondary generation unchanged** - the acceptance-matrix mid-attempt-rewrite case
  - primary missing/unreadable (§1.1)
  - activation binding stale (§1.1)
  - pinned group no longer present (§4.2 revocation)
- `grant_boundary_now` resolves any journal first, then reads BOTH stores at the boundary and runs the check - the caller-visible contract that nothing is cached across the attempt.

## Validation (5 new tests, 16 in the module)

- Clean publish round-trips with no leftover journal.
- **Simulated crash after the journal**: recovery completes from the journal payload; the recovered store is exactly the intended generation; second resolve is Clean.
- Binding disagreement refuses before any write (journal absent).
- The grant boundary: Grant on exact state; refuses for bumped generation, mid-attempt primary rewrite (byte-level), missing primary, absent secondary, and revoked group.
- `grant_boundary_now` end-to-end: green, then a legacy primary rewrite between attempt start and boundary refuses - the rename-fd contract exercised at the file level.

`cargo fmt --check`; `cargo clippy -p irlume-core --all-targets -- -D warnings` clean; full irlume-core suite green.

archledger:
Fourth Phase 1 slice of ADR-0024 (#731), added to #732's branch: `multi_camera::authz` - the enrollment authorization primitives.

## What it adds (§4)

- **`EnrollmentAuthorization`**: the credential-management token, minted by the daemon ONLY through its single public constructor with an explicit `AuthorizationVia` - a fresh password verification by the target account or an elevated local peer.
- **The structural rule, enforced by the type**: `AuthorizationVia` has exactly two variants. There is no face-granted variant, no authentication-path type converts into the token, and a compile-pinned exhaustive match documents that no third arm can appear silently. "The new camera never authorizes its own addition" is not a runtime check that could be forgotten - it is the absence of a path.
- **Fresh**: bounded validity window (ceiling: one day; callers may mint shorter); expiry refuses.
- **Scoped exactly**: validation compares account AND the complete operation - for an addition, the precise group id and the complete pair. A different group, a hybrid pair (one endpoint swapped), or a removal are all different operations and refuse.
- **One-shot**: consumption is the generation bump of a successful publication; `ensure_not_consumed` refuses a replayed authorization id against the store's recorded last-authorization.

## Validation (4 new tests, 20 in the module)

- Exact scope: valid grant for the exact operation; refusals for wrong account, other group, hybrid pair, and removal-vs-addition.
- Freshness: inside/outside the window; an over-long window cannot be minted at all.
- One-shot: first use passes, different id passes, replay refuses.
- The no-face-variant property pinned by exhaustive match + malformed-mint refusals.

`cargo fmt --check`; `cargo clippy -p irlume-core --all-targets -- -D warnings` clean; full irlume-core suite green.

## Phase 1 remaining

The live-auth wiring (grant-context pinning in the daemon, view consumption at matching/calibration, selection pinning) and the acceptance-matrix integration tests across all four slices.

archledger:
Final Phase 1 core slice on this branch: `multi_camera::coordinator` — the per-attempt object tying the four foundations together, plus the acceptance-matrix integration tests.

## The coordinator

- `SecondaryAuthContext::pin(secondary_path, primary_path, live_rgb, live_ir)`: resolves any leftover commit journal (recover-forward) → loads both stores → composes the camera-scoped views → requires the live pair to resolve to an **ACTIVE** group. Hybrid pairs and stale activations refuse at pin time — a group is never selected by its biometric score (§2).
- Matching and calibration read only `group_view()` — the pinned group's scoped candidates (§3).
- `boundary_check_now()`: the serialized final check, re-reading BOTH stores at the decision point (§4.2), refusing on any drift from the pinned state.

## Integration tests — the acceptance-matrix software rows, end to end

1. **Full attempt story**: pin → read the desk group's scoped view (pitch 0.52, 3 candidates — isolation by measurement) → boundary grants on the unchanged state.
2. **Hybrid live pair never pins** (RGB from the enrolled group, IR from elsewhere).
3. **Mid-attempt legacy primary rewrite**: refused at the boundary with the §4.2 reason; restoring the byte-identical file grants again — the boundary judges CURRENT state, not history.
4. **Authorized revocation by publication**: a password-minted RemoveGroup authorization publishes a store without the group and a bumped generation; the pinned attempt refuses.
5. **Journal resolution before pinning**: a completed generation-2 publish resolves cleanly; pin sees generation 2.
6. **Absent secondary store**: pin refuses; the primary file is untouched for the password fallback.

26 module tests total; `cargo fmt --check`; `cargo clippy -p irlume-core --all-targets -- -D warnings` clean; full irlume-core suite green.

## Phase 1 core status

The complete primitive layer now exists and is integration-tested together: secondary store, camera-scoped views, cross-store commit + grant boundary, enrollment authorization, and the coordinator. What remains is the daemon/engine wiring (request surfaces, engine consumption of the views, selection pinning in the live path) — which lands with the Phase 2 integrated flows.

PR #741, opened 2026-09-17
perf(camera): IR startup flush 10 -> 5, from the fleet fine sweep

archledger:
## Summary

The IR startup flush is a fleet constant paid on the auth critical path by every IR session start (~675ms per start at the 15 fps floor; twice per concurrent attempt with the paired path). The value was 10.

The repo's own fleet measurement (docs/research/2026-08-22-startup-flush-fleet-measurement.md) already contains the fine sweep that justifies less: the **NexiGo N930W is the only node whose startup transient is real** (flush-0 measures 14.018 fps - FAIL; dequeue 3 is the mixed tail at 14.676-14.705; **dequeue 4+ is settled at >=14.734 fps**). ASUS IR has no measurable transient; RGB is 0 everywhere. Ten frames carried 2x headroom over that measurement.

This sets the flush to **5** - one frame of margin past the measured transient. Safety shape unchanged: a future camera whose transient exceeds its flush still fails CLOSED (the window misses its floor, loudly - the doc's stated guarantee), and with PR740's amortization a marginal miss degrades to the full-fill fallback at no correctness cost. Expected effect: ~340ms saved per IR session start.

This is the same evidence -> fleet-constant correction the earlier 30 -> 10 reduction made from the same document, not a per-camera knob (ADR-0023's zero-executable-fields decision stands).

## Tests

camera 743: the eight startup tests that pin the flush-derived arithmetic updated to the new derivation (counts shift by exactly 5; the retained-seed recoverable-window fixture's long interval moves five sequences earlier to keep testing the same recovery geometry). auth 263, fmt/clippy/rustdoc clean.


archledger:
Closing with live evidence from archhost (the N930W, the fleet's knife-edge IR camera at ~1.3% above its floor):

- flush 10 (current main): probe admits on every warm attempt (verified with PR740's admission dlog during its verification round).
- flush 5: ZERO probe admissions; RateEstablishment back to ~3.8s on warm attempts - and worse, each miss invalidates the cached completion, so every attempt re-pays cold.
- flush 7: ZERO probe admissions, same shape.

Root cause of the disagreement with the fine sweep: the 2026-08-22 measurement validated full 30-delta windows seeded at dequeue 4+ - long windows average out the N930W's per-boot delivery variance. The ADR-0021 probe judges a 5-delta window, a point sample that needs its own margin past the transient, and at flush 5/7 that window sits inside variance the sweep's methodology could not see.

The constant stays 10 on main: it is the only value live-verified to keep the amortization path working, and the amortization (PR740) is worth ~2s/attempt where a smaller flush was worth ~0.3s. Reopening this requires 5-delta-window evidence, not 30-delta-window evidence.

archledger:
CORRECTION to the close evidence above: part of it was a measurement artifact (the admission line is a dlog, silent unless debug logging is enabled, which it wasn't during the flush-5/7 runs). With debug on, BOTH flush 10 and flush 5 admit.

The actual finding from traced attempts (RateEstablishment stage, granted runs): ~3.92-3.94s on EVERY attempt at either flush value, admitted or not. The admissions are the RGB side only - the IR probe never admits on the N930W: its 5-delta window sits inside the knife-edge delivery variance (camera at ~1.3% above floor), so IR pays its full fill every attempt and the stage time is IR-fill-dominated either way. The flush constant is therefore immaterial today; keeping 10 on main stays correct, and the real remaining win is an IR-side amortization that tolerates that variance (e.g., a wider probe window or a per-camera probe floor) - needs its own design pass with 5-delta-window evidence.

archledger:
State correction: this PR merged (auto-merge was armed and completed before the close landed - the disarm command in the same batch failed on a repo-path parse error and was not retried). Flush 5 is on main. Given the corrected finding in my previous comment (the flush value is immaterial while the IR side never admits - both values RGB-admit and IR pays its full fill either way), flush 5 standing is acceptable-to-slightly-better (saves 5 frames on every cold IR fill). The 'constant stays 10' sentence in the close comment is superseded.

